PR #12, opened 2018-01-30
Add AzureMachineClass[List] to Scheme

rfranzke:
After integrating Azure the new structs have not been added to the scheme. The node-controller-manager cannot report events on those objects.


rfranzke:
I've updated the document in the GCP support PR (#13).

hardikdr:
Great, Thanks.

PR #13, opened 2018-01-30
Enabling support for GCP

rfranzke:
These commits enable support of GCP.


Gopkg.toml
```diff
@@ -45,3 +45,12 @@
 [[constraint]]
   branch = "release-1.8"
   name = "k8s.io/kubernetes"
+
+[[constraint]]
```
msohn: the new dependencies should be reflected in NOTICE.md

I think @hoeltcl should review changes to dependencies to ensure we don't add dependencies which come with a license we don't want and we cannot consume according to internal standards. We should agree on a process here and document it in https://github.com/gardener/documentation/blob/master/CONTRIBUTING.md

rfranzke: I've added the new Google Go SDK description to the `NOTICE.md` file.
The update of other dependency is respecting the constraints in the `Gopkg.toml` file, thus that should be fine.
I don't think somebody should manually review each PR regarding new dependencies/dependencies with disallowed licenses/... We should rather invest into some automation which does these things automatically. Already now with only a few number of projects the double maintenance of `Gopkg.toml` and `NOTICE.md` is very tedious...

msohn: yes, that's a good idea. I'll also discuss this with @hoeltcl

pkg/driver/driver_gcp.go
```diff
@@ -0,0 +1,242 @@
...
+	oauthClient := oauth2.NewClient(ctx, jwt.TokenSource(ctx))
+	computeService, err := compute.New(oauthClient)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return ctx, computeService, nil
+}
+
+func waitUntilOperationCompleted(computeService *compute.Service, project, zone, operationName string) error {
```
prashanth26: Do we need this? Our controller is okay with a non-blocking (async) create call to the cloudProvider. The confirmation of create happens when the node object registers with the cluster.

rfranzke: Not sure, so you don't receive an answer from GCP Go SDK regarding success of your request, but only an operation id which you can use to continuously ask for the state of it.
For example, while testing I did not enter a valid service account email. I did not receive an error here: https://github.com/gardener/node-controller-manager/pull/13/files/389e77556ff495c7efa4a9dec0b647ca853e8f4c..321f9396a3083ea602fb4789b42950c9a73d51d1#diff-af75cdff1c0efd51940289aeedf2ef4eR135
Thus, in the end I did not know whether an instance is getting created or not (without asking for the state of operation).
Same applies to deletion, without asking the operation for its state the driver would never know whether a instance has been successfully terminated or not. When it reports success but the VM still exist, we end up with tons of orphans.

pkg/controller/machine.go
```diff
@@ -110,9 +109,7 @@ func (c *controller) reconcileClusterMachine(machine *v1alpha1.Machine) error {
 		return err
 	}
 
-	splitProviderId := strings.Split(machine.Spec.ProviderID, "/")
-	machineID := splitProviderId[len(splitProviderId)-1]
-	driver := driver.NewDriver(machineID, secretRef, machine.Spec.Class.Kind, MachineClass, machine.Name)
+	driver := driver.NewDriver(machine.Spec.ProviderID, secretRef, machine.Spec.Class.Kind, MachineClass, machine.Name)
```
hardikdr: Thanks for this catch.

pkg/driver/driver_gcp.go
hardikdr: Update: We discussed regarding this over the call. 
We found that, different cloud-provider sdks manages the call analogues to createMachine differently. For example AWS-sdk immediately responds with error, if configuration provided to the call is buggy, on the other hand gcp does not throw any error immediately. gcp-sdk rather expects the user to check the status field of output object, after createMachine call.

We concluded that, we will keep the implementation different based on cloud-provider, considering the fact that this difference would be baked inside the driver-specific code, which is acceptable.

hardikdr:
@rfranzke Can you please resolve the conflicts with master branch, so that we can then merge it.

PR #16, opened 2018-02-05
AWS VMs "Name" tag defaults to machine object name

### LINKED ISSUE #10: VM name should match to machine name
When creating a `MachineDeployment` with name `x`, the machine-controller-manager would create a `MachineSet` and out of that `Machine`s. These `Machine` objects have names `x-<hash>`. However, the names of the VMs in AWS are only `x`.

It would be nice if the VMs would have the same name like the `Machine` objects so that one could easily see which VM belongs to which object.

rfranzke:
- In the `AWSMachineClass` object it is possible to define tags which are applied to the VM.
- If no tag with key `Name` has been defined in the class, the AWS driver will default it to the name of the machine object it acts on.
- Otherwise, the value provided in the class is taken.

closes #10

### DIFF
--- a/pkg/driver/driver_aws.go
+++ b/pkg/driver/driver_aws.go
@@ -99,6 +99,11 @@ func (d *AWSDriver) Create() (string, string, error) {
 
 	// Add tags to the created machine
 	tagList := []*ec2.Tag{}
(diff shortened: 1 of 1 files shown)


prashanth26:
AWS VM name tag defaults to machine object name

PR #19, opened 2018-02-05
add support for integration of seed cluster

hardikdr:
- Made Machine, MachineClass, MachineSet, MachineDeployment namespaced.
  - Exposed `namespace` flag where controller would look for and create CRDs.
- Added new command-line flag `kubeconfig-seed` for seed-integration.
  - purpose is to provide invisibility to Machine-CRDs from the cluster.
  - however, if `kubeconfig-seed` is not provided then old hook of `kubeconfig` will work as usual.


Makefile
```diff
@@ -22,7 +22,7 @@ export GOBIN
 
 .PHONY: dev
 dev:
-	@go run cmd/node-controller-manager/controller_manager.go --kubeconfig=dev/kubeconfig.yaml --v=2
+	@go run cmd/node-controller-manager/controller_manager.go --kubeconfig=dev/kubeconfig.yaml --kubeconfig-seed=dev/kubeconfig-seed.yaml --namespace=default --v=2
```
prashanth26: Would it be better if we gave more understandable names for the kubeconfigs. Maybe _control-kubeconfig and target-kubeconfig_ ?

docs/development/local_setup.md
```diff
@@ -61,7 +61,7 @@ $ kubectl apply -f kubernetes/crds.yaml
 
 ### Get started
 
-Create a `dev` directory and copy the Kubeconfig of the kubernetes cluster used for development purposes to `dev/kubeconfig.yaml`.
+Create a `dev` directory and copy the Kubeconfig of the both kubernetes clusters used for development purposes to `dev/kubeconfig.yaml` and `dev/kubeconfig-seed.yaml`.
```
prashanth26: We need to make the documentation more understandable. Also `make dev` fails to start right now when there is no kubeconfig-seed, we should make it more robust ?

kubernetes/machine-deployment.yaml
```diff
@@ -4,11 +4,12 @@ apiVersion: machine.sapcloud.io/v1alpha1
 kind: MachineDeployment
 metadata:
   name: test-machine-deployment # Name of the machine deploy
+  namespace: test
 spec:
   #paused: true [uncommment this when you want to pause]
   #rollbackTo:
     #revision: 0 [uncommment this when you want to rollback to previous version]
-  replicas: 3 # Number of healthy replicas that should always be healthy
+  replicas: 2 # Number of healthy replicas that should always be healthy
```
prashanth26: Can we change it back to 3 itself?

Makefile
rfranzke: I agree with Prashanth!

docs/development/local_setup.md
rfranzke: Agreed, we must be able to start as well when no "management kubeconfig" has been provided.

kubernetes/azure-machine-class.yaml
```diff
@@ -4,6 +4,7 @@ apiVersion: machine.sapcloud.io/v1alpha1
 kind: AzureMachineClass
 metadata:
   name: test-azure # Name of aws machine class goes here
+  namespace: test
```
prashanth26: Same has to be added for AWS class example as well.

cmd/node-controller-manager/app/controllermanager.go
```diff
@@ -177,20 +192,24 @@ func StartControllers(s *options.NCMServer,
 	if availableResources[awsGVR] || availableResources[azureGVR] || availableResources[gcpGVR] {
```
prashanth26: availableResources[] checks for CRDs in shoot cluster currently. What we'll need is a kubeClient that points to the seed cluster's coreV1 objects ?

rfranzke:
You need to fix `pkg/apis/machine/install/install.go` (regarding `RootScopedKinds`); what does line 39 there?

rfranzke:
We need to split the example RBAC ClusterRoles/Bindings in the `kubernetes` folder. When using a different cluster for managing the CRDs then we need to create in both clusters proper RBAC roles.

Makefile
hardikdr: Yup I would agree with that. `control-kubeconfig` and `target-kubeconfig` sounds good,  except I wanted to be little convenient to the user, not aware of separate control-k8-cluster, `kubeconfig` then could have helped. ;)
Will make the change soon.

docs/development/local_setup.md
hardikdr: We can add, how to manipulate 2 kubeconfigs and have it inside the documentation in detail.  `make dev` would have failed because I was expecting 2 kubeconfigs under `dev/`, may be I will make changes in the make-rules to work fine when only one kubeconfig is available.

kubernetes/machine-deployment.yaml
hardikdr: Yeah I also like prime numbers :D

kubernetes/azure-machine-class.yaml
hardikdr: Sure, make sense, missed it somehow.

cmd/node-controller-manager/app/controllermanager.go
hardikdr: Yes, this is interesting. 
We should ideally have 3 internal clients, controlCoreClient, controlMachineClient , targetCoreClient. We would need to access `secrets`, `crd` via controlCoreClient.

PR #23, opened 2018-02-08
Fix buggy tags for AWS driver

hardikdr:
All of the AWS machines created under MachineDeployment were assuming same tag-name due to the bug.


rfranzke:
Ha, thank's...

PR #29, opened 2018-02-14
Better validation and made examples consistent

prashanth26:
1. Better validation for ClassKind
2. Made examples more consistent
3. Changed log levels while adding machine deployment

### DIFF
--- a/kubernetes/deployment/control-cluster-role.yaml
+++ b/kubernetes/deployment/control-cluster-role.yaml
@@ -1,41 +1,41 @@
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRole
 metadata:
-  name: machine-controller-manager
+  name: machine-controller-manager-control
 rules:
-- apiGroups:
- - machine.sapcloud.io
- resources:
- - awsmachineclasses
- - azuremachineclasses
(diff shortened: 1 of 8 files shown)


kubernetes/deployment/control-cluster-role.yaml
```diff
@@ -1,41 +1,41 @@
...
+   - deletecollection
+   - get
+   - list
+   - patch
+   - update
+   - watch
+-  apiGroups:
+   - ""
+   resources:
+   - nodes
```
rfranzke: FYI: I'm pretty sure these roles are too wide, i.e. allow too much access rights. For example, in the control cluster the MCM does not need to do anything with the nodes.
Can you determine proper ClusterRoles for both control and target clusters with access rights of minimum viable scopes?

hardikdr: Ah yes, "nodes" got in because we initially targetedit for the case when user have both target/control cluster same. In that case user can only create cluster-role for control-cluster and everything should work.
we have not separated the clusters, we can not remove "nodes" from the Clusterole of the control-cluster. Same could thought for "endpoint" being in target-cluster.
Will check for more points which can be removed.

Issue #67, opened 2018-04-18
Nodes not being drained before deletion


prashanth26:
Nodes are not being drained every time during machine deletion/rolling update.

PR #68, opened 2018-04-18
Bugfix: Draining issue and Azure error handling

prashanth26:
- The drain nodes logic was reading an older version of the object while making drain call, there by fetching the wrong last operation timestamp. This lead to forceful draining of nodes. This was fixed by updating the updateMachineStatus method to return the new object instead of the old one.
- Azure didn't handle error handling for cases when NIC wasn't created properly. This commit fixes that by embedding appropriate error messages when errors occur in Azure driver machine creation logic.

Resolves: https://github.com/gardener/machine-controller-manager/issues/67


prashanth26:
@rfranzke

rfranzke:
Thank you!

Issue #73, opened 2018-04-23
Orphan NICs issue


prashanth26:
The Azure driver doesn't handle error handling well during machine creation. If the NIC creation passes and VM creation fails, the NIC deletion doesn't go through in the deletion logic. 

Solution: Handle deletion of NIC when VM creation fails.

Issue #74, opened 2018-04-23
Handle proper threads handling for safety loop


prashanth26:
The safety loop right now keeps spinning up go routines on every instance of safety loop. But this is not good as when many orphan machine resources (VMs,NIC,DIsks) are being deleted, the go routines over-flood the container and leads to excessive resource usage. Finally leads to hanging of container.

prashanth26:
- It was observed that there are many Azure API requests going through, leading it to reaching its rate limits. This is due to MCM (and possibly Gardener and Kube-controller-manager) pinging Azure often to check the status of deployed resources. 

- Fixing this issue would also solve this issue of rate limiting by changing the orphan VMs loop interval to 10mins or more.

PR #85, opened 2018-05-16
Bugfix: Azure NIC and VM creation made into a transaction

prashanth26:
Azure driver's VM creation logic consisted of two steps, NIC creation and VM creation respectively. When creation of NIC succeeds and VM creation fails, it leads to creation of orphan NICs. This commit fixes that.

Resolves: https://github.com/gardener/machine-controller-manager/issues/73

### DIFF
--- a/pkg/driver/driver_azure.go
+++ b/pkg/driver/driver_azure.go
@@ -109,6 +109,14 @@ func (d *AzureDriver) Create() (string, string, error) {
 	nicParameters, err = interfacesClient.Get(resourceGroup, nicName, "")
 	err = onErrorFail(err, fmt.Sprintf("interfaces.Get for NIC '%s' failed", nicName))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


prashanth26:
@rfranzke 

This partially fixes the issue caused earlier in the staging where 1000s of orphan NICs were created. It basically deletes any NIC for which VM couldn't be created, which was the issue earlier. But there is a second part to it [Refer:https://github.com/gardener/machine-controller-manager/issues/74] that leads to hanging of container when thousands of orphan VMs are spun up in parallel. This issue still needs to be resolved and shall be done later. 

Also I deleted all the orphan NICs found on the `staging-azure-seed` cluster.

Fixes: https://github.com/gardener/machine-controller-manager/issues/73
TBD: https://github.com/gardener/machine-controller-manager/issues/74

rfranzke:
Thank you!

Issue #82, opened 2018-05-11
MachineSet got frozen during scale down


rfranzke:
The `pl-1` cluster in the `perf-load` project on the `dev` landscape got frozen during scale down. Hence, when the user wanted to scale up again, nothing happened.

The log of MCM for the relevant event:

```
I0510 12:37:37.745168       1 event.go:218] Event(v1.ObjectReference{Kind:"MachineDeployment", Namespace:"shoot-perf-load-pl-1", Name:"shoot-perf-load-pl-1-worker-jmrk9-z1", UID:"c4fbdd58-4ed3-11e8-aed6-0a580a2c0111", APIVersion:"machine.sapcloud.io", ResourceVersion:"121492915", FieldPath:""}): type: 'Normal' reason: 'ScalingMachineSet' Scaled down machine set shoot-perf-load-pl-1-worker-jmrk9-z1-55dc58d7b5 to 2
I0510 12:37:37.798548       1 machineset.go:405] Too many replicas for  shoot-perf-load-pl-1/shoot-perf-load-pl-1-worker-jmrk9-z1-55dc58d7b5, need 2, deleting 38
I0510 12:37:37.798829       1 controller_utils.go:611] Controller shoot-perf-load-pl-1-worker-jmrk9-z1-55dc58d7b5 deleting machine shoot-perf-load-pl-1-worker-jmrk9-z1-55dc58d7b5-mpwnv
I0510 12:37:37.798846       1 controller_utils.go:611] Controller shoot-perf-load-pl-1-worker-jmrk9-z1-55dc58d7b5 deleting machine shoot-perf-load-pl-1-worker-jmrk9-z1-55dc58d7b5-wpcqg
I0510 12:37:37.798862       1 controller_utils.go:611] Controller shoot-perf-load-pl-1-worker-jmrk9-z1-55dc58d7b5 deleting machine shoot-perf-load-pl-1-worker-jmrk9-z1-55dc58d7b5-5nk8h
I0510 12:37:37.798845       1 controller_utils.go:611] Controller shoot-perf-load-pl-1-worker-jmrk9-z1-55dc58d7b5 deleting machine shoot-perf-load-pl-1-worker-jmrk9-z1-55dc58d7b5-fbsgm
[...]

Issue #81, opened 2018-05-09
One machine too many deleted when scaling down


vlerenc:
I had scaled-out the dev/perf-load/pl-1 cluster for the performance team to 40 machines and then I scaled it down from 40 to 2. Strangely, instead of keeping 2 existing machines (my expectation), I only kept one. 39 machines got terminated and I got a new one instead to make up for the one that I shouldn't have lost. Do you have an explanation? We should try to keep the machine set as stable as possible and avoid unnecessary terminations and this looked like a termination that we should have avoided, wouldn't you agree?

hardikdr:
We should also check the bug here, where sometimes FrozenCondition is missing but freeze label exists on the MachineSet or vice versa.

hardikdr:
Hi Vedran,
Gaurav had been working on it recently, but could not really reproduce it, I also tried sometime back but dint help. We plan to get this bug resolved by next release.

vlerenc:
Thanks, I can reproduce it nicely:
- I create a standard shoot with two nodes and an older Kubernetes version, e.g. v1.9
- I upgrade the cluster to the next minor version, e.g. v1.10
I did that a couple of times and my `kubectl` watch always shows at one point in time only one v1.9 machine, then two v1.10 machines come back. Maybe this helps to reproduce the issue? Anyway, no real hurry, just thinking that it would be good to resolve it. People start noticing what we run more and more (now they question our forced security patch roll-outs - a valid question, AWS does it like we are doing it, but GKE only warns/recommends - maybe an option for us as well, but that's an entirely different topic; just mentioning it, because my impression is, we see more and more semi-productive workloads and some parties/software is not yet "optimally adapted" to Kubernetes, which is why we should make it even more stable).

ggaurav10:
Hi 
Is it possible that this is due to the value `spec.strategy.rollingUpdate.maxUnavailable` set to 1 (I believe this is the default) in machine deployment?
This value signifies the maximum number of nodes that can be unavailable out of the set number of replicas during rolling updates.

vlerenc:
Ah, now that you say it, this sounds like a reasonable explanation. However, the intention back then was somehow another, I believe. The new machines should be already `Ready`, before starting to drain the `maxUnavailable` number of machines (or more precisely?, for every machine that shall be drained a new `Ready` machine should be available, so that the scheduled pods find immediately a new node). Would do you think?

ggaurav10:
i skimmed through
[...]

PR #86, opened 2018-05-18
Fix unexpected freeze during scale-down 

hardikdr:
MachineSet currently gets frozen unexpectedly during major scale-down. That was because safety-controller does not have any means to understand if it's genuine scale-down or over-shooting of the machine-objects. 
I am solving the issue, by reducing the scope of the Freeze-function, where now it applies only while scale-up and scale-down works as it is even when MachineSet is Frozen. Open for suggestions.

### DIFF
--- a/pkg/controller/deployment.go
+++ b/pkg/controller/deployment.go
@@ -451,7 +451,7 @@ func (dc *controller) reconcileClusterMachineDeployment(key string) error {
 	// If MachineDeployment is frozen and no deletion timestamp, don't process it
 	if deployment.Labels["freeze"] == "True" && deployment.DeletionTimestamp == nil {
 		glog.V(3).Infof("MachineDeployment %q is frozen, and hence not processeing", deployment.Name)
-		return nil
+		//return nil
 	}
 
 	// Validate MachineDeployment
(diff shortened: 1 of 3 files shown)


prashanth26:
Issues to look at -

- https://github.com/gardener/machine-controller-manager/issues/81
- https://github.com/gardener/machine-controller-manager/issues/82
- https://github.com/gardener/machine-controller-manager/issues/88

prashanth26:
Currently this only resolves issue: https://github.com/gardener/machine-controller-manager/issues/82 right?

pkg/controller/deployment.go
```diff
@@ -451,7 +451,7 @@ func (dc *controller) reconcileClusterMachineDeployment(key string) error {
 	// If MachineDeployment is frozen and no deletion timestamp, don't process it
 	if deployment.Labels["freeze"] == "True" && deployment.DeletionTimestamp == nil {
 		glog.V(3).Infof("MachineDeployment %q is frozen, and hence not processeing", deployment.Name)
-		return nil
+		//return nil
 	}
```
ggaurav10: Do we need this `if` block now? If it's left out for logging, then the log statement should also change to reflect the new logic.

ggaurav10: Fixed in #115

PR #89, opened 2018-05-22
Propagate machine status to machineset and machinedeployment

ggaurav10:
The summary of machineobjects which are not in "Running" phase is propagated to machineset and machinedeployment. Decision pending on:

1. which fields are required to propagate

1. Under what conditions should the summary be propagated

### DIFF
--- a/pkg/apis/machine/v1alpha1/types.go
+++ b/pkg/apis/machine/v1alpha1/types.go
@@ -225,6 +225,25 @@ const (
 	ConditionUnknown ConditionStatus = "Unknown"
 )
 
+//MachineSummary store the summary of machine.
+type MachineSummary struct {
+	// +optional
+	Name *string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
+
+	// ProviderID represents the provider's unique ID given to a machine
+	// +optional
+	ProviderID *string `json:"providerID,omitempty"`
+
+	// Phase of the machine
+	Phase *MachinePhase `json:"phase,omitempty" protobuf:"bytes,1,opt,name=phase,casttype=MachinePhase"`
+
+	// Last operation refers to the status of the last operation performed
+	LastOperation *LastOperation `json:"lastOperation,omitempty"`
+
+	// OwnerRef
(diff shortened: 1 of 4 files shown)


rfranzke:
Looking forward for this PR to be merged!

pkg/apis/machine/v1alpha1/types.go
```diff
@@ -333,6 +349,9 @@ type MachineSetStatus struct {
 
 	// ObservedGeneration
 	ObservedGeneration int64 `json:"observedGeneration,inline"`
+
+	// FailedMachines has summary of machines on which lastOperation Failed
+	FailedMachines *[]MachineSummary `json:"failedMachines,inline"`
```
prashanth26: why are the 'json' for failed machines diffferent between machineSet and machineDeployments?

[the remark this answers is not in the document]
ggaurav10: the tags are kept in same format as other fields in the respective structs.

prashanth26:
Gaurav's argument - https://github.com/kubernetes/kubernetes/blob/1a7a609cb1ee1c7746d1ebd01eb1262b106431a3/staging/src/k8s.io/api/apps/v1/types.go

prashanth26:
LGTM. @hardikdr you can have a final look.

pkg/controller/deployment_machineset_util.go
```diff
@@ -109,6 +113,25 @@ func calculateMachineSetStatus(is *v1alpha1.MachineSet, filteredMachines []*v1al
...
+			failedMachines = append(failedMachines, machineSummary)
+		}
+	}
+
+	// Update the FailedMachines field only if we see new failures
+	// Clear FailedMachines if ready replicas equals total replicas,
+	// which means the machineset doesn't have any machine objects which are in any failed state
+	if len(failedMachines) > 0 {
+		newStatus.FailedMachines = &failedMachines
+	} else if int32(readyReplicasCount) == is.Status.Replicas {
```
hardikdr: Why relying on the readyReplicasCount ?
Is there a possibility of having enough ready machine-objects but few machines stuck with an error in the control-plane?

PR #94, opened 2018-05-29
Some unit tests for safety controller.

amshuman-kr:
Added ginkgo sub-package github.com/onsi/ginkgo/extensions/table to vendoring.

Added bugfixes in a couple of places in machine_safety.go to handle machinesets without machinedeployments.

### DIFF
--- a/pkg/controller/machine_safety.go
+++ b/pkg/controller/machine_safety.go
@@ -94,7 +94,7 @@ func (c *controller) checkAndFreezeORUnfreezeMachineSets(wg *sync.WaitGroup) {
 		lowerThreshold := higherThreshold - c.safetyOptions.SafetyDown
 
 		machineDeployments := c.getMachineDeploymentsForMachineSet(machineSet)
-		if machineDeployments != nil && len(machineDeployments) >= 1 {
+		if len(machineDeployments) >= 1 {
 			machineDeployment := machineDeployments[0]
 			if machineDeployment != nil {
 				surge, err := intstrutil.GetValueFromIntOrPercent(
@@ -481,7 +481,7 @@ func (c *controller) freezeMachineSetsAndDeployments(machineSet *v1alpha1.Machin
 	}
 
 	machineDeployments := c.getMachineDeploymentsForMachineSet(machineSet)
-	if machineDeployments != nil && len(machineDeployments) >= 1 {
+	if len(machineDeployments) >= 1 {
 		machineDeployment := machineDeployments[0]
 		if machineDeployment != nil {
 			for {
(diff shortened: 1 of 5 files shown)


pkg/controller/machine_safety.go
```diff
@@ -93,23 +93,27 @@ func (c *controller) checkAndFreezeORUnfreezeMachineSets(wg *sync.WaitGroup) {
 		// Unfreeze machineset when replica count reaches higherThreshold - SafetyDown
 		lowerThreshold := higherThreshold - c.safetyOptions.SafetyDown
 
-		machineDeployment := c.getMachineDeploymentsForMachineSet(machineSet)[0]
-		if machineDeployment != nil {
+		machineDeployments := c.getMachineDeploymentsForMachineSet(machineSet)
+		if machineDeployments != nil && len(machineDeployments) >= 1 {
+			machineDeployment := machineDeployments[0]
+			if machineDeployment != nil {
```
prashanth26: Does this if condition add any value? or am i missing something?

pkg/controller/machine_safety.go
```diff
@@ -476,31 +480,35 @@ func (c *controller) freezeMachineSetsAndDeployments(machineSet *v1alpha1.Machin
...
-			}
-			clone.Status = newStatus
-			clone.Labels["freeze"] = "True"
-			_, err = c.controlMachineClient.MachineDeployments(clone.Namespace).Update(clone)
-			if err == nil {
-				break
+	machineDeployments := c.getMachineDeploymentsForMachineSet(machineSet)
+	if machineDeployments != nil && len(machineDeployments) >= 1 {
+		machineDeployment := machineDeployments[0]
+		if machineDeployment != nil {
```
prashanth26: same as above

amshuman-kr: Same as my reply above :-)

pkg/controller/machine_safety.go
amshuman-kr: This fixes the bug that safety controller panics if a machineset does not have a corresponding machinedeployment.

This is the issue Padma found in safety controller.

amshuman-kr: The panic is "index out of bounds" because machineDeployments array is empty if there are no machinedeployments for the machineset.

prashanth26: You are already testing for array length to be greater than or equal to 1 right? so array would atleast have one element.

prashanth26: And i understood the purpose of the change, I was concerned about the additional if condition alone.

prashanth26: Yeah even I was wondering about the last point. Let's keep it then perhaps, no harm.

amshuman-kr: I have removed the spurious nil check before the len(). But the nested nil check is required, I think.

amshuman-kr:
Thanks, @prashanth26 !

Issue #78, opened 2018-04-26
Update failed machine object state


prashanth26:
Currently when a machine goes into failed state and say later the node joins the cluster after a while, the machine is deleted anyways in the next reconciliation by the machineSet controller. 

Solution: Update machine status of Failed machine objects if it succeeds after a while and the object still exists.

prashanth26:
This PR resolves this - https://github.com/gardener/machine-controller-manager/pull/97

PR #97, opened 2018-05-30
Issue #78: Update failed machine object state

ggaurav10:
Fix: Reconcile the machines that are in "Failed" state also because they might become "Ready" after a while

### DIFF
--- a/pkg/controller/machine.go
+++ b/pkg/controller/machine.go
@@ -252,7 +252,8 @@ func (c *controller) updateMachineState(machine *v1alpha1.Machine, node *v1.Node
 
 	if machine.Status.LastOperation.State != "Successful" {
 
-		if machine.Status.LastOperation.Type == "Create" {
+		if machine.Status.LastOperation.Type == "Create" &&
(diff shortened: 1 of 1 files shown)


pkg/controller/machine.go
```diff
@@ -271,7 +271,7 @@ func (c *controller) updateMachineState(machine *v1alpha1.Machine, node *v1.Node
 					c.updateMachineStatus(machine, lastOperation, currentStatus)
 
 				} else*/
-			if machine.Status.LastOperation.Description == "Creating machine on cloud provider" && len(machine.Status.Conditions) > 1 && machine.Status.Conditions[len(machine.Status.Conditions)-1].Status == "True" {
+			if len(machine.Status.Conditions) > 1 && machine.Status.Conditions[len(machine.Status.Conditions)-1].Status == "True" {
```
prashanth26: Can we change the fetching of the machine status condition by the condition type (instead of index based), to avoid mis match if the conditions go haywire?

prashanth26:
Thanks. Looks good now.

PR #99, opened 2018-06-06
Keep the last failed operations in the machinedpeloyment and machineset

ggaurav10:


prashanth26:
We probably need to discuss how these statuses and healthy machines info would be passed on the gardener?

hardikdr:
Agreed with Prashanth, we need to discuss around how Gardener can independently consume this change.

ggaurav10:
Based on discussions today, closing this PR as current implementation (#89) of updating failed machines' status into machinedeployment and machineset is as per the needs of Gardener.

prashanth26:
Looks good.

PR #100, opened 2018-06-06
Added test for updateMachineConditions in machine.go

PadmaB:
This commit includes:
- change in updateMachineConditions method [in machine.go] to return error
- Test for udpateMachineConditions in machine.go

### DIFF
--- a/pkg/controller/machine.go
+++ b/pkg/controller/machine.go
@@ -248,7 +248,7 @@ func (c *controller) getMachineFromNode(nodeName string) (*v1alpha1.Machine, err
 }
 
 func (c *controller) updateMachineState(machine *v1alpha1.Machine, node *v1.Node) error {
-	machine = c.updateMachineConditions(machine, node.Status.Conditions)
+	machine, _ = c.updateMachineConditions(machine, node.Status.Conditions)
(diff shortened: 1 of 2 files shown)


pkg/controller/machine.go
```diff
@@ -248,7 +248,7 @@ func (c *controller) getMachineFromNode(nodeName string) (*v1alpha1.Machine, err
 }
 
 func (c *controller) updateMachineState(machine *v1alpha1.Machine, node *v1.Node) error {
-	machine = c.updateMachineConditions(machine, node.Status.Conditions)
+	machine, _ = c.updateMachineConditions(machine, node.Status.Conditions)
```
prashanth26: Should we print a warning if updateMachineCondition is returning an error?

PadmaB: Ideally, we should do proper error handling. Should we even proceed in case of errors, needs similar changes in other methods as well, thought of doing this together for other methods as well. Ok with you?

prashanth26: Sounds good.

PadmaB:
@prashanth26 Thanks !

PR #103, opened 2018-06-08
Added test for unfreezeMachineSetsAndDeployments in machine_safety.go

PadmaB:
PR includes:
1. check in machine_safety.go if machineDeployment exists for machineSet
2. test for unfreezeMachineSetsAndDeployments in machine_safety.go

### DIFF
--- a/pkg/controller/machine_safety.go
+++ b/pkg/controller/machine_safety.go
@@ -553,30 +553,33 @@ func (c *controller) unfreezeMachineSetsAndDeployments(machineSet *v1alpha1.Mach
 		glog.Warning("Updated failed, retrying - ", err)
 	}
 
-	machineDeployment := c.getMachineDeploymentsForMachineSet(machineSet)[0]
-	if machineDeployment != nil {
-		for {
-			// Get the latest version of the machineDeployment so that we can avoid conflicts
-			machineDeployment, err := c.controlMachineClient.MachineDeployments(machineDeployment.Namespace).Get(machineDeployment.Name, metav1.GetOptions{})
-			if err != nil {
-				// Some error occued while fetching object from API server
-				glog.Error(err)
-				break
+	machineDeployments := c.getMachineDeploymentsForMachineSet(machineSet)
+	if len(machineDeployments) >= 1 {
+		machineDeployment := machineDeployments[0]
+		if machineDeployment != nil {
+			for {
+				// Get the latest version of the machineDeployment so that we can avoid conflicts
+				machineDeployment, err := c.controlMachineClient.MachineDeployments(machineDeployment.Namespace).Get(machineDeployment.Name, metav1.GetOptions{})
+				if err != nil {
+					// Some error occued while fetching object from API server
+					glog.Error(err)
+					break
+				}
+				clone := machineDeployment.DeepCopy()
+				if clone.Labels == nil {
+					clone.Labels = make(map[string]string)
+				}
+				newStatus := clone.Status
+				RemoveMachineDeploymentCondition(&newStatus, v1alpha1.MachineDeploymentFrozen)
+				clone.Status = newStatus
+				delete(clone.Labels, "freeze")
+				_, err = c.controlMachineClient.MachineDeployments(clone.Namespace).Update(clone)
+				if err == nil {
+					break
+				}
+				// Keep retrying until update goes through
+				glog.Warning("Updated failed, retrying - ", err)
 			}
-			clone := machineDeployment.DeepCopy()
-			if clone.Labels == nil {
-				clone.Labels = make(map[string]string)
-			}
-			newStatus := clone.Status
-			RemoveMachineDeploymentCondition(&newStatus, v1alpha1.MachineDeploymentFrozen)
-			clone.Status = newStatus
-			delete(clone.Labels, "freeze")
-			_, err = c.controlMachineClient.MachineDeployments(clone.Namespace).Update(clone)
-			if err == nil {
-				break
-			}
-			// Keep retrying until update goes through
-			glog.Warning("Updated failed, retrying - ", err)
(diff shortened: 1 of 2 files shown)


pkg/controller/machine_safety.go
```diff
@@ -543,30 +543,33 @@ func (c *controller) unfreezeMachineSetsAndDeployments(machineSet *v1alpha1.Mach
...
-				clone.Labels = make(map[string]string)
-			}
-			newStatus := clone.Status
-			RemoveMachineDeploymentCondition(&newStatus, v1alpha1.MachineDeploymentFrozen)
-			clone.Status = newStatus
-			delete(clone.Labels, "freeze")
-			_, err = c.controlMachineClient.MachineDeployments(clone.Namespace).Update(clone)
-			if err == nil {
-				break
+	machineDeployments := c.getMachineDeploymentsForMachineSet(machineSet)
```
hardikdr: Wondering why do we need to replace the existing way of fetching the MachineDeployment?
Is there any rationale or is it just being considered as a good practice?

pkg/controller/machine_safety_test.go
```diff
@@ -71,6 +71,121 @@ var _ = Describe("machine", func() {
 		Entry("one machineset", nil, map[string]string{"machineset-1": ""}, nil, []string{"machineset-1"}, nil),
 	)
 
+	DescribeTable("##unfreezeMachineSetsAndDeployments",
```
hardikdr: Cant comment at the moment on the test. Would be really great, if we can have a session soon on how are we introducing the tests overall for the repo with details.

pkg/controller/machine_safety.go
PadmaB: The logic is simply enhanced with some null checks, it is a good practice. Either we enforce that MachineDeployment cannot be null for MachineSet or do a null check while fetching the same

pkg/controller/machine_safety_test.go
PadmaB: @amshuman-kr plans to give a session sometime this week. Do you  want to keep this PR open until then?

amshuman-kr: Scheduled for Thursday 11am.

pkg/controller/machine_safety.go
amshuman-kr: I don't see the existing way is changed. Only additional length and nil checks. The diff tool seems to be not up to the task to show actual change here :-)

pkg/controller/machine_safety_test.go
```diff
@@ -71,6 +71,121 @@ var _ = Describe("machine", func() {
...
+				return &controller{
+					controlMachineClient:    fakeTypedClient,
+					machineSetLister:        machineSets.Lister(),
+					machineDeploymentLister: machineDeployments.Lister(),
+					machineSetSynced:        machineSets.Informer().HasSynced,
+					machineDeploymentSynced: machineDeployments.Informer().HasSynced,
+				}
+			}
+
+			testMachineSet := &machinev1.MachineSet{
```
amshuman-kr: It might be better to enhance the existing newMachineSet function to handle this.

I will do this. You can rebase after that, perhaps?

pkg/controller/machine_safety_test.go
```diff
@@ -71,6 +71,121 @@ var _ = Describe("machine", func() {
...
+				},
+			}
+			if machineSetIsFrozen {
+				testMachineSet.Labels["freeze"] = "True"
+				msStatus := testMachineSet.Status
+				mscond := NewMachineSetCondition(machinev1.MachineSetFrozen, machinev1.ConditionTrue, "testing", "freezing the machineset")
+				SetCondition(&msStatus, mscond)
+			}
+
+			testMachineDeployment := &machinev1.MachineDeployment{
```
amshuman-kr: It might be better to add a newMachineDeployment function like the existing newMachineSet and integrate it with newIntialObjects.

I will do this and you can maybe rebase after that.

pkg/controller/machine_safety_test.go
```diff
@@ -71,6 +71,121 @@ var _ = Describe("machine", func() {
...
+				},
+				Spec: machinev1.MachineDeploymentSpec{
+					Selector: &metav1.LabelSelector{
+						MatchLabels: map[string]string{
+							"name": "testMachineDeployment",
+						},
+					},
+				},
+			}
+			if parentIsFrozen {
```
amshuman-kr: Better to handle this via initialMachineDeployments, I think.

pkg/controller/machine_safety_test.go
PadmaB: Thought it will be too generic, but will wait for your changes and then adapt accordingly

pkg/controller/machine_safety_test.go
PadmaB: same as above

pkg/controller/machine_safety_test.go
amshuman-kr: @PadmaB I have committed the change https://github.com/amshuman-kr/machine-controller-manager/commit/bf1ccb4a4cdd89ea8896457c4aeabda7a216503c to complete the helper function to initialise objects.

But I will create the PR only tomorrow after I have added the full round-trip test.

hardikdr:
Is there anything else to be discussed for this PR to be merged ?
@amshuman-kr @PadmaB

amshuman-kr:
LGTM.

@PadmaB do you have any pending changes?

PadmaB:
Ready to be merged. I have adapted the relevant changes from the testing framework, though not fully followed the conventions. Open for feedback, else we can merge and if required going forward.

PadmaB:
Thanks!

Issue #69, opened 2018-04-20
AWS: Missing `iops` when using `io1 volume type


rfranzke:
When creating a machine with volume type `io1` the MCM does not specify the `iops` parameter, resulting in the following error:
`I0420 06:21:27.447052       1 machine.go:301] Error while creating machine <machineName>: InvalidParameterCombination: The parameter iops must be specified for io1 volumes.`

prashanth26:
[This PR](Handling IOPs when creating volume of type AWS io1) closes this issue

vlerenc:
I believe, `io1` wasn't reenabled as it is still missing in cloud profile and dashboard, right? Change https://github.wdf.sap.corp/kubernetes/landscape-setup/commit/3bb8c9295732d784265218e2fb1033c8a0365454#diff-7f6d64702b8513c2de9ab0737dbbd00f needs to be reverted to get io1 back again.

cc @rfranzke @freegroup

rfranzke:
We do not expose `iops` field to the Shoot manifest. We cannot simply revert it but need adaption in Gardener and Shoot API spec to support `io1` volumes.

vlerenc:
Then we should reopen this ticket, I guess.

PadmaB:
Necessary changes has been made in MCM, changes are required from Gardener and Dashboard.
**Ref:** https://github.com/gardener/gardener/issues/726 and https://github.com/gardener/dashboard/issues/306

PR #105, opened 2018-06-13
Handling IOPs field when creating volume of type AWS io1

ggaurav10:
Fix for issue #69

### DIFF
--- a/pkg/apis/machine/validation/awsmachineclass.go
+++ b/pkg/apis/machine/validation/awsmachineclass.go
@@ -119,13 +119,12 @@ func validateBlockDevices(blockDevices []machine.AWSBlockDeviceMappingSpec, fldP
 	if len(blockDevices) > 1 {
 		allErrs = append(allErrs, field.Required(fldPath.Child(""), "Can only specify one (root) block device"))
 	} else if len(blockDevices) == 1 {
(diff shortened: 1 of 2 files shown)


pkg/driver/driver_aws.go
```diff
@@ -77,6 +77,9 @@ func (d *AWSDriver) Create() (string, string, error) {
 			VolumeType: &volumeType,
 		},
 	}
+	if volumeType == "io1" {
```
prashanth26: Where is the validation for the same?

pkg/apis/machine/validation/awsmachineclass.go
```diff
@@ -125,6 +125,9 @@ func validateBlockDevices(blockDevices []machine.AWSBlockDeviceMappingSpec, fldP
 		if blockDevices[0].Ebs.VolumeType == "" {
 			allErrs = append(allErrs, field.Required(fldPath.Child("ebs.volumeType"), "Please mention a valid ebs volume type"))
 		}
+		if blockDevices[0].Ebs.VolumeType == "io1" && blockDevices[0].Ebs.Iops == 0 {
```
prashanth26: Can we change this to an else if block? Also could we change the validation iops and size to <= 0 instead of == 0

PR #109, opened 2018-06-19
MachineSet updates status even when deletionTimestamp is set

prashanth26:


prashanth26:
@hardikdr and @ggaurav10 kindly review this PR.

pkg/controller/machineset.go
```diff
@@ -493,35 +493,37 @@ func (c *controller) reconcileClusterMachineSet(key string) error {
...
 	}
-	//glog.V(2).Infof("Print manageReplicasErr: %v ",manageReplicasErr) //Remove
 
 	machineSet = machineSet.DeepCopy()
 	newStatus := calculateMachineSetStatus(machineSet, filteredMachines, manageReplicasErr)
 
 	// Always updates status as machines come up or die.
 	updatedMachineSet, err := updateMachineSetStatus(c.controlMachineClient, machineSet, newStatus)
-	if err != nil {
+	if err != nil && !apierrors.IsNotFound(err) {
```
ggaurav10: It will be good to log the error `IsNotFound{}` even if we don't want to return it.

pkg/controller/machineset.go
```diff
@@ -493,35 +493,37 @@ func (c *controller) reconcileClusterMachineSet(key string) error {
...
 		// Multiple things could lead to this update failing. Requeuing the machine set ensures
 		// Returning an error causes a requeue without forcing a hotloop
-		glog.V(2).Infof("update machine failed with: %v", err) //Remove
+		glog.Errorf("Update machineSet %s failed with: %s", machineSet.Name, err)
 		return err
 	}
```
ggaurav10: The next lines in the code doesn't check if `updatedMachineSet` is `nil` but still references using it. This was not an issue earlier as in that case err will always be non-nil, and we `return`ed. But now, we should be checking if this was `nil`

pkg/controller/machineset.go
prashanth26: This log occurs often (object not found), and logging this would flood the logs with unnecessary statements?

ggaurav10: Then may be in some higher log level? so that it will not be printed by default.

pkg/controller/machineset.go
prashanth26: Okay, you are right. Let me fix that.

pkg/controller/machineset.go
```diff
@@ -521,7 +523,9 @@ func (c *controller) reconcileClusterMachineSet(key string) error {
 	if err != nil {
 		// Multiple things could lead to this update failing. Requeuing the machine set ensures
 		// Returning an error causes a requeue without forcing a hotloop
-		glog.V(2).Infof("update machine failed with: %v", err) //Remove
+		if !apierrors.IsNotFound(err) {
```
prashanth26: @ggaurav10 better now?

ggaurav10: yes

PR #111, opened 2018-06-25
Unit tests for checkMachineTimeout

prashanth26:
- Updated machine types
- Regenerated machine clients
- Logic for machine joining or re-joining is updated
- Added test cases for checkMachineTimeout


pkg/apis/machine/types.go
```diff
@@ -213,6 +213,24 @@ const (
...
+
+	// MachineLastOperationUpdate indicates that the last operation was an update
+	MachineLastOperationUpdate MachineLastOperationType = "Update"
+
+	// MachineLastOperationHealthCheck indicates that the last operation was a create
+	MachineLastOperationHealthCheck MachineLastOperationType = "HealthCheck"
+
+	// MachineLastOperationDelete indicates that the last operation was a create
+	MachineLastOperationDelete MachineLastOperationType = "Delete"
+)
```
ggaurav10: I think we should rename these to something more generic -Eg. `MachineOperationCreate`
Multiple structs will be able to use it in that case, and not just `LastOperation`

hardikdr: Isn't HealthCheck is something we would do[if not now] periodically anyways, and should not be part of the MachineLastOperationType struct? Just checking if semanticcs really match.

pkg/apis/machine/types.go
```diff
@@ -335,6 +353,25 @@ type MachineSetStatus struct {
...
 	// ObservedGeneration
 	ObservedGeneration int64
+
+	// FailedMachines has summary of machines on which lastOperation Failed
+	FailedMachines *[]MachineSummary
+}
+
+//MachineSummary store the summary of machine.
+type MachineSummary struct {
+	// +optional
```
hardikdr: shouldn't this be compulsory?

pkg/apis/machine/types.go
```diff
@@ -335,6 +353,25 @@ type MachineSetStatus struct {
...
+	Name string
+
+	// ProviderID represents the provider's unique ID given to a machine
+	// +optional
+	ProviderID string
```
hardikdr: Shouldn't this also be compulsory ?

pkg/apis/machine/types.go
```diff
@@ -335,6 +353,25 @@ type MachineSetStatus struct {
 
 	// ObservedGeneration
 	ObservedGeneration int64
+
+	// FailedMachines has summary of machines on which lastOperation Failed
```
hardikdr: These fields make more sense if they are optional. Can you please align it with upstream changes in this PR.
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/extensions/types.go#L725

pkg/apis/machine/v1alpha1/types.go
```diff
@@ -211,6 +211,24 @@ const (
...
+// These are the valid statuses of machines.
+const (
+	// MachineLastOperationCreate indicates that the last operation was a create
+	MachineLastOperationCreate MachineLastOperationType = "Create"
+
+	// MachineLastOperationUpdate indicates that the last operation was an update
+	MachineLastOperationUpdate MachineLastOperationType = "Update"
+
+	// MachineLastOperationHealthCheck indicates that the last operation was a create
+	MachineLastOperationHealthCheck MachineLastOperationType = "HealthCheck"
```
hardikdr: See comment above for health-check.

pkg/apis/machine/types.go
```diff
@@ -171,7 +171,7 @@ type LastOperation struct {
 	State MachineState
 
 	// Type of operation
-	Type string
+	Type MachineLastOperationType
```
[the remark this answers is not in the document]
prashanth26: There were many such string labels with similar names. Hence I am sticking to a more detailed one which also talks about its parent's structure. Refer the machine/types.go for other similar types.

pkg/apis/machine/types.go
prashanth26: @ggaurav10 I don't think there are any other structs making use of the key word create. Correct me if I am wrong.

@hardikdr - Yes, its true that HealthCheck operation is done periodically. But Here we are updating the lastOperation only when the machine state is going to unknown, I thought it would make sense to keep in the last operation status as it would give us an idea when an machine just went unhealthy. And moreover lastOperation refers to the last operation taken by the controller right? and in this case controller updated the health status. And I thought this is something that the object must depict?

pkg/apis/machine/types.go
prashanth26: Okay shall do that.

pkg/apis/machine/types.go
prashanth26: Shall make this changes.

pkg/apis/machine/types.go
prashanth26: Shall make this changes.

pkg/apis/machine/v1alpha1/types.go
prashanth26: Same answer as above. But if it doesn't make sense, we still could revert the changes.

pkg/apis/machine/types.go
ggaurav10: I meant from future extensibility point of view. This `type` is made generic, it can be used elsewhere also.

pkg/apis/machine/types.go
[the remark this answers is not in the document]
ggaurav10: If we make it generic, its use will not be limited only to `lastOperation`.

pkg/apis/machine/types.go
prashanth26: As suggested by @ggaurav10. I have updated `MachineLastOperation` variables to `MachineOperation` variables.

pkg/apis/machine/types.go
prashanth26: Done.

pkg/apis/machine/types.go
prashanth26: Done.

prashanth26:
Made suggested changes. Kindly review when you get time.

pkg/apis/machine/types.go
prashanth26: Done.

Issue #26, opened 2018-02-10
Reserve Excess Cluster Capacity


vlerenc:
# Story
* As user I want to be able to schedule my workload (if of reasonable size) in my shoot cluster in seconds, even if all nodes of the shoot cluster are busy, so that I am more productive.
* As operator I want to be able to create new shoot cluster control planes in the seed clusters in seconds, even if all nodes of the seed cluster are busy, so that I am more productive.

# Motivation
The [cluster autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) seems to react only when pods can't be scheduled on any of the current nodes due to insufficient resources. That means, it takes minutes after the event until the pod can be scheduled.

# Acceptance Criteria
- [ ] We run the machine deployment with excess capacity (in the given limits), a configurable combination of an absolute excess capacity minimum and a percentage, e.g. (calculation here in machines, but may be alternatively based on CPU/Memory, if that's easier/better): max(1, 5% of the current number of machines)

# Implementation Proposal
This could be done in the machine controller manager, but if collides with the scale-down strategy in the [cluster autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler), consider whether a contribution would be possible or do it in the fork (#25).

# Remarks
Rafael Franzke commented: There is an proposal for allowing exactly the above scenario (reserving cluster resources by cluster-autoscaler) which is planned to be release with Kubernetes v1.8 (according to the 1.8 feature tracking spreadsheet). See this for more information: https://github.com/kubernetes/autoscaler/pull/77/files

# Definition of Done
- [ ] Knowledge is distributed: Have you spread your knowledge in pair programming/code review?
- [ ] Unit Tests are provided: Have you written automated unit tests or added manual NGPTT tickets?
- [ ] Integration Tests are provided: Have you written automated integration tests?
- [ ] Minimum API exposure: If you have added public API, was it really necessary/is it minimal?
- [ ] Operations guide: Have you updated the [operations guide](https://github.wdf.sap.corp/kubernetes/kube-docs/wiki/Operations%20Guide)?
- [ ] User documentation: Have you informed Sylvia Pick about user relevant changes?

vlerenc:
[...]

PR #113, opened 2018-06-25
[WIP] Proposal for excess reserve capacity

hardikdr:
This is a proposal to address the issue: https://github.com/gardener/machine-controller-manager/issues/26

### DIFF
--- a/docs/proposals/excess_reserve_capacity.md
+++ b/docs/proposals/excess_reserve_capacity.md
@@ -1,0 +1,55 @@
+# Excess Reserve Capacity - WIP
+
+## Goal:
+Currently, autoscaler optimizes the number of machines for a given application-workload. Along with effective resource utilization, this feature brings concern where, many times, when new application instances are created - they don't find space in existing cluster. This leads the cluster-autoscaler to create new machines via MachineDeployment, which can take from 3-4 minutes to ~10 minutes, for the machine to really come-up and join the cluster. In turn, application-instances have to wait till new machines join the cluster.
+
+One of the promising solutions to this issue is Excess Reserve Capacity. Idea is to keep a certain number of machines or percent of resources[cpu/memory] always available, so that new workload, in general, can be scheduled immediately unless huge spike in the workload. Also, the user should be given enough flexibility to choose how many resources or how many machines should be kept alive and non-utilized as this affects the Cost directly.
+
+## Note:
+* We decided to go with Approach-4 which is based on low priority pods. Please find more details here: https://github.com/gardener/gardener/issues/254
+* Approach-3 looks more promising in long term, we may decide to adopt that in future based on developments/contributions in autoscaler-community. 
+
+## Possible Approaches:
+Following are the possible approaches, we could think of so far.
+
+### Approach 1: Enhance Machine-controller-manager to also entertain the excess machines.
+* Machine-controller-manager currently takes care of the machines in the shoot cluster starting from creation-deletion-health check to efficient rolling-update of the machines. From the architecture point of view, MachineSet makes sure that X number of machines are always **running and healthy**. MachineDeployment controller smartly uses this facility to perform rolling-updates.
+
+* We can expand the scope of MachineDeployment controller to maintain excess number of machines by introducing new parallel independent controller named *MachineTaint* controller. This will result in MCM to include Machine, MachineSet, MachineDeployment, MachienSafety, MachineTaint controllers. MachineTaint controller does not need to introduce any new CRD - analogy fits where taint-controller also resides into kube-controller-manager.
+
+* Only Job of MachineTaint controller will be:
+    * List all the Machines under each MachineDeployment.
+    * Maintain taints of [*noSchedule* and *noExecute*](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/) on `X` latest MachineObjects.
+        * There should be an event-based informer mechanism where MachineTaintController gets to know about any Update/Delete/Create event of MachineObjects - in turn, maintains the *noSchedule* and *noExecute* taints on all the *latest* machines.
+        * Why latest machines?
+            * Whenever autoscaler decides to add new machines - essentially ScaleUp event - taints from the older machines are removed and newer machines get the taints. This way X number of Machines immediately becomes free for new pods to be scheduled.
+            * While ScaleDown event, autoscaler specifically mentions which machines should be deleted, and that should not bring any concerns. Though we will have to put proper label/annotation defined by autoscaler on taintedMachines, so that autoscaler does not consider the taintedMachines for deletion while scale-down.
+                * Annotation on tainted node: `"cluster-autoscaler.kubernetes.io/scale-down-disabled": "true"`
+* Implementation Details:
+    * Expect new **optional field** *ExcessReplicas* in `MachineDeployment.Spec`. MachineDeployment controller now adds both `Spec.Replicas` and `Spec.ExcessReplicas`[if provided], and considers that as a standard desiredReplicas.
+        * Current working of MCM will not be affected if ExcessReplicas field is kept nil.
+    * MachineController currently reads the *NodeObject* and sets the MachineConditions in MachineObject. Machine-controller will now also read the taints/labels from the MachineObject - and maintains it on the *NodeObject*.
+
+* We expect cluster-autoscaler to intelligently make use of the provided feature from MCM.
+    * CA gets the input of *min:max:excess* from Gardener. CA continues to set the `MachineDeployment.Spec.Replicas` as usual based on the application-workload.
+    * In addition, CA also sets the `MachieDeployment.Spec.ExcessReplicas` .
+    * Corner-case:
+        * CA should decrement the excessReplicas field accordingly when *desiredReplicas+excessReplicas* on MachineDeployment goes beyond *max*.
+
+
+### Approach 2: Enhance Cluster-autoscaler by simulating fake pods in it.
+* There was already an attempt by community to support this feature.
+    * Refer for details: https://github.com/kubernetes/autoscaler/pull/77/files
+
+### Approach 3: Enhance cluster-autoscaler to support pluggable scaling-events.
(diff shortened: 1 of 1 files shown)


proposals/excess_reserve_capacity.md
```diff
@@ -0,0 +1,49 @@
...
+    * Corner-case:
+        * CA should decrement the excessReplicas field accordingly when *desiredReplicas+excessReplicas* on MachineDeployment goes beyond *max*.
+
+
+### Approach 2: Enhance Cluster-autoscaler by simulating fake pods in it.
+* There was already an attempt by community to support this feature.
+    * Refer for details: https://github.com/kubernetes/autoscaler/pull/77/files
+
+### Approach 3: Enhance cluster-autoscaler to support pluggable scaling-algorithms.
+* Forked version of cluster-autoscaler could be improved to plug-in the algorithm for excess-reserve capacity.
```
amshuman-kr: @hardikdr It would be nice if you could add a bit more of the details we discussed for this approach last Friday. Things such as using golang channel to separate the algorithms to trigger scaling  (hard-coded in cluster-autoscaler, currently) from the algorithms about how to to achieve the scaling (already pluggable in cluster-autoscaler). This kind of separation can help us introduce/plug-in new algorithms (such as based node resource utilisation) without affecting existing code-base too much while almost completely re-using the code-base for the actual scaling.

Also, as discussed, this approach is not specific to our fork of cluster-autoscaler. It can be made upstream eventually as well.

BTW this approach was about pluggable algorithms to trigger scaling events and not scaling algorithms as such for which we can use existing algorithms in cluster-autoscaler.

proposals/excess_reserve_capacity.md
```diff
@@ -0,0 +1,49 @@
...
+    * MachineController currently reads the *NodeObject* and sets the MachineConditions in MachineObject. Machine-controller will now also read the taints/labels from the MachineObject - and maintains it on the *NodeObject*.
+
+* We expect cluster-autoscaler to intelligently make use of the provided feature from MCM.
+    * CA gets the input of *min:max:excess* from Gardener. CA continues to set the `MachineDeployment.Spec.Replicas` as usual based on the application-workload.
+    * In addition, CA also sets the `MachieDeployment.Spec.ExcessReplicas` .
+    * Corner-case:
+        * CA should decrement the excessReplicas field accordingly when *desiredReplicas+excessReplicas* on MachineDeployment goes beyond *max*.
+
+
+### Approach 2: Enhance Cluster-autoscaler by simulating fake pods in it.
```
amshuman-kr: To me, approach 2 is a no go.
1. Because fakepods looks like a bad hack.
2. The PR was closed due to no response from the author after lengthy discussion on this topic in the community.

proposals/excess_reserve_capacity.md
```diff
@@ -0,0 +1,49 @@
...
+* There was already an attempt by community to support this feature.
+    * Refer for details: https://github.com/kubernetes/autoscaler/pull/77/files
+
+### Approach 3: Enhance cluster-autoscaler to support pluggable scaling-algorithms.
+* Forked version of cluster-autoscaler could be improved to plug-in the algorithm for excess-reserve capacity.
+* Needs further discussion around upstream support.
+
+### Approach 4: Make intelligent use of Low-priority pods.
```
amshuman-kr: Approach 4 is also a hack like approach 2. But at least it has a chance to work because it is not dependent on the closed PR.

But this would be workable only as a short-term solution only for seed clusters. For shoot clusters we need a more general solution.

proposals/excess_reserve_capacity.md
```diff
@@ -0,0 +1,49 @@
...
+* Implementation Details:
+    * Expect new **optional field** *ExcessReplicas* in `MachineDeployment.Spec`. MachineDeployment controller now adds both `Spec.Replicas` and `Spec.ExcessReplicas`[if provided], and considers that as a standard desiredReplicas.
+        * Current working of MCM will not be affected if ExcessReplicas field is kept nil.
+    * MachineController currently reads the *NodeObject* and sets the MachineConditions in MachineObject. Machine-controller will now also read the taints/labels from the MachineObject - and maintains it on the *NodeObject*.
+
+* We expect cluster-autoscaler to intelligently make use of the provided feature from MCM.
+    * CA gets the input of *min:max:excess* from Gardener. CA continues to set the `MachineDeployment.Spec.Replicas` as usual based on the application-workload.
+    * In addition, CA also sets the `MachieDeployment.Spec.ExcessReplicas` .
+    * Corner-case:
+        * CA should decrement the excessReplicas field accordingly when *desiredReplicas+excessReplicas* on MachineDeployment goes beyond *max*.
```
amshuman-kr: Approach 1 looks workable.

However, one disadvantage I can see with this approach is that the granularity for excess capacity can only be a single machine. It cannot easily be made more generic to consider more fine-grained resource-based (cpu, memory, disk etc.) excess capacity.

proposals/excess_reserve_capacity.md
hardikdr: Hi Amshu, I do not yet have complete clarity on this approach, and wanted to avoid bugs in the proposal due to my misunderstanding- so kept it open-WIP.
It would be really helpful if you could put-down the exact details you would like me to add in the proposal or let's discuss it together tomorrow offline and i would then add the details in the proposal. I hope that's fine.

proposals/excess_reserve_capacity.md
ggaurav10: One minor disadvantage of this approach would also be that it will put a restriction on users to schedule their own low(-est) priority pods because they will compete with these pods.

proposals/excess_reserve_capacity.md
amshuman-kr: Sure. We can update the proposal later.

amshuman-kr: I was thinking about refactoring https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/core/static_autoscaler.go#L122 to introduce a separation (using golang channels?) of scaling events from scaling algorithms.

proposals/excess_reserve_capacity.md
prashanth26: To me the excess capacity logic, should ideally be taken care of by the cluster autoscaler. Say even we implement this logic, and eventually the autoscaler would come up with algorithms to do similar things as excess reserve capacity - then this would create problems where two entities now would try to handle excess capacity. And then we would have to revert this change in the MCM's core logic again.

Although approach 1 sounds workable, it looks more like an temporary hack to me and I am not quite sure if it's the right way to go.

vlerenc: I understand your point @prashanth, but don't know how to otherwise reach the goal. As for yours @amshu I thought the same, but then again the granularity is always a machine anyways.

I have a few questions:
1. How is the initial state (no load) handled? Are you starting with #min-#excess untainted and #excess tainted machines? I mean, we'd like to see a new machine the moment the load + desired excess reaches the overall (min) capacity.
2. In essence, the idea is to untaint a tainted machine the moment the cluster autoscaler requests a new one (fast) and start provisioning a new machine that will be forcefully tainted when available (slow). If even more machines are requested (no untained machines available), new machines will be provisioned, but not tainted on startup, right?
3. How will this work out "Though we will have to put proper label/annotation defined by autoscaler on taintedMachines, so that autoscaler does not consider the taintedMachines for deletion while scale-down."?

hardikdr: As discussed over the call, we may utilize the PodPriority feature to reserve excess capacity in cluster. 
- After an initial investigation, PodPriority seems smooth enough to satisfy our use-case.

Though answering the questions to clarify the proposed approach:
 -> 1: Initially min+excess[tainted] would have been available. As we always want to maintain the min number of machines as specified by autoscaler.
-> 2: Yes, that is correct. Though I would further mention once the number of machines approaches MAX, we would reduce the tainted machines accordingly - eventually, cluster will have the max number of machines if required based on load- with no tainted machines.
-> 3: Autoscaler itself provides feature- where we can put a predefined annotation on the Node-object which will be excluded from the scale-down.
Hope that answers the queries :)

vlerenc: Hmm... wasn't your idea to use low prio pods and move the topic into the seed cluster bootstrapper? Then we wouldn't have to do anything in the MCM.

hardikdr: Yes, PodPriority/Low-priority-pods is still the preferred approach.

I felt like to answer the doubt for sake of answering related to other approach which was based on changes in MCM - though we would be not adapting this one. Sorry for the confusion :)

amshuman-kr: I am confused too. Isn't PodPriority about using low priority pods to reserve capacity for additional control planes?

If so, then why do we need to worry about excess and tainted machines?

vlerenc: Ah, thanks Hardik. You actually also wrote it: you just wanted to give answers to the questions raised to complete the discussion.

Then I take...

> After an initial investigation, PodPriority seems smooth enough to satisfy our use-case.

...as statement, that you tried it out and we can move the topic to the seed cluster bootstrapper and close this ticket as `won't-fix` for the time being?

Can you please elaborate what "seems smooth enough" mean? How fast do the low-prio pods get evicted and the new/actual ones get scheduled?

vlerenc:
What should happen with this PR now that the topic will be handled in the Gardener?

vlerenc:
Well, we could put (merge) it into a folder structure that holds not only implemented, but also discussed/proposed (that's what the folder says today) extensions. This way, the content and recommendation wouldn't be lost. Normally, I would delete unused code (since it's anyway in git), but if we never merge the PR, it's basically lost for good (in GitHub with nobody to ever find again).

amshuman-kr:
Merging the PR works too. But I think we should add the current thinking and preferences in the proposal before merging it.

hardikdr:
Sure, sounds good !! 
I will add necessary comments in the PR, from our discussions then we can merge it.

PR #116, opened 2018-06-28
Liveliness endpoint

### LINKED ISSUE #50: Add /healthz and /liveliness endpoint
Kubelet should know if there is a problem with the MCM and restart the Pod proactively. [Helper library](https://github.com/heptiolabs/healthcheck).

prashanth26:
The following PR replaces the existing health endpoint to make it more useable.

Closes #50.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -36,6 +36,7 @@ import (
 	corelisters "k8s.io/client-go/listers/core/v1"
 
 	machinescheme "github.com/gardener/machine-controller-manager/pkg/client/clientset/versioned/scheme"
+	"github.com/gardener/machine-controller-manager/pkg/handlers"
 	"github.com/gardener/machine-controller-manager/pkg/options"
 	runtimeutil "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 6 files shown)


pkg/controller/controller.go
```diff
@@ -426,7 +430,8 @@ func (c *controller) Run(workers int, stopCh <-chan struct{}) {
 	}
 
 	<-stopCh
-	glog.Info("Shutting down Machine Controller Manager ")
+	glog.V(1).Info("Shutting down Machine Controller Manager ")
+	handlers.UpdateHealth(false)
```
ggaurav10: Is it possible that just after health status is updated here, liveness probe fails, and the container is restarted before all the `defer` statements execute.

prashanth26: The container takes 30s before restarting, and this is a very rare scenario to get to this as, the controller doesn't actually terminate by itself. So i think this should be fine.

prashanth26:
@hardikdr The PR is reviewable now. You can have a look at it.

pkg/controller/controller.go
ggaurav10: I am just afraid that this will introduce some indeterministic behavior unnecessarily, however rare it (killing of container when shutting down because of health failure) might be.
Approving the PR.

prashanth26:
@ggaurav10 Thank you 👍

PR #124, opened 2018-07-23
Problem: can't add non image disk on GCP

spheromak:
Solution: Only the root disk on GCP has an image requirement. You can
attach other block storage without a source image, and this change
allows for that.


hardikdr:
Thanks for the PR.
Generally we expect that application instance would request a non-boot disk via PVC/PV mechanism and MCM would not create any in-parallel to boot-disk.
I am curious if you have any specific use-case for this.

Though PR looks good to me, just a side-note would be that extra non-boot disks could potentially become orphan if VM is lost and `AutoDelete` is not set to `true`. Essentially there is no way for MCM/Kubernetes to garbage-collect those Disks due to lack of metadata.

PR #125, opened 2018-07-23
Problem: Can't assign static addresses on GCP

spheromak:
Solution: Allow the user to specify 'createStaticExternal',
'createStaticInternal' to have the driver make new static addresses and
attach them when making an interface. Also allows for attaching a
specific static internal/external via  'attachInternal' and
'attachExternal' keys on an interface object. Where the value should be
the internal or external IPv4 address. At this time the driver doesn't
handle removal of static addresses, and it is assumed the user would
handle that via other means.

### DIFF
--- a/pkg/apis/machine/v1alpha1/types.go
+++ b/pkg/apis/machine/v1alpha1/types.go
@@ -1101,6 +1101,20 @@ type GCPMetadata struct {
 type GCPNetworkInterface struct {
 	Network    string `json:"network,omitempty"`
 	Subnetwork string `json:"subnetwork,omitempty"`
+
+	// CreateStaticExternal creates a new IP ant ensures this node gets attached to it.
+	// If this is specified you can't use AttachExternal
+	CreateStaticExternal bool `json:"createStaticExternal,omitemptu"`
+
+	// CreateStaticInternal creates a new IP ant ensures this node gets attached to it.
+	// If this is specified you can't use AttachInternal
+	CreateStaticInternal bool `json:"createStaticInternal,omitemptu"`
+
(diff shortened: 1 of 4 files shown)


pkg/apis/machine/v1alpha1/types.go
```diff
@@ -1101,6 +1101,20 @@ type GCPMetadata struct {
 type GCPNetworkInterface struct {
 	Network    string `json:"network,omitempty"`
 	Subnetwork string `json:"subnetwork,omitempty"`
+
+	// CreateStaticExternal creates a new IP ant ensures this node gets attached to it.
```
msohn: typo: s/ant/and/

pkg/driver/driver_gcp.go
```diff
@@ -116,14 +154,43 @@ func (d *GCPDriver) Create() (string, string, error) {
...
 			AccessConfigs: []*compute.AccessConfig{{}},
 		}
-		if len(nic.Network) != 0 {
+
+		if nic.CreateStaticExternal {
+			adr, err := d.createStaticAddress(ctx, computeService, project, gcpExternalAddress)
+			if err != nil {
+				return "Error", "Error", err
+			}
+			// currently theres only ever one access config on the nic object even tho it's a slice
```
msohn: s/theres/there is/

prashanth26:
Hi @spheromak ,
                            Would you be addressing the review comments?

Thanks & Regards,
Prashanth

spheromak:
sorry we have gone a different way at my job, and I don't intend to carry this work forward.

PR #136, opened 2018-08-12
Add support for Alibaba Cloud

lu-zhenyu:
With this initial version, one can create or delete virtual machines in the Alicloud.

**Release note**:

```noteworthy user
Integrated support for Ali cloud in MCM
```


pkg/apis/machine/types.go
```diff
@@ -1117,3 +1117,75 @@ type GCPServiceAccount struct {
...
+
+// AlicloudSystemDisk describes SystemDisk for Alicloud.
+type AlicloudSystemDisk struct {
+	Category string
+	Size     int
+}
+
+// AlicloudTags describes Tags for Alicloud.
+type AlicloudTags struct {
+	Tag1Key   string
```
prashanth26: Would it make more sense to have a list of tags rather than having one for each variable?

prashanth26: Would it make more sense to have a list of tags rather than having one for each variable?

lu-zhenyu: We use this format because Alicloud doesn't support nested tags in its API (refer to https://help.aliyun.com/document_detail/63440.html, and search key word "Tag.n.Key"). We can make it as a list, but need extra logic of converting formats and validating list length.

prashanth26: Hi @lu-zhenyu ,
The reason I am asking you to switch to list is bcoz we are trying to align all our machine classes into a single one, and when that happens we will need a list. Hence maybe you can read as a list and convert it while passing it to the SDK?

Refer - https://github.com/gardener/machine-controller-manager/blob/master/pkg/apis/machine/v1alpha1/types.go#L773, https://github.com/gardener/machine-controller-manager/blob/master/kubernetes/aws-machine-class.yaml#L19-L23

pkg/driver/driver_alicloud.go
```diff
@@ -0,0 +1,287 @@
...
+	Tag1Key                 string           `position:"Query" name:"Tag.1.Key"`
+	Tag2Key                 string           `position:"Query" name:"Tag.2.Key"`
+	Tag3Key                 string           `position:"Query" name:"Tag.3.Key"`
+	Tag4Key                 string           `position:"Query" name:"Tag.4.Key"`
+	Tag5Key                 string           `position:"Query" name:"Tag.5.Key"`
+	Tag1Value               string           `position:"Query" name:"Tag.1.Value"`
+	Tag2Value               string           `position:"Query" name:"Tag.2.Value"`
+	Tag3Value               string           `position:"Query" name:"Tag.3.Value"`
+	Tag4Value               string           `position:"Query" name:"Tag.4.Value"`
+	Tag5Value               string           `position:"Query" name:"Tag.5.Value"`
```
hardikdr: Can we please have an array of tags - as mentioned in the previous comment ? Looks like this will prevent us from having more than 5 tags.

pkg/driver/driver_alicloud.go
```diff
@@ -0,0 +1,287 @@
...
+func (c *AlicloudDriver) GetExisting() (string, error) {
+	return c.MachineID, nil
+}
+
+func (c *AlicloudDriver) getVMDetails(machineID string) ([]ecs.Instance, error) {
+	searchClusterName := ""
+	searchNodeRole := ""
+
+	if c.AlicloudMachineClass.Spec.Tags != nil {
+		if strings.Contains(c.AlicloudMachineClass.Spec.Tags.Tag1Key, "kubernetes.io/cluster/") {
```
hardikdr: Can we please check if any of the tags contains `kubernetes.io/cluster`, current implementation will prevent us from not having it at first position.
https://github.com/gardener/machine-controller-manager/blob/master/pkg/driver/driver_aws.go#L194-L200

hardikdr:
@lu-zhenyu I think it make sense to merge this PR before we have reliable-implementation for out-of-tree support for cloud-providers. Migration-note from in-tree to out-tree would remain same for cloud-providers, and should not be a large effort. 
Merging this PR will help colleagues using this change to avoid creating a parallel branch and keep-resyncing with the master-branch.

pkg/driver/driver_alicloud.go
lu-zhenyu: noted, will change.

pkg/apis/machine/types.go
lu-zhenyu: noted, will do.

kubernetes/alicloud-machine-class.yaml
```diff
@@ -0,0 +1,33 @@
...
+  systemDisk:
+    category: cloud_efficiency # cloud, cloud_efficiency, cloud_ssd, ephemeral_ssd
+    size: 30 # 20-500
+  instanceChargeType: PostPaid # Prepaid or PostPaid (default)
+  internetChargeType: PayByTraffic # PayByBandwidth or PayByTraffic (default)
+  internetMaxBandwidthIn: 5 # 1-200
+  internetMaxBandwidthOut: 0 # 0-100
+  spotStrategy: NoSpot # NoSpot, SpotWithPriceLimit, SpotAsPriceGo
+  tags:
+    tag1Key: kubernetes.io/cluster/**** # This is mandatory at first tag as the safety controller uses this tag to identify VMs created by this controller. Replace **** string with your desired cluster name.
```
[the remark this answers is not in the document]
lu-zhenyu: resolved in 0104d86

hardikdr:
@lu-zhenyu  Please let us know when all the changes are done and we should review again.
@prashanth26  Do you have any more inputs for this PR?

lu-zhenyu:
@hardikdr, my part is done. You can review now.

lu-zhenyu:
Can't login in concourse-ci. Who can help on this?

Issue #144, opened 2018-08-31
Alicloud SDK issue on failing to get TimeZone info


jia-jerry:
When apply a machin deploy an Alicloud Machine class, the program will get a panic inside the Docker container.

panic: open /usr/local/go/lib/time/zoneinfo.zip: no such file or directory [recovered]
	panic: open /usr/local/go/lib/time/zoneinfo.zip: no such file or directory

goroutine 377 [running]:
github.com/gardener/machine-controller-manager/vendor/k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/Users/i041966/projects/go-root/src/github.com/gardener/machine-controller-manager/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:58 +0x107
panic(0x18d1820, 0xc4210eec20)
	/usr/local/go/src/runtime/panic.go:502 +0x229
github.com/gardener/machine-controller-manager/vendor/github.com/aliyun/alibaba-cloud-sdk-go/sdk/utils.GetTimeInFormatISO8601(0x18fcac0, 0xc420bd0a80)
	/Users/i041966/projects/go-root/src/github.com/gardener/machine-controller-manager/vendor/github.com/aliyun/alibaba-cloud-sdk-go/sdk/utils/utils.go:57 +0x13f
github.com/gardener/machine-controller-manager/vendor/github.com/aliyun/alibaba-cloud-sdk-go/sdk/auth.completeRpcSignParams(0x1e31fe0, 0xc42032e420, 0x1e1a5e0, 0xc421156008, 0xc420010b50, 0xb, 0x0, 0x0)
	/Users/i041966/projects/go-root/src/github.com/gardener/machine-controller-manager/vendor/github.com/aliyun/alibaba-cloud-sdk-go/sdk/auth/rpc_signature_composer.go:47 +0x1e1
github.com/gardener/machine-controller-manager/vendor/github.com/aliyun/alibaba-cloud-sdk-go/sdk/auth.signRpcRequest(0x1e31fe0, 0xc42032e420, 0x1e1a5e0, 0xc421156008, 0xc420010b50, 0xb, 0xc4202cd760, 0x7)
	/Users/i041966/projects/go-root/src/github.com/gardener/machine-controller-manager/vendor/github.com/aliyun/alibaba-cloud-sdk-go/sdk/auth/rpc_signature_composer.go:26 +0x73
[...]

PR #146, opened 2018-08-31
Fix panic issue caused by Alicloud SDK -- No zoninfo.zip file found

jia-jerry:
Fix a panic issue
Fixes #
#144 

**Release note**:
NONE


jia-jerry:
Hi @PadmaB @hardikdr,
It is strange I didn't get code reviewer automatically. Would you please check it? Thanks a lot.

hardikdr:
Thank for the PR- sure, I will have soon have a look on it.

Dockerfile
```diff
@@ -2,6 +2,10 @@ FROM alpine:3.6
 
 RUN apk add --update bash curl
 
+# Resources used by Golang libary
+ENV ZONEINFO=/zone-info/zoneinfo.zip
+ADD ./assets/zoneinfo.zip /zone-info/zoneinfo.zip
```
hardikdr: I am wondering if there is any way we could avoid dependency on zoneinfo.zip in Dockerfile or overall?

prashanth26: I second with @hardikdr on this. If there is an alternative way to have this we would prefer that. Adding files into the Docker image may not be appreciated users of MCM.

jia-jerry: It is caused by code /vendor/github.com/aliyun/alibaba-cloud-sdk-go/sdk/utils/utils.go
```golang
func GetTimeInFormatISO8601() (timeStr string) {
	var gmt *time.Location
	var err error
	if LoadLocationFromTZData != nil && TZData != nil {
		gmt, err = LoadLocationFromTZData("GMT", TZData)
	} else {
		gmt, err = time.LoadLocation("GMT")
	}
	if err != nil {
		panic(err)
	}
	return time.Now().In(gmt).Format("2006-01-02T15:04:05Z")
}
```

By default, LoadLocationFromTZData is nil and the program goes to the *else* block. time.LoadLocation function always requires zoneinfo data in some paths. As a result, err is not nil and panic occurs. If go lib is installed, the zoneinfo.zip will always exists and everything goes well. But for bare Linux Docker image, no such a data file exists. 

There are 2 solutions to avoid panic issue:
1.  According to the implementation of go lib time/zoneinfo.go, add zoneinfo.zip file just as what I did in the PR. Pros: the logic will always be correct because is a go libary implementation. Cons: An extra data file will be added
2. Implement function LoadLocation with no extra data file supported and assign to utils.LoadLocationFromTZData. Pros: No extra file. Cons: This is not a simple fuction and no one can expect you make no bugs.

What do you suggest?

jia-jerry: File zoneinfo.zip is from $GOROOT/lib/time/zoneinfo.zip. 
This is README at the same folder:
```
The zoneinfo.zip archive contains time zone files compiled using
the code and data maintained as part of the IANA Time Zone Database.
The IANA asserts that the database is in the public domain.

For more information, see
http://www.iana.org/time-zones
ftp://ftp.iana.org/tz/code/tz-link.htm
http://tools.ietf.org/html/rfc6557

To rebuild the archive, read and run update.bash.
```

This is update.bash file
```bash
#!/bin/bash
# Copyright 2012 The Go Authors. All rights reserved.
# Use of this source code is governed by a BSD-style
# license that can be found in the LICENSE file.

# This script rebuilds the time zone files using files
# downloaded from the ICANN/IANA distribution.
# Consult http://www.iana.org/time-zones for the latest versions.

# Versions to use.
CODE=2017c
DATA=2017c

set -e
rm -rf work
mkdir work
cd work
mkdir zoneinfo
curl -L -O http://www.iana.org/time-zones/repository/releases/tzcode$CODE.tar.gz
curl -L -O http://www.iana.org/time-zones/repository/releases/tzdata$DATA.tar.gz
tar xzf tzcode$CODE.tar.gz
tar xzf tzdata$DATA.tar.gz

# Turn off 64-bit output in time zone files.
# We don't need those until 2037.
perl -p -i -e 's/pass <= 2/pass <= 1/' zic.c

make CFLAGS=-DSTD_INSPIRED AWK=awk TZDIR=zoneinfo posix_only

# America/Los_Angeles should not be bigger than 1100 bytes.
# If it is, we probably failed to disable the 64-bit output, which
# triples the size of the files.
size=$(ls -l zoneinfo/America/Los_Angeles | awk '{print $5}')
if [ $size -gt 1200 ]; then
	echo 'zone file too large; 64-bit edit failed?' >&2
	exit 2
fi

cd zoneinfo
rm -f ../../zoneinfo.zip
zip -0 -r ../../zoneinfo.zip *
cd ../..

echo
if [ "$1" = "-work" ]; then
	echo Left workspace behind in work/.
else
	rm -rf work
fi
echo New time zone files in zoneinfo.zip.
```

hardikdr: Thanks Eric for the info. 
From the looks, does not seem to be an issue, I will give some buffer to Matthias and merge later.

PR #161, opened 2018-09-19
Proposal for GRPC based implementation of Cloud Providers for MCM

ggaurav10:
Proposal for GRPC based implementation of Cloud Providers for MCM

Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/docs/proposals/external_providers_grpc.md
+++ b/docs/proposals/external_providers_grpc.md
@@ -1,0 +1,56 @@
+# GRPC based implementation of Cloud Providers - WIP
+
+## Goal:
+Currently the Cloud Providers' (CP) functionalities ( Create(), Delete(), List() ) are part of the Machine Controller Manager's (MCM)repository. Because of this, adding support for new CPs into MCM requires merging code into MCM which may not be required for core functionalities of MCM itself. Also, for various reasons it may not be feasible for all CPs to merge their code with MCM which is an Open Source project.
+
+Because of these reasons, it was decided that the CP's code will be moved out in separate repositories so that they can be maintained separately by the respective teams. Idea is to make MCM act as a GRPC server, and CPs as GRPC clients. The CP can register themselves with the MCM using a GRPC service exposed by the MCM. Details of this approach is discussed below.
+
+## How it works:
+MCM acts as GRPC server and listens on a pre-defined port 5000. It implements below GRPC services. Details of each of these services are mentioned in next section.
+* `Register()`
+* `GetMachineClass()`
+* `GetSecret()`
+
+## GRPC services exposed by MCM:
+
+### Register()
+`rpc Register(stream DriverSide) returns (stream MCMside) {}`
+
+The CP GRPC client calls this service to register itself with the MCM. The CP passes the `kind` and the `APIVersion` which it implements, and MCM maintains an internal map for all the registered clients. A GRPC stream is returned in response which is kept open througout the life of both the processes. MCM uses this stream to communicate with the client  for machine operations: `Create()`, `Delete()` or `List()`.
+The CP client is responsible for reading the incoming messages continuously, and based on the `operationType` parameter embedded in the message, it is supposed to take the required action. This part is already handled in the package `grpc/infraclient`.
+To add a new CP client, import the package, and implement the `ExternalDriverProvider` interface:
+
+```
+type ExternalDriverProvider interface {
+	Create(machineclass *MachineClassMeta, credentials, machineID, machineName string) (string, string, error)
+	Delete(machineclass *MachineClassMeta, credentials, machineID string) error
+	List(machineclass *MachineClassMeta, credentials, machineID string) (map[string]string, error)
+}
+```
+
+### GetMachineClass()
+`rpc GetMachineClass(MachineClassMeta) returns (MachineClass) {}`
+
+As part of the message from MCM for various machine operations, the name of the machine class is sent instead of the full machine class spec. The CP client is expected to use this GRPC service to get the full spec of the machine class. This optionally enables the client to cache the machine class spec, and make the call only if the machine calass spec is not already cached.
+
+### GetSecret()
+`rpc GetSecret(SecretMeta) returns (Secret) {}`
+
+As part of the message from MCM for various machine operations, the Cloud Config (CC) and CP credentials are not sent. The CP client is expected to use this GRPC service to get the secret which has CC and CP's credentials from MCM. This enables the client to cache the CC and credentials, and to make the call only if the data is not already cached.
+
+## How to add a new Cloud Provider's support
+Import the package `grpc/infraclient` and `grpc/infrapb` from MCM (currently in MCM's "grpc-driver" branch)
+* Implement the interface `ExternalDriverProvider`
+    * `Create()`: Creates a new machine
+    * `Delete()`: Deletes a machine
+    * `List()`: Lists machines
+* Use the interface `MachineClassDataProvider`
+    * `GetMachineClass()`: Makes the call to MCM to get machine class spec
+    * `GetSecret()`: Makes the call to MCM to get secret containing Cloud Config and CP's credentials
+
+### Example implementation:
+This approach is implemented in MCM's "grpc-driver" branch:
+https://github.com/gardener/machine-controller-manager/tree/grpc-driver
+
+Refer GRPC based implementation for AWS client:
+https://github.com/ggaurav10/aws-driver-grpc


docs/proposals/external_providers_grpc.md
```diff
@@ -0,0 +1,48 @@
...
+
+### GetSecret()
+`rpc GetSecret(SecretMeta) returns (Secret) {}`
+
+As part of the message from MCM for various machine operations, the Cloud Config (CC) is not sent. The CP client is expected to use this GRPC service to get the secret which has CC and CP's credentials from MCM. This enables the client to cache the CC and credentials, and to make the call only if the data is not already cached.
+
+## How to add a new Cloud Provider's support
+Import the package `grpc/infraclient` and `grpc/infrapb` from MCM (currently in MCM's "grpc-driver" branch)
+* Implement the interface `ExternalDriverProvider`
+    * `GetMachineClassType()`: Returns the kind and APIVersion of machine class
```
hardikdr: Wondering if there is a way to avoid it ?

hardikdr: Hm. I believe machine-api, in general, will have single apiversion for all the objects under it. We can evolve as the need come, and for now, maybe we assume the exact same version for MachineClass as of machine objects?

hardikdr: As discussed, let's, please hardwire this: https://github.com/ggaurav10/aws-driver-grpc/blob/master/cmd/external-aws-client/external-aws-client.go#L35 to directly here: https://github.com/ggaurav10/aws-driver-grpc/blob/master/pkg/grpc/infraclient/infraclient.go#L124

This will save external-driver from implementing a method.

docs/proposals/external_providers_grpc.md
```diff
@@ -0,0 +1,48 @@
...
+* `Register()`
+* `GetMachineClass()`
+* `GetSecret()`
+
+## GRPC services exposed by MCM:
+
+### Register()
+`rpc Register(stream DriverSide) returns (stream MCMside) {}`
+
+The CP GRPC client calls this service to register itself with the MCM. The CP passes the `kind` and the `APIVersion` which it implements, and MCM maintains an internal map for all the registered clients. A GRPC stream is returned in response which is kept open througout the life of both the processes. MCM uses this stream to communicate with the client  for machine operations: `Create()`, `Delete()` or `List()`.
```
hardikdr: > MCM maintains an internal map for all the registered clients

Is it possible to avoid keeping the state in-memory ?
* if MCM restarts it looses the state. Driver then needs to to figure out the restart incident and re-register. 

How about persisting the driver-name[provisioner in storageclasses] - in the MachineClass ?? Can you please add this as a possible way - we can then check the implementation details. 
For example in StorageClass:  https://kubernetes.io/docs/concepts/storage/storage-classes/#aws

ggaurav10: If MCM goes down, the connection/stream also breaks. Thus, driver anyway need to create a new connection in that case.

prashanth26: I agree with @ggaurav10 and @amshuman-kr on this.

docs/proposals/external_providers_grpc.md
```diff
@@ -0,0 +1,47 @@
...
+The CP client is responsible for reading the incoming messages continuously, and based on the `operationType` parameter embedded in the message, it is supposed to take the required action. This part is already handled in the package `grpc/infraclient`. To add a new CP client, import the package, and implement the `ExternalDriverProvider` interface.
+
+### GetMachineClass()
+`rpc GetMachineClass(MachineClassMeta) returns (MachineClass) {}`
+
+As part of the message from MCM for various machine operations, the name of the machine class is sent instead of the full machine class spec. The CP client is expected to use this GRPC service to get the full spec of the machine class. This enables the client to cache the machine class spec, and make the call only if the machine calass spec is not already cached.
```
prashanth26: s/This enables the client to cache the machine class spec, and make the call only if the machine class spec is not already cached./This **optionally** enables drivers to cache machine objects instead of repeatedly receiving machineClasses across every request.

docs/proposals/external_providers_grpc.md
```diff
@@ -0,0 +1,47 @@
...
+
+### GetSecret()
+`rpc GetSecret(SecretMeta) returns (Secret) {}`
+
+As part of the message from MCM for various machine operations, the Cloud Config (CC) is not sent. The CP client is expected to use this GRPC service to get the secret which has CC and CP's credentials from MCM. This enables the client to cache the CC and credentials, and to make the call only if the data is not already cached.
```
prashanth26: s/the Cloud Config (CC) is/the Cloud Config (CC) and CP credentials are

docs/proposals/external_providers_grpc.md
```diff
@@ -0,0 +1,47 @@
...
+* `Register()`
+* `GetMachineClass()`
+* `GetSecret()`
+
+## GRPC services exposed by MCM:
+
+### Register()
+`rpc Register(stream DriverSide) returns (stream MCMside) {}`
+
+The CP GRPC client calls this service to register itself with the MCM. The CP passes the `kind` and the `APIVersion` which it implements, and MCM maintains an internal map for all the registered clients. A GRPC stream is returned in response which is kept open througout the life of both the processes. MCM uses this stream to communicate with the client  for machine operations: `Create()`, `Delete()` or `List()`.
```
amshuman-kr: In the context of cluster-api/machine-api integration, there may not be machine classes per cloud provider, we need a different mechanism to differentiate and route the machines to the right drivers.

Is there something in the provider config or machine class that can help us choose the right driver? Does the machine-api address the possibility of having one MCM that works with multiple cloud providers? If so how is the correct driver or cloud provider chosen in the API?

docs/proposals/external_providers_grpc.md
```diff
@@ -0,0 +1,47 @@
...
+
+As part of the message from MCM for various machine operations, the name of the machine class is sent instead of the full machine class spec. The CP client is expected to use this GRPC service to get the full spec of the machine class. This optionally enables the client to cache the machine class spec, and make the call only if the machine calass spec is not already cached.
+
+### GetSecret()
+`rpc GetSecret(SecretMeta) returns (Secret) {}`
+
+As part of the message from MCM for various machine operations, the Cloud Config (CC) and CP credentials are not sent. The CP client is expected to use this GRPC service to get the secret which has CC and CP's credentials from MCM. This enables the client to cache the CC and credentials, and to make the call only if the data is not already cached.
+
+## How to add a new Cloud Provider's support
+Import the package `grpc/infraclient` and `grpc/infrapb` from MCM (currently in MCM's "grpc-driver" branch)
```
amshuman-kr: We have to change the package names ASAP. Frankly, `grpc/infraclient` and `grpc/infrapb` are ugly. The "infra" part is not adding anything.

1. How about `grpc/client` and `grpc/server`?
But this does not say anything about drivers.
1. How about `driver/grpc/client` and `driver/grpc/server`?
1. Or simply `grpc/driver` (meaning client) and `grpc/controller` (meaning server)?

docs/proposals/external_providers_grpc.md
```diff
@@ -0,0 +1,47 @@
...
+* Implement the interface `ExternalDriverProvider`
```
amshuman-kr: We should specify the signature of the interface methods.

amshuman-kr: We should chose a better name for `ExternalDriverProvider`. Might `ExternalDriver` be enough?

docs/proposals/external_providers_grpc.md
ggaurav10: no, AFAIK. @hardikdr ?

docs/proposals/external_providers_grpc.md
zanetworker: +1 for 2 we can also just use `grpc/mcm` which is more intuitive because controller is an umbrella term.

amshuman-kr: So, `grpc/driver` and `grpc/mcm`?

ggaurav10: so there are 3 directories in `pkg/grpc/`
1. `infraclient/` has the grpc client related code for bootstrapping the actual cloud provider code (not used by MCM)
1. `infrapb/` has the grpc core files (needed by both MCM and cloud provider)
1. `infraserver/` has the grpc server related code to bootstrap the actual MCM server (not used by external client; that's why no mention of this in the proposal doc)

The `infraclient` is not required by MCM; this only needs to be imported and used by the external cloud provider. MCM acts as place holder currently for this.
Refer https://github.com/ggaurav10/aws-driver-grpc/tree/master/pkg for reference.

Suggestions:
1. `pkg/grpc/infraclient` -> `pkg/driver/grpc/client` (actually not used by MCM. Currently placeholder so that cloud provider code can use it)
1. `pkg/grpc/infrapb` -> `pkg/driver/grpc/grpc_svc`
1. `pkg/grpc/infraserver` -> `pkg/driver/grpc/server`

There is already a `pkg/driver/` which currently holds in-tree driver implementations for aws, azure, etc. which will be removed eventually.

PR #173, opened 2018-10-03
Enhancement/k8s version update

### LINKED ISSUE #91: Update Kubernetes version 1.11
Currently MCM uses Kubernetes version 1.9 and this introduces github.com/juju/ratelimit/ which has an LGPL license. We need to investigate the possibility of moving to Kubernetes version 1.11 to avoid this issue as this has been resolved in the latest Kubernetes version 1.11. However we have to keep in mind of the impact this would have on supporting older version of Kubernetes clusters (1.8 specifically).

prashanth26:
This PR updates K8s version support to 1.10 and hence eliminates dependency on juju/ratelimit which had introduced licensing issues. 

Fixes #91 

- Update of k8s version to 1.10 was possible but to 1.11 caused unresolvable  issues with vendors 
- However upgrading to just 1.10 solves our dependency with juju/ratelimit
- For ease of review, I have split the commits into two - changes in toml and lock file, changes in the vendor file.

**Release note**:

```improvement operator
Updates vendor dependecies to support K8s version 1.10
```


prashanth26:
I have temporarily made a fix for the tests to work with k8s 1.10, refer - https://github.com/gardener/machine-controller-manager/pull/173/commits/768de7779216228e8f45dc0ab6cef8ee7ea25ee7.

If I understand it correctly, the watch mechanism is supported implicitly in the fake reactors in 1.10 and we have a workaround for it in our testing framework. This would require some effort and freezing it for now. For now, all works fine.

Kindly review when possible.

Thanks & Regards,
Prashanth

PR #190, opened 2018-11-21
Packet driver support

deitch:

Add support for [Packet](https://www.packet.com) cloud in-tree. Eventually, when out-of-tree drivers are ready, will be replaced with out-of-tree support.

This remains a work-in-progress (hence the `[WIP]` tag on the subject). At the recommendation of @rfranzke , I followed the directions [here](https://github.com/gardener/machine-controller-manager/blob/master/docs/development/new_cp_support.md) as best I could, but I am sure there are issues.

To be ready for merging, it still needs:

1. Someone to review and comment, make sure I didn't do anything egregiously wrong, which I can then fix.
2. Testing against Packet infrastructure. Once the first step is done and I make changes, I then will test.

**Release note**:

```improvement operator
Add support for Packet cloud provider https://www.packet.com
```


prashanth26:
Hi @deitch ,

We really appreciate the PR. We have been stuck with a couple of other things. We shall get back to you as soon as possible with a review.

Thanks for your patience.
Prashanth

deitch:
Hi @prashanth26 

> I apologize for the delay in review

No worries. You warned me on slack, so I knew it would take a bit of time.

> You missed out to add the vendors, probably because we forgot to mention it in the steps

Yeah, I didn't know about it. I guess this means you check `vendors/` into version control?

> dep ensure --update causes (as mentioned in the Makefile) issues

I didn't see that in the `Makefile`? In any case, is it basically `make revendor` and then check it in?

> It would be great to have some unit tests for the changes your have made

I agree. I usually am quite good about unit tests (or read that as, "humble enough to accept that my stuff won't work without them"...), but wasn't sure how you do tests here. The only tests I saw were for [alicloud](
https://github.com/gardener/machine-controller-manager/blob/master/pkg/driver/driver_alicloud_test.go) [controller](https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/controller_suite_test.go), although the latter isn't really a test of the cloud provider.

Care to share some pointers or examples you have used in the past?

prashanth26:
> Yeah, I didn't know about it. I guess this means you check vendors/ into version control?

Yes. I mean't check vendors into the version control.

> I didn't see that in the Makefile? In any case, is it basically make revendor and then check it in?

Make revendor might cause issues. Hence just running `dep ensure` once you have dep installed should work

> I agree. I usually am quite good about unit tests (or read that as, "humble enough to accept that my stuff won't work without them"...), but wasn't sure how you do tests here. The only tests I saw were for alicloud controller, although the latter isn't really a test of the cloud provider. Care to share some pointers or examples you have used in the past?

We currently don't have too many tests in the repository as we have also just started adding tests. Right now, as you mentioned we don't have any tests for the drivers. But we have started working on adding tests for the controller package. We do this by faking the driver, refer [this](https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/controller_suite_test.go#L220-L310). My first guess would be that some sort of faking might be need with the cloud API calls as well. I am not very sure. I shall check with @amshuman-kr and @hardikdr who are more knowledgeable with this and get back to you.

deitch:
Thanks. I’m mostly off this week, but I’ll try to get some in, and look for your response.

pkg/driver/driver_packet.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		glog.Errorf("Could not create machine: %v", err)
+		return "", "", err
+	}
+	return d.encodeMachineID(device.Facility.ID, device.ID), device.Hostname, nil
+}
+
+// Delete method is used to delete a Packet machine
+func (d *PacketDriver) Delete() error {
+
```
prashanth26: Does it make sense to check if the VM actually exists, before trying to delete it? Refer [this](https://github.com/gardener/machine-controller-manager/blob/master/pkg/driver/driver_aws.go#L138-L145). Even checking based on SDK call return would suffice.

deitch: OK checking.

deitch: My understanding then is that `Delete()` should be idempotent? If I call `Delete()` on a device that no longer exists (or never did), it should return `nil` rather than an error? When I call delete:

* if after call, no such device will exist (whether because it didn't exist before, or because it successfully deleted): return `nil`
* if delete fails, return `error`

Is that right?

deitch: In that case, I am going to do it slightly differently. The packet go client just wraps a REST interface. I am going to leave it as is, but capture `404` returns and handle them.

deitch:
@prashanth26 I did the following:

1. rebased off of latest `master` and added the `404` check and appropriate return (1 commit)
2. re-autogenerated code (1 commit)
3. re-ran `dep ensure` (1 commit)

pkg/driver/driver_packet.go
prashanth26: Thanks 👍  Capturing a 404 should also work.

prashanth26:
The PR looks good to me. Just having one final look.

hardikdr:
Looks good to me as well. 
@deitch We do not have a means of testing this PR, I would request you to test it thoroughly please, and let us know if something needs to be improved. 
Again we really appreciate your contribution, looking forward to having more contributions ahead :)

@prashanth26  please merge it once done with final check.

prashanth26:
Hi @deitch ,

Did you try building the controller after updating the vendors? I think there is a version mismatch and the building of controller is failing. 

Check out this if you can - https://concourse.ci.infra.gardener.cloud/teams/gardener/pipelines/machine-controller-manager-master/jobs/master-pull-request-job/builds/24. I have pasted the error messages below, just in case.

```
Failed to compile controller:

# github.com/gardener/machine-controller-manager/pkg/driver
pkg/driver/driver_packet.go:55:3: cannot use d.PacketMachineClass.Spec.Facility (type []string) as type string in field value
pkg/driver/driver_packet.go:112:38: too many arguments in call to svc.Devices.List
	have (string, *packngo.ListOptions)
	want (string)
pkg/driver/driver_packet.go:141:10: undefined: packngo.NewClientWithAuth
path is /tmp/build/a94a8fe5/pull-request-gardener_machine-controller-manager-pr_master/tmp/src/github.com/gardener/machine-controller-manager/test/output
```

deitch:
I must have messed up somewhere when rebasing and reapplying. We will figure it out.

deitch:
> Check out this if you can - https://concourse.ci.infra.gardener.cloud/teams/gardener/pipelines/machine-controller-manager-master/jobs/master-pull-request-job/builds/24.

Hmm, I get a `401 Unauthorized` when trying to view it.

deitch:
No, it looks like it is picking up an older version of the `packngo` SDK.

deitch:
OK, now it does the right thing. I had to constrain to the correct version in `Gopkg.toml`, and now it is fine. Thanks @prashanth26

deitch:
@prashanth26 can you relaunch those concourse integration tests?

deitch:
@hardikdr wrote:

> We do not have a means of testing this PR, I would request you to test it thoroughly please

I am trying to figure out a few things:

1. A decent way to test _just_ MCM without deploying entire clusters (garden and shoot). Since this has a clearly defined API, it should be possible to test just MCM against packet, should it not? Can you share some pointers?
2. How do you tests against AWS and GCE? It looks like you have some interesting testing [here](https://github.com/gardener/machine-controller-manager/blob/master/.ci/integration-tests) which includes various cloud providers [here](https://github.com/gardener/machine-controller-manager/blob/master/.ci/integration-tests#L21) and runs tests [here](https://github.com/gardener/machine-controller-manager/blob/master/.ci/integration-tests#L220) and [here](https://github.com/gardener/machine-controller-manager/blob/master/.ci/integration-tests#L237). How do we leverage that to do packet correctly as well?

deitch:
Hmm, I found [this](https://github.com/gardener/machine-controller-manager/blob/master/docs/development/local_setup.md#testing-machine-classes), but it doesn't quite work:

```
./managevm --secret ./dev/packet-secret.yaml --machineclass ./kubernetes/machine_classes/packet-machine-class.yaml --classkind PacketMachineClass --machinename test
2018/12/18 11:11:24 Unknown class kind PacketMachineClass
```

Same if I try class kind `Packet`. Help?

deitch:
OK, I figured out that the CLI is a separate thing, not in the docs. I added it there too.

prashanth26:
``` 
@prashanth26 can you relaunch those concourse integration tests?
```
I have triggered them. I can see that the build succeeds now. However, the CI tests might still fail as something is wrong with the Azure integration setup. We are trying to fix it here - https://github.com/gardener/machine-controller-manager/pull/203. 

```
Hmm, I found this, but it doesn't quite work:

./managevm --secret ./dev/packet-secret.yaml --machineclass ./kubernetes/machine_classes/packet-machine-class.yaml --classkind PacketMachineClass --machinename test
2018/12/18 11:11:24 Unknown class kind PacketMachineClass
Same if I try class kind Packet. Help?
``` 
I guess you figured this out. Thanks for pointing it out. 

``` 
I am trying to figure out a few things: A decent way to test just MCM without deploying entire clusters (garden and shoot). Since this has a clearly defined API, it should be possible to test just MCM against packet, should it not? Can you share some pointers?
How do you tests against AWS and GCE? It looks like you have some interesting testing here which includes various cloud providers here and runs tests here and here. How do we leverage that to do packet correctly as well?
```
I shall get back to you on this.

```
OK, I figured out that the CLI is a separate thing, not in the docs. I added it there too.
```
Thanks for pointing it out. Which commit has your changes to the docs? Maybe i missed it.

deitch:
> I shall get back to you on this.

Thanks

> Which commit has your changes to the docs?

https://github.com/gardener/machine-controller-manager/pull/204

deitch:
Done. I can confirm that running the tests manually from [here](https://github.com/gardener/machine-controller-manager/blob/master/docs/development/local_setup.md#testing-machine-classes) successfully create and destroy machines.

Anything next?

prashanth26:
Okay, thanks for that. It looks good to me as well. 
I need to run your tests, once in the pipeline to confirm everything works fine, which I would do once the issues with the pipeline is fixed later today. I am also waiting for review @hardikdr who is traveling and should be back by Friday. So I might need a couple of days before I can merge it. 

Thanks & Regards,
Prashanth

deitch:
Sure, sounds good. Will look for the notification.

prashanth26:
Hi @deitch ,

Everything looks fine. But can you rebase with current master so that we can re-run your tests with our latest changes to the CI pipeline?

Thanks in advance,
Prashanth

deitch:
Yup.

deitch:
Rebased.

deitch:
And CI passes! :-)

prashanth26:
Hi @deitch ,

The CI tests only run after adding the label. So it's running now. But I am positive that it would work now. Once it goes through, we shall merge it.

deitch:
Sounds good to me!

> interesting testing here which includes various cloud providers here and runs tests here and here
> I shall get back to you on this.

When when you have that, let me know. I will be happy to make this part of the automated testing.

prashanth26:
> When when you have that, let me know. I will be happy to make this part of the automated testing.

- The way we perform integration-tests is by deploying a cluster (shoot) using Gardener on which we try to do various operations like create/delete/scale-up/rolling-update and see if the behavior of machines joining is as expected.
- As right now, Gardener doesn't support Packet it might be difficult for us to use the CI pipeline to perform tests for Packet.
- And our plan very soon (@hardikdr is working on the same) is to support drivers out-of-tree and we might have separate repositories for each provider. Hence the way do integration tests might differ in the very near future.
- Hence I think it might be better to wait for these changes to happen before we add integration tests.
- And regarding unit tests, once we have some unit tests for the other providers I shall give you a heads up and we could add tests for packet together.

prashanth26:
We had some network issues with our CI pipeline. Had to re-run the tests. Looks good now. All checks are passing. Merging it. 

Thanks for the PR @deitch. We really appreciate it 👍

deitch:
> Gardener doesn't support Packet it might be difficult for us to use the CI pipeline to perform tests for Packet

Although we are getting there (which was the point of this PR). 

However, since, as you said, the goal is to pull all of these out-of-tree, I am ok with waiting for that.

> Thanks for the PR @deitch. We really appreciate it

With pleasure! I will hop onto the Slack channel to discuss what other PRs we need to support it.

PR #193, opened 2018-11-22
Using default hostname in Alicloud.

### LINKED ISSUE #187: Host name is in invalid format in Alibaba Cloud
For Alibaba Cloud API, there is a restriction for host name. It is a string of at least two characters. “hostname” cannot start or end with “.” or “-“. In addition, **two or more consecutive** “.” or “-“ symbols are not allowed. 
https://www.terraform.io/docs/providers/alicloud/r/instance.html#host_name
https://help.aliyun.com/document_detail/63440.html?spm=a2c4g.11186623.6.879.7cb751c9a051Gu (It is in Chinese)

However, the hostname constructed by MCM has consecutive '-', which is not allowed to Alicloud API.

jia-jerry:
Fix the bug when node name is not recognized by Machine

Fixes #187

 
** To Reviewers **:
PR #188 is not elegant. This PR is better.

### DIFF
--- a/pkg/driver/driver_alicloud.go
+++ b/pkg/driver/driver_alicloud.go
@@ -42,33 +42,6 @@ type AlicloudDriver struct {
 	MachineName          string
 }
 
-func toInstanceTags(tags map[string]string) ([]ecs.RunInstancesTag, error) {
-	result := []ecs.RunInstancesTag{{}, {}}
-	hasCluster := false
-	hasRole := false
-
-	for k, v := range tags {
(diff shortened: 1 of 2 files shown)


jia-jerry:
@ggaurav10 , would you please have a review ?

hardikdr:
@jia-jerry thanks for the PR, and sorry for the delay. At the moment, we do not have a means of testing PR in alicloud env, for now you will have to double check it on your env until we get it in our integration-test framework :)
I had a quick look, overall looks good to me.
@prashanth26 @ggaurav10  Can you please have a look and merge if looks good?

PR #202, opened 2018-12-17
Optimise pod handling in machine updates

### LINKED ISSUE #120: Optimise Pod Handling in Machine Roll-Out
When we do a rolling update, we don't give Kubernetes any hints to schedule the drained pods on the new machines rather than the olds ones, right? This in turn means, a rolling update may hit a pod more than once. Is there a chance to "guide" the scheduler in scheduling the displaced pods rather on the new than the old machines, maybe with some temporary node labels but ideally without modifying/patching the deployment/pod specs, which would be pretty messy? Maybe, the easiest solution could be to actually `cordon` the old machines the moment the first batch of new machines joined the cluster (I am wary to do it earlier because that could mean that pods can't be scheduled anymore until new machines have joined the cluster). What do you think?

prashanth26:

Fixes #120 

**Release note**:

```noteworthy operator
Prefers scheduling of pods on newer machines during roll-outs
```

### DIFF
--- a/pkg/controller/controller_utils.go
+++ b/pkg/controller/controller_utils.go
@@ -38,7 +38,7 @@ import (
 	machineapi "github.com/gardener/machine-controller-manager/pkg/client/clientset/versioned/typed/machine/v1alpha1"
 	hashutil "github.com/gardener/machine-controller-manager/pkg/util/hash"
 	taintutils "github.com/gardener/machine-controller-manager/pkg/util/taints"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -820,7 +820,7 @@ func AddOrUpdateTaintOnNode(c clientset.Interface, nodeName string, taints ...*v
 		if !updated {
 			return nil
 		}
-		return PatchNodeTaints(c, nodeName, oldNode, newNode)
+		return UpdateNodeTaints(c, nodeName, oldNode, newNode)
 	})
 }
 
@@ -877,7 +877,7 @@ func RemoveTaintOffNode(c clientset.Interface, nodeName string, node *v1.Node, t
 		if !updated {
 			return nil
 		}
-		return PatchNodeTaints(c, nodeName, oldNode, newNode)
+		return UpdateNodeTaints(c, nodeName, oldNode, newNode)
 	})
 }
 
@@ -902,6 +902,15 @@ func PatchNodeTaints(c clientset.Interface, nodeName string, oldNode *v1.Node, n
 	}
 
 	_, err = c.Core().Nodes().Patch(string(nodeName), types.StrategicMergePatchType, patchBytes)
+	return err
+}
+
+// UpdateNodeTaints updates node's taints.
+func UpdateNodeTaints(c clientset.Interface, nodeName string, oldNode *v1.Node, newNode *v1.Node) error {
+	newNodeClone := oldNode.DeepCopy()
+	newNodeClone.Spec.Taints = newNode.Spec.Taints
+
+	_, err := c.Core().Nodes().Update(newNodeClone)
 	return err
 }
 
--- a/pkg/controller/controller_utils.go
+++ b/pkg/controller/controller_utils.go
@@ -881,7 +881,9 @@ func RemoveTaintOffNode(c clientset.Interface, nodeName string, node *v1.Node, t
 	})
 }
 
-// PatchNodeTaints patches node's taints.
+// PatchNodeTaints is for updating the node taints from oldNode to the newNode
+// It makes a TwoWayMergePatch by comparing the two objects
+// It calls the Patch() method to do the final patch
 func PatchNodeTaints(c clientset.Interface, nodeName string, oldNode *v1.Node, newNode *v1.Node) error {
 	oldData, err := json.Marshal(oldNode)
 	if err != nil {
@@ -905,13 +907,18 @@ func PatchNodeTaints(c clientset.Interface, nodeName string, oldNode *v1.Node, n
 	return err
 }
 
-// UpdateNodeTaints updates node's taints.
+// UpdateNodeTaints is for updating the node taints from oldNode to the newNode
+// using the nodes Update() method
 func UpdateNodeTaints(c clientset.Interface, nodeName string, oldNode *v1.Node, newNode *v1.Node) error {
 	newNodeClone := oldNode.DeepCopy()
 	newNodeClone.Spec.Taints = newNode.Spec.Taints
 
 	_, err := c.Core().Nodes().Update(newNodeClone)
-	return err
+	if err != nil {
+		return fmt.Errorf("failed to create update taints for node %q: %v", nodeName, err)
+	}
+
+	return nil
 }
 
 // WaitForCacheSync is a wrapper around cache.WaitForCacheSync that generates log messages
--- a/pkg/controller/deployment_rollback.go
+++ b/pkg/controller/deployment_rollback.go
@@ -165,7 +165,7 @@ func (dc *controller) removeTaintNodesBackingMachineSets(machineSet *v1alpha1.Ma
 		Effect: "PreferNoSchedule",
 	}
 
-	// Iterate through all machines and place the PreferNoSchedule taint
+	// Iterate through all machines and remove the PreferNoSchedule taint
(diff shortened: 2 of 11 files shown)


pkg/controller/deployment_rolling.go
```diff
@@ -236,3 +251,90 @@ func (dc *controller) scaleDownOldMachineSetsForRollingUpdate(allISs []*v1alpha1
...
+		for _, machine := range filteredMachines {
+			if machine.Labels["node"] != "" {
+				err = AddOrUpdateTaintOnNode(
+					dc.targetCoreClient,
+					machine.Labels["node"],
+					&taints,
+				)
+				if err != nil {
+					glog.Errorf("Node tainting failed for node: %s, %s", machine.Labels["node"], err)
+					return err
```
ggaurav10: In case the error occurs after labelling some (not all) machines in a machine set, we return error. Does it fail the whole rolling update operation without retry? If yes, then we might also need to remove the successful taints/annotations before returning. WDYT?
Same goes for all other places where we return because of failure - not just in this function, but in subsequent function calls too.

prashanth26: This is something important that you have brought up here and I think it makes sense. 

- Does it fail the whole rolling update operation without retry? 
Yes it does fail the while rolling update for then. It retries it later in the next machineDeployment reconciliation which is forced every 10 mins.

-  If yes, then we might also need to remove the successful taints/annotations before returning
I was thinking of an alternate approach where even if the tainting operation fails, we continue the machine rolling update (as it is meant to carry forward even if this step fails). The tainting would then be re-tried in the next machineDeployment reconciliation (which should be in the next few seconds as the machine deployment would be updated soon)?

prashanth26:
- Shouldn't we handle the rollback scenario? Basically, remove the taints from the machines, in case the update is paused and rolled back.

I thought that it would be okay to ignore this as even if it rolls back, these nodes would only have prefer-no-schedule, which still means pods will be scheduled here if its the last option. Also, I haven't seen usages for rollbacks much. I do agree that ideally, this should be there, but I am not sure if it's required here. I have no hard reservations though.

pkg/controller/deployment_rolling.go
ggaurav10: yes. that makes sense too.

prashanth26: I shall do that then.

pkg/controller/deployment_rollback.go
```diff
@@ -120,3 +129,96 @@ func (dc *controller) updateMachineDeploymentAndClearRollbackTo(d *v1alpha1.Mach
 	_, err := dc.controlMachineClient.MachineDeployments(d.Namespace).Update(d)
 	return err
 }
+
+// removeTaintNodesBackingMachineSets taints all nodes backing the machineSets
```
ggaurav10: s/taints/remove taints/

pkg/controller/deployment_rollback.go
```diff
@@ -120,3 +129,96 @@ func (dc *controller) updateMachineDeploymentAndClearRollbackTo(d *v1alpha1.Mach
...
+func (dc *controller) removeTaintNodesBackingMachineSets(machineSet *v1alpha1.MachineSet) error {
+
+	if machineSet.Annotations["PreferNoSchedule"] == "" {
+		// No taint exists
+		glog.Warningf("No taint exits on machineSet: %s. Hence not removing.", machineSet.Name)
+		return nil
+	}
+
+	glog.V(3).Infof("Trying to untaint MachineSet object %q with prefer-no-schedule to avoid scheduling of pods", machineSet.Name)
```
ggaurav10: s/avoid/enable/

pkg/controller/deployment_rollback.go
prashanth26: Done. Refer - https://github.com/gardener/machine-controller-manager/pull/202/commits/b55bea8f53e3b13bc5f06e11dbc3dbe0326d6b17.

pkg/controller/deployment_rollback.go
prashanth26: Done. Refer - https://github.com/gardener/machine-controller-manager/pull/202/commits/b55bea8f53e3b13bc5f06e11dbc3dbe0326d6b17.

pkg/controller/deployment_rollback.go
```diff
@@ -120,3 +129,98 @@ func (dc *controller) updateMachineDeploymentAndClearRollbackTo(d *v1alpha1.Mach
...
+		return err
+	}
+
+	taints := v1.Taint{
+		Key:    PreferNoSchedule,
+		Value:  "True",
+		Effect: "PreferNoSchedule",
+	}
+
+	// Iterate through all machines and place the PreferNoSchedule taint
```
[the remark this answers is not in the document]
prashanth26: Done.

hardikdr:
@prashanth26 
>I thought that it would be okay to ignore this as even if it rolls back, these nodes would only have prefer-no-schedule, which still means pods will be scheduled here if its the last option.

So what is the conclusion here then, are we really not handling the rollback?
Ideally, if we support a rollback mechanism, we should make sure it's not broken/convoluted due to any of the new feature. ;)
Btw PreferNoSchedule can affect rollback's even in a deeper sense, what if after rollback is complete few machines have PNS taint and rest does not have. It basically means few get preferred for no good reason while pod-scheduling. :/

pkg/controller/deployment_rollback.go
```diff
@@ -120,3 +129,97 @@ func (dc *controller) updateMachineDeploymentAndClearRollbackTo(d *v1alpha1.Mach
 	_, err := dc.controlMachineClient.MachineDeployments(d.Namespace).Update(d)
 	return err
 }
+
+// removeTaintNodesBackingMachineSets removes taints from all nodes backing the machineSets
+func (dc *controller) removeTaintNodesBackingMachineSets(machineSet *v1alpha1.MachineSet) error {
```
hardikdr: Can we please make this method more generic? 
- ATM PNS taint seems to be hard-coded, we could get PNS as parameter and use same method for diff taints as well.

Also if this function plans to entertain only one MachineSet than can we please name  it as "MachineSet" at the end-singular. 
wdyt, does it makes sense?

pkg/controller/controller_utils.go
```diff
@@ -905,6 +905,15 @@ func PatchNodeTaints(c clientset.Interface, nodeName string, oldNode *v1.Node, n
 	return err
 }
 
+// UpdateNodeTaints updates node's taints.
+func UpdateNodeTaints(c clientset.Interface, nodeName string, oldNode *v1.Node, newNode *v1.Node) error {
```
hardikdr: Can we please add more comments around what exactly are these [set of] methods for?
- nodeName in signature but not used inside- may be for uniformity?

pkg/controller/deployment_rollback.go
```diff
@@ -120,3 +129,97 @@ func (dc *controller) updateMachineDeploymentAndClearRollbackTo(d *v1alpha1.Mach
...
+
+	if machineSet.Annotations[PreferNoSchedule] == "" {
```
hardikdr: What happens if, due to any reason, PNS is not there as a key in the map, just curious? Will that causes Runtime error?

pkg/controller/deployment_rollback.go
```diff
@@ -120,3 +129,97 @@ func (dc *controller) updateMachineDeploymentAndClearRollbackTo(d *v1alpha1.Mach
...
+		// No taint exists
+		glog.Warningf("No taint exits on machineSet: %s. Hence not removing.", machineSet.Name)
```
hardikdr: typo: s/exits/exists

prashanth26:
> @prashanth26
> 
> > I thought that it would be okay to ignore this as even if it rolls back, these nodes would only have prefer-no-schedule, which still means pods will be scheduled here if its the last option.
> 
> So what is the conclusion here then, are we really not handling the rollback?
> Ideally, if we support a rollback mechanism, we should make sure it's not broken/convoluted due to any of the new feature. ;)
> Btw PreferNoSchedule can affect rollback's even in a deeper sense, what if after rollback is complete few machines have PNS taint and rest does not have. It basically means few get preferred for no good reason while pod-scheduling. :/
> 
> PS: Seems like rollback scene had been taken care, just that didn't reflect here in conversation.

Yes, rollback has been taken care off.

pkg/controller/deployment_rollback.go
prashanth26: Fixed this issue. Refer this commit for all changes - https://github.com/gardener/machine-controller-manager/pull/202/commits/81cc655ce61e22a44780544079bf690e1ce29bb1

PR #209, opened 2019-01-08
[WIP]Support for out-of-tree/external cloud-provider specific drivers

### LINKED ISSUE #178: MCM Out-Of-Tree Extensibility
### Problem

There are at least four possible approaches for out-of-tree drivers as documented in this [proposal](https://github.com/gardener/machine-controller-manager/pull/169).

1. MCM as GRPC server and the driver as GRPC client using bidirectional streaming over long-running GRPC connections (established by the driver during registration) for communication.
1. MCM as GRPC client and the driver as GRPC server using non-streaming regular GRPC calls.
1. Drivers as webhooks with MCM making calls to the driver webhook using some REST API.
1. Drivers a controllers.

We have implemented option 1 already as a PoC. Now we need to discuss and decide on the right approach.

#### MCM as GRPC server and driver as GRPC client
* Pros
  * Already [implemented](https://github.com/gardener/machine-controller-manager/blob/grpc-driver/cmd/external-aws-client/external-aws-client.go).
  * Can work even with `SecretRef` and `CloudConfig` not standardised in the `MachineClass` spec.
  * Enables caching of `MachineClass`, `Secret` and `CloudConfig` if necessary.
  * More secure? Only MCM needs to expose a GRPC endpoint.
  * Drivers can work as sidecar containers or as sidecar `deployments`.
  * Cluster administrator can control drivers' access to `kube-apiserver` and if necessary block it completely.
* Cons
  * Complex and potentially confusing design involving streaming API.
  * Implementing the driver could be complex as it involves handling the streaming messages.
* Mitigation
  * The complexity can be mitigated by writing a GRPC-based [call-back framework](https://github.com/gardener/machine-controller-manager/blob/grpc-driver/pkg/driver/grpc/client/provider.go) which can be used by driver implementation so that they can concentrate on the [cloud-provider side](https://github.com/gardener/machine-controller-manager/blob/grpc-driver/pkg/external/driver/aws/aws.go) without having to worry about the GRPC and MCM.

#### MCM as GRPC client and the driver as GRPC server
* Pros
  * Simpler design involving non-streaming regular GRPC calls.
  * This approach might be similar to the approach taken in CSI etc.
  * Drivers can work as sidecar containers or as sidecar `deployments`.
  * Cluster administrator can control drivers' access to `kube-apiserver` and if necessary block it completely.
* Cons
  * There are subtle differences in the requirements for out-of-tree drivers and those for CSI. For example, multi-tenancy.
  * Requires that `SecretRef` and `CloudConfig` are standardised in the `MachineClass` spec.
  * Caching of `MachineClass`, `Secret` and `CloudConfig` would be hard.
  * Less secure? Drivers need to expose a GRPC endpoint.
  * Implementing a driver would still be complex as it involves implementing a secure GRPC server.
* Mitigation  
  * Complexity can be mitigated by providing a GRPC-based call-back framework similar to [option 1](https://github.com/gardener/machine-controller-manager/blob/grpc-driver/pkg/driver/grpc/client/provider.go) which can be used by driver implementation so that they can concentrate on the [cloud-provider side](https://github.com/gardener/machine-controller-manager/blob/grpc-driver/pkg/external/driver/aws/aws.go) to enable the driver implementation to concentrate on the cloud-provider side without having to worry about the GRPC and MCM.
  * Difficulty in enabling caching and flexibility without having the `SecretRef` and `CloudConfig` standardised in the `MachineClass` can be mitigated by making the MCM *also* a GRPC server (*in addition* to the drivers as GRPC server) to expose API to fetch `MachineClass`, `Secret` and `CloudConfig`.

#### Drivers as webhooks
* Pros
  * Simpler design and REST API has a lower barrier for entry than GRPC.
  * Cluster administrator can control drivers' access to `kube-apiserver` and if necessary block it completely. But this would have some implications on the mitigations below.
* Cons
  * Requires that `SecretRef` and `CloudConfig` are standardised in the `MachineClass` spec.
  * Caching of `MachineClass`, `Secret` and `CloudConfig` would be practically impossible.
* Mitigation
  * The only possible mitigation for not having the `SecretRef`, `CloudConfig` standardised in the `MachineClass` would be by making the driver fetch these from the `kube-apiserver` directly. But this would not only break the webhook pattern but also defeat the very purpose of having a driver framework and the fourth option might be better suited.

#### Drivers as controllers
* Pros
  * Already [adopted](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/cloud/aws/controllers/machine/controller.go) in the cluster-api community.
  * Common design-pattern in Kubernetes and does not introduce new patterns/frameworks.
  * Cluster administrator can control drivers' access to `kube-apiserver` but not block it completely.
  * Support from cloud providers like the [AWS Service Operator](https://github.com/awslabs/aws-service-operator) make this approach more attractive.
* Cons
  * Writing a controller is complex.
  * Some access to `kube-apiserver` would be required from the drivers (read access to `Machines`, `MachineClasses`, `Secrets` and `CloudConfigs` and write access to the `Status` sub-resource of `Machines`).
  * MachineController (at least the cloud-provider-specific parts) would have to be separated out of MCM image.
* Mitigation
  * Complexity of writing a controller can be mitigated by writing a control-loop-based call-back framework similar to [option 1](https://github.com/gardener/machine-controller-manager/blob/grpc-driver/pkg/driver/grpc/client/provider.go) which can be used by driver implementation so that they can concentrate on the [cloud-provider side](https://github.com/gardener/machine-controller-manager/blob/grpc-driver/pkg/external/driver/aws/aws.go) instead of the complexity of writing a controller.
  * The need to have some minimal access to `kube-apiserver` from the drivers cannot be mitigated.
  * The MachineController (at least the cloud-provider-specific parts) can be separated into a separate image and the rest of MCM into a separate image.

hardikdr:
* Introduces new CRD `MachineClass` 
    * Removes the cloud provider specific CRDs such `AWSMachineClass` once all external providers are implemented- hence WIP.
* Implements the gRPC-client based driver to communicate with external-drivers
* Sample external-driver for AWS- https://github.com/gardener/machine-controller-manager-provider-aws
    * This project will be migrated under Gardener org soon.

Fixes #178 

**Release note**:

```improvement operator
Added support for out-of-tree gRPC-based cloud-provider specific drivers
```


prashanth26:
@rfranzke will have it in mind :)

prashanth26:
This PR shall be closed in favor of a Kubernetes controller based approach to manage OOT drivers. New approach implementation can be found here - https://github.com/gardener/machine-controller-manager/pull/460.

PR #248, opened 2019-03-25
Bugfix/azure wait for disk detach

### LINKED ISSUE #242: Azure machines being stuck on deletion
## Issue
It is often observed that Azure machines are stuck on deletion

## Possible cause
As suggested by colleagues at Microsoft, it is observed that detaching of PVs could take longer than 5mins leading to this in a consistent state on hung VMs

## Solution
Need to enhance drain timeouts?

prashanth26:
This PR attempts to fix the Azure issue of VMs stuck in deletion state. Now Azure waits for disks being detached and drains to complete before deleting the VM.

Fixes #242 

**Release note**:

```noteworthy user
The drain is always invoked even the case of forceful deletion
```
```noteworthy user
Drain now tries to evict pods and if eviction fails, it forcefully deletes the pods
```
```improvement operator
Azure explicitly detaches data disks before VM deletion
```

### DIFF
--- a/pkg/controller/machine.go
+++ b/pkg/controller/machine.go
@@ -456,41 +456,45 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
 			// Timeout value obtained by subtracting last operation with expected time out period
 			timeOut := metav1.Now().Add(-timeOutDuration).Sub(machine.Status.CurrentStatus.LastUpdateTime.Time)
 
-			// To perform drain 2 conditions must be satified
-			// 1. force-deletion: "True" label must not be present
-			// 2. Deletion operation must be less than 5 minutes old
-			if machine.Labels["force-deletion"] != "True" && timeOut < 0 {
-				buf := bytes.NewBuffer([]byte{})
-				errBuf := bytes.NewBuffer([]byte{})
-
-				nodeName := machine.Status.Node
-				drainOptions := NewDrainOptions(
-					c.targetCoreClient,
-					timeOutDuration, // TODO: Will need to configure timeout
-					nodeName,
-					int(timeOutDuration.Seconds()),
-					true,
-					true,
-					true,
-					buf,
-					errBuf,
-				)
-				err = drainOptions.RunDrain()
-				if err != nil {
-					lastOperation := v1alpha1.LastOperation{
-						Description:    "Drain failed - " + err.Error(),
-						State:          v1alpha1.MachineStateFailed,
-						Type:           v1alpha1.MachineOperationDelete,
-						LastUpdateTime: metav1.Now(),
-					}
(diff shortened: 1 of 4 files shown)


prashanth26:
@hardikdr /needs-review

pkg/controller/machine.go
```diff
@@ -470,7 +468,7 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
 					c.targetCoreClient,
 					timeOutDuration, // TODO: Will need to configure timeout
 					nodeName,
-					-1,
+					int(timeOutDuration.Seconds()),
```
hardikdr: What is it for ?

pkg/driver/driver_azure.go
```diff
@@ -380,6 +379,33 @@ func (d *AzureDriver) getvms(machineID string, listOfVMs VMs) error {
...
+		glog.Errorf("Failed to list VMs. Error Message - %s", err)
+		return err
+	}
+	metrics.APIRequestCount.With(prometheus.Labels{"provider": "azure", "service": "virtual_machine"}).Inc()
+
+	if len(*vm.StorageProfile.DataDisks) > 0 {
+		// There are disks attached hence need to detach them
+		vm.StorageProfile.DataDisks = &[]compute.DataDisk{}
+
+		_, errChan := vmClient.CreateOrUpdate(d.AzureMachineClass.Spec.ResourceGroup, machineID, vm, cancel)
```
hardikdr: Can you please explain how does it work? 
Where are we making explicit detach calls to VM? , or does the empty-disk array instructs API to delete all disk?

pkg/controller/machine.go
prashanth26: I thought to cap the graceful termination of any pod by the drain timeout should help in overall drain option going through in the drain timeout. But i dunno if it works as expected.

I am hoping that if a pod is set to a graceful termination of `2Hours`, and our drain timeout is '5mins` we max would only give 5mins?

pkg/driver/driver_azure.go
prashanth26: Yes hardik. VM with an update call with empty data disks is how detachment of data disks is done in Azure. They lack a proper SDK call for it. Refer here - https://github.com/Azure/azure-sdk-for-go/issues/1638#issuecomment-384733952

pkg/controller/machine.go
prashanth26: I have reverted this change here - https://github.com/gardener/machine-controller-manager/pull/248/commits/7510f9ce74835b969614d94cb0cfa65f392becea#diff-d8287fe74b5273163c9f6b6c635ad912R475.

hardikdr: Thanks.

PR #249, opened 2019-04-01
Improvement in drain logic

hardikdr:
**What this PR does / why we need it**: This PR slightly improves the efficiency of the drain by eliminating unnecessary calls to API-server.  There seem to be more probable improvements in the drain[distrupting core-logic] which can be entertained in subsequent PR. These changes could also be integrated while switching to drain-lib supported by the community.

Fixes #

**Release note**:

```improvement operator
Drain logic makes use of lesser API calls
```

### DIFF
--- a/pkg/controller/drain.go
+++ b/pkg/controller/drain.go
@@ -71,7 +71,7 @@ const (
 	EvictionSubresource = "pods/eviction"
 
 	// Interval is the default Poll interval
-	Interval = time.Second * 1
+	Interval = time.Second * 5
 
 	daemonsetFatal      = "DaemonSet-managed pods (use --ignore-daemonsets to ignore)"
 	daemonsetWarning    = "Ignoring DaemonSet-managed pods"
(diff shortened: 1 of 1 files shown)


pkg/controller/drain.go
```diff
@@ -154,22 +154,8 @@ func (o *DrainOptions) getController(namespace string, controllerRef *metav1.Own
...
-	if controllerRef == nil {
-		return nil, nil
-	}
-
-	// We assume the only reason for an error is because the controller is
-	// gone/missing, not for any other cause.
-	// TODO(mml): something more sophisticated than this
-	// TODO(juntee): determine if it's safe to remove getController(),
-	// so that drain can work for controller types that we don't know about
-	_, err := o.getController(pod.Namespace, controllerRef)
```
ggaurav10: Do we execute this statement in usual drain scenario?
If not, how does it give any improvement in performance? Does removing this break any other path? Why was this statement there.

hardikdr: Yes, this method is called in usual scenarios, making calls to API server. I don't think it breaks any code path if removed.

PR #265, opened 2019-05-12
Bugfix: Existing machine-object now adopts node-label

hardikdr:
We realized existing machine-objects were not picking up the node-label. This PR fixes that bug.

Fixes #

**Release note**:

```improvement operator
Bugfix: Existing machine-objects now adopts the node-label.
```

### DIFF
--- a/pkg/controller/machine.go
+++ b/pkg/controller/machine.go
@@ -308,11 +308,26 @@ func (c *controller) updateMachineState(machine *v1alpha1.Machine) (*v1alpha1.Ma
 		return machine, err
 	}
 
-	clone, err := c.updateMachineConditions(machine, node.Status.Conditions)
+	machine, err = c.updateMachineConditions(machine, node.Status.Conditions)
 	if err != nil {
 		return machine, err
 	}
-	return clone, nil
+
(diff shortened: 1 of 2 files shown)


pkg/controller/machine.go
```diff
@@ -312,6 +312,20 @@ func (c *controller) updateMachineState(machine *v1alpha1.Machine) (*v1alpha1.Ma
 	if err != nil {
 		return machine, err
 	}
+
+	if clone.Labels == nil {
```
prashanth26: I think we need to take a deepCopy before making the changes here?

hardikdr: thanks, I wasn't really sure to deep-copy here. We can certainly deep-copy, do you have any possible reasoning to support it?

prashanth26: From what I understand, we would be working from an object from the local cache directly. if we don't deep copy. Deep copying creates a local copy on this method and then acts on it.

hardikdr: As discussed over a call, made few changes.
Thanks, this was a good catch..

prashanth26: Looks good now.

PR #266, opened 2019-05-15
[WIP] Ensure packet tags are string array and not key-value map

deitch:

The packet tags were listed as a key/value map instead of array of strings. This fixes it.

Raised by @rfranzke in https://github.com/gardener/gardener/pull/916

**Release note**:

```improvement operator
Fixes error with packet tags
```


deitch:
Cancel this.

PR #270, opened 2019-05-24
Fixing handling of global timout for node drain

### LINKED ISSUE #269: How is globalTimeout handled when draining a node
When draining a node, `evictPods()` is called. It is supposed to wait for duration `globalTimeout` for evicting all the pods.
https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/drain.go#L367
If each of the pods take different times to be evicted, then the timeout timer starts again from 0 because of `for`-ever loop.

ggaurav10:
Global timeout for drain restarts after each successful/unsuccessful pod eviction. This PR fixes this issue.

Fixes #269 

**Release note**:

```improvement operator

```

### DIFF
--- a/pkg/controller/drain.go
+++ b/pkg/controller/drain.go
@@ -37,6 +37,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 )
@@ -313,13 +314,12 @@ func (o *DrainOptions) deleteOrEvictPods(pods []api.Pod) error {
 }
 
 func (o *DrainOptions) evictPods(pods []api.Pod, policyGroupVersion string, getPodFn func(namespace, name string) (*api.Pod, error)) error {
-	doneCh := make(chan bool, len(pods))
-	errCh := make(chan error, 1)
+	returnCh := make(chan error, 1)
 
 	glog.V(3).Info("The following pods were evicted -")
 
 	for _, pod := range pods {
-		go func(pod api.Pod, doneCh chan bool, errCh chan error) {
+		go func(pod api.Pod, returnCh chan error) {
 			var err error
 			glog.V(3).Info("\t", pod.Name)
 			for {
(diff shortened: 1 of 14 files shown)


pkg/controller/drain.go
```diff
@@ -313,13 +314,12 @@ func (o *DrainOptions) deleteOrEvictPods(pods []api.Pod) error {
 }
 
 func (o *DrainOptions) evictPods(pods []api.Pod, policyGroupVersion string, getPodFn func(namespace, name string) (*api.Pod, error)) error {
-	doneCh := make(chan bool, len(pods))
-	errCh := make(chan error, 1)
+	returnCh := make(chan error, 1)
```
hardikdr: Should the buffered channel be initialized to len(pods) , as there could be more elements in it?

pkg/controller/drain.go
```diff
@@ -328,46 +328,49 @@ func (o *DrainOptions) evictPods(pods []api.Pod, policyGroupVersion string, getP
...
 	}
-	for {
+	globalTimeoutCh := time.After(globalTimeout)
+	numPods := len(pods)
+	for doneCount < numPods {
 		select {
-		case err := <-errCh:
-			return err
-		case <-doneCh:
+		case err := <-returnCh:
```
hardikdr: Will the error while evicting the pod is considered as doneCount++? This is intentional right? 
I am not sure of the best behavior here, as the pod then will anyways be not processed later, so we can keep it this way.

pkg/controller/drain.go
ggaurav10: This logic is taken from:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/drain/drain.go#L354

I guess we would end up serialising the "exiting of Go routines" also, instead of serialising just the "incrementing `doneCount`".

pkg/controller/drain.go
ggaurav10: Yes, error is considered for the increment. Taken from here:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/drain/drain.go#L397

Open to better options. The errors are just logged, and basically noop.

pkg/controller/drain.go
hardikdr: I see. Not sure, at the lowest level how golang would behave, but I understand if by any chance more than one elements are piped into the channel, it would panic. 
Let's keep it this way then.

pkg/controller/drain.go
hardikdr: Okay, I think then makes sense to count error'ed eviction as done.

PR #273, opened 2019-05-28
Fix packet issues

deitch:
**What this PR does / why we need it**:

Fixes several issues with the Packet MCM:

* Tags were assumed to be a key-value map on input and then converted to an array of string. This was inconsistent with what Gardener passed to it, and added complexity. These are now a simple string array.
* SSH Keys were assumed to be a key-value map containing the key UUID and the fingerprint, but Packet doesn't consume anything other than the UUID. This also was inconsistent with what Gardener passed to it and added complexity. These are now a simple string array of the UUIDs.
* The SSH Keys were not being passed to Packet API in the driver. Now it is.

**Special notes for your reviewer**:
As discussed with @rfranzke on Slack.

**Do not merge this yet,** until I indicate that it is ready and remove the `[WIP]`. In addition to the above issues, there is a bug wherein it crashes. Once that is fixed, this will be ready to go.

EDITED: good to merge.

**Release note**:

```improvement operator
Fixes packet driver
```

PR #275, opened 2019-06-04
Serialise eviction for pods with PVs

### LINKED ISSUE #262: Serialize the eviction of pods with volumes
**Feature (What you would like to be added):**
MCM to intelligently handle the eviction of pods with volumes

**Motivation (Why is this needed?):**
Rolling update of nodes take longer when more volumes are attached to the node. An average of 10+ volumes are attached to the Nodes in the SEED cluster, rolling updates taking longer directly impacts the downtime of ETCD.
Similar impact on the Shoot cluster workloads is possible  

**Approach/Hint to the implement solution (optional):**
One option could be to implement a drain controller in MCM - yet to be detailed out

ggaurav10:
Currently all pods are evicted in parallel. This PR changes this behaviour for pods with PVs. The pods with PVs will be evicted serially. Eviction of next pod will start after volume for previously evicted pod is detached from the node.
This improves the machine drain time.

Fixes #262 

**Release note**:

```improvement operator
When draining a node, pods with PVCs are evicted serially. Next pod eviction waits for PV of previously evicted pod to detach from the node. Operators can expect faster machine drains for their machines.
```


hardikdr:
thanks for the PR.
@ggaurav10  This branch seems to have conflicting files, can you please resolve them.

ggaurav10:
> thanks for the PR.
> @ggaurav10 This branch seems to have conflicting files, can you please resolve them.

done.

pkg/controller/drain.go
```diff
@@ -373,6 +387,236 @@ func (o *DrainOptions) evictPods(pods []api.Pod, policyGroupVersion string, getP
...
+		}
+	}
+	glog.V(4).Info("Detached PVs: ", pvNames)
+	detachCh <- true
+}
+
+func (o *DrainOptions) getVolIDsFromDriver(pvNames []string) ([]string, error) {
+	pvSpecs := []corev1.PersistentVolumeSpec{}
+	for _, pvName := range pvNames {
+		pv, err := o.client.CoreV1().PersistentVolumes().Get(pvName, metav1.GetOptions{})
```
hardikdr: If we miss out on the volume due to temp-issues of APIServer, do we skip waiting for that volume? I dont have inclination towards any approach, just pointing out the observed behaviour. :-P

pkg/driver/driver.go
```diff
@@ -28,6 +28,7 @@ type Driver interface {
 	Delete() error
 	GetExisting() (string, error)
 	GetVMs(string) (VMs, error)
+	GetVolNames([]corev1.PersistentVolumeSpec) ([]string, error)
```
hardikdr: I really wish, if there could be a generic way to fetch this information and we dont require providers to implement yet another method. Or we can think of this as an optional feature, where only those providers looking for a serialised eviction implement it.

pkg/controller/drain.go
```diff
@@ -373,6 +387,236 @@ func (o *DrainOptions) evictPods(pods []api.Pod, policyGroupVersion string, getP
...
+	}
+	glog.V(4).Info("Volume names from driver: ", volumeIDs)
+
+	found := true
+	for found {
+		found = false
+		node, err := o.client.Core().Nodes().Get(nodeName, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			glog.V(4).Info("Node not found: ", nodeName)
+			detachCh <- true
```
amshuman-kr: This is technically not true. It might be better to make `detachCh` `chan error` and emit error here and log it on the consuming side.

pkg/controller/drain.go
```diff
@@ -373,6 +387,236 @@ func (o *DrainOptions) evictPods(pods []api.Pod, policyGroupVersion string, getP
...
+
+		glog.V(4).Info("Attached volumes: ", node.Status.VolumesAttached)
+		attachedVols := node.Status.VolumesAttached
+		if len(attachedVols) == 0 {
+			glog.V(4).Info("No volumes attached to the node")
+			break
+		}
+
+		for _, volumeID := range volumeIDs {
+			for _, attachedVol := range attachedVols {
```
amshuman-kr: Same shallow copy issue as [above](https://github.com/gardener/machine-controller-manager/pull/275/files#diff-face99aebd3cd79577003f7df747c863R335) but with less impact because the number of volumes per pod will be less.

pkg/controller/drain.go
```diff
@@ -373,6 +387,236 @@ func (o *DrainOptions) evictPods(pods []api.Pod, policyGroupVersion string, getP
...
+				found, _ = regexp.MatchString(volumeID, string(attachedVol.Name))
```
amshuman-kr: Do we need `regexp.MatchString`? Won't a simpler comparison suffice?

pkg/driver/driver_alicloud.go
```diff
@@ -286,3 +286,18 @@ func (c *AlicloudDriver) toInstanceTags(tags map[string]string) ([]ecs.RunInstan
 
 	return result, nil
 }
+
+// GetVolNames parses volume names from pv specs
+func (c *AlicloudDriver) GetVolNames(specs []corev1.PersistentVolumeSpec) ([]string, error) {
+	names := []string{}
+	for _, spec := range specs {
```
amshuman-kr: Same shallow copy issue as [above](https://github.com/gardener/machine-controller-manager/pull/275/files#diff-face99aebd3cd79577003f7df747c863R335) but with less impact because the number of volumes per pod will be less.

pkg/driver/driver_aws.go
```diff
@@ -305,3 +305,17 @@ func (d *AWSDriver) decodeMachineID(id string) string {
 	splitProviderID := strings.Split(id, "/")
 	return splitProviderID[len(splitProviderID)-1]
 }
+
+// GetVolNames parses volume names from pv specs
+func (d *AWSDriver) GetVolNames(specs []corev1.PersistentVolumeSpec) ([]string, error) {
+	names := []string{}
+	for _, spec := range specs {
```
amshuman-kr: Same shallow copy issue as [above](https://github.com/gardener/machine-controller-manager/pull/275/files#diff-face99aebd3cd79577003f7df747c863R335) but with less impact because the number of volumes per pod will be less.

pkg/driver/driver_azure.go
```diff
@@ -598,3 +598,17 @@ func (d *AzureDriver) decodeMachineID(id string) string {
 	splitProviderID := strings.Split(id, "/")
 	return splitProviderID[len(splitProviderID)-1]
 }
+
+// GetVolNames parses volume names from pv specs
+func (d *AzureDriver) GetVolNames(specs []corev1.PersistentVolumeSpec) ([]string, error) {
+	names := []string{}
+	for _, spec := range specs {
```
amshuman-kr: Same shallow copy issue as [above](https://github.com/gardener/machine-controller-manager/pull/275/files#diff-face99aebd3cd79577003f7df747c863R335) but with less impact because the number of volumes per pod will be less.

pkg/driver/driver_gcp.go
```diff
@@ -329,3 +329,17 @@ func extractProject(serviceaccount []byte) (string, error) {
 	}
 	return j.Project, nil
 }
+
+// GetVolNames parses volume names from pv specs
+func (d *GCPDriver) GetVolNames(specs []corev1.PersistentVolumeSpec) ([]string, error) {
+	names := []string{}
+	for _, spec := range specs {
```
amshuman-kr: Same shallow copy issue as [above](https://github.com/gardener/machine-controller-manager/pull/275/files#diff-face99aebd3cd79577003f7df747c863R335) but with less impact because the number of volumes per pod will be less.

pkg/driver/driver_openstack.go
```diff
@@ -453,3 +453,17 @@ func (d *OpenStackDriver) recentImageIDFromName(client *gophercloud.ServiceClien
 	}
 	return "", fmt.Errorf("could not find an image id for image name %s", imageName)
 }
+
+// GetVolNames parses volume names from pv specs
+func (d *OpenStackDriver) GetVolNames(specs []corev1.PersistentVolumeSpec) ([]string, error) {
+	names := []string{}
+	for _, spec := range specs {
```
amshuman-kr: Same shallow copy issue as [above](https://github.com/gardener/machine-controller-manager/pull/275/files#diff-face99aebd3cd79577003f7df747c863R335) but with less impact because the number of volumes per pod will be less.

pkg/driver/driver_packet.go
```diff
@@ -172,3 +172,26 @@ func (d *PacketDriver) decodeMachineID(id string) string {
 	splitProviderID := strings.Split(id, "/")
 	return splitProviderID[len(splitProviderID)-1]
 }
+
+func packetTagsMapToString(tags map[string]string) []string {
```
amshuman-kr: Is this used somewhere?

pkg/driver/driver_packet.go
```diff
@@ -172,3 +172,26 @@ func (d *PacketDriver) decodeMachineID(id string) string {
...
+	ret := make([]string, 0)
+	for k, v := range tags {
+		ret = append(ret, fmt.Sprintf("%s:%s", k, v))
+	}
+	return ret
+}
+
+func packetTagsStringToMap(tags []string) map[string]string {
+	ret := map[string]string{}
```
amshuman-kr: Is this used somewhere?

pkg/controller/drain.go
```diff
@@ -373,6 +387,236 @@ func (o *DrainOptions) evictPods(pods []api.Pod, policyGroupVersion string, getP
...
+		time.Sleep(5 * time.Second)
+	}
+	return
+}
+
+func (o *DrainOptions) getPvs(pod corev1.Pod) []string {
+	pvs := []string{}
+	for _, vol := range pod.Spec.Volumes {
+		if vol.PersistentVolumeClaim != nil {
+			pvc, err := o.client.CoreV1().PersistentVolumeClaims(pod.Namespace).Get(vol.PersistentVolumeClaim.ClaimName, metav1.GetOptions{})
```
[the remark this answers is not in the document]
amshuman-kr: Can we not replace this GET call in a loop with a single LIST call outside the loop with a field selector? This way the number of calls as well as the associated retries can be reduced. But we have to be careful about pagination.

pkg/controller/drain.go
```diff
@@ -373,6 +387,236 @@ func (o *DrainOptions) evictPods(pods []api.Pod, policyGroupVersion string, getP
...
+		glog.Errorf("Error when getting volume IDs from the driver. Err: %v", err)
+		detachCh <- false
+		return
+	}
+	glog.V(4).Info("Volume names from driver: ", volumeIDs)
+
+	found := true
+	for found {
+		found = false
+		node, err := o.client.Core().Nodes().Get(nodeName, metav1.GetOptions{})
```
amshuman-kr: Low-prio: Why not a WATCH on the node instead of the polling?

pkg/controller/drain.go
```diff
@@ -373,6 +387,236 @@ func (o *DrainOptions) evictPods(pods []api.Pod, policyGroupVersion string, getP
...
+			break
+		}
+		glog.V(4).Info("Eviction for some pods will be retried after 5 seconds.")
+		pods = retryPods
+		time.Sleep(5 * time.Second)
+	}
+	return
+}
+
+func (o *DrainOptions) getPvs(pod corev1.Pod) []string {
```
amshuman-kr: Why not return a deep structure including the volume ids (with a nested call to the drivers) and avoid repeated GET PVC call [below](https://github.com/gardener/machine-controller-manager/pull/275/files#diff-face99aebd3cd79577003f7df747c863R572)?

pkg/controller/drain.go
amshuman-kr: Please make the driver call in [`getPVs`](https://github.com/gardener/machine-controller-manager/pull/275/files#diff-face99aebd3cd79577003f7df747c863R502) and skip this func altogether.

amshuman-kr:
@ggaurav10 Please note that my suggestion for clubbing the multiple GET PVC calls into a single LIST PVC with field selector is not a high-prio one. But the change of data structure and the moving of driver call into the `getPvs` func is.

pkg/controller/drain.go
```diff
@@ -373,6 +387,236 @@ func (o *DrainOptions) evictPods(pods []api.Pod, policyGroupVersion string, getP
...
+			case <-time.After(o.PvDetachTimeout):
+				returnCh <- fmt.Errorf("Timout when evicting pod: %v", pod)
+			}
+		}
+		if len(retryPods) == 0 {
+			break
+		}
+		glog.V(4).Info("Eviction for some pods will be retried after 5 seconds.")
+		pods = retryPods
+		time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
ggaurav10: sure. will make it a command line option - `podEvictionRetryInterval`.

pkg/driver/driver_packet.go
ggaurav10: This got added after rebasing. Its already there in master branch

pkg/driver/driver.go
hardikdr: I think it should fall back to the previous behavior of parallel detach if nil[empty] is returned here. The current code will skip those volumes hence skip waiting, and eventually evict pods quickly. So it's not exactly the same but very similar to the previous behavior, which seems fine to me.

pkg/controller/drain.go
```diff
@@ -373,6 +401,313 @@ func (o *DrainOptions) evictPods(pods []api.Pod, policyGroupVersion string, getP
...
+				volList = append(volList, vol)
+			}
+		}
+		pvMap[pod] = volList
+	}
+	glog.V(4).Info("Removed shared volumes. Filtered list of pods with volumes: ", pvMap)
+}
+
+func (o *DrainOptions) evictPodsWithPv(pods []*corev1.Pod, policyGroupVersion string, getPodFn func(namespace, name string) (*api.Pod, error), returnCh chan error) {
+	detachCh := make(chan error, 1)
```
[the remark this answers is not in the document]
ggaurav10: Closing the channel before making sure that all the go routines writing to this channel are done with their task will lead to panic.
As discussed, `waitForDetach()` function is now called in-line instead of as a go routine, and moved the timeout check inside the function.

pkg/controller/drain.go
```diff
@@ -313,39 +338,42 @@ func (o *DrainOptions) deleteOrEvictPods(pods []api.Pod) error {
...
+		}
+		if !hasPv {
+			podsWithoutPv = append(podsWithoutPv, pod)
+		}
+	}
+	return podsWithPv, podsWithoutPv
+}
+
 func (o *DrainOptions) evictPods(pods []api.Pod, policyGroupVersion string, getPodFn func(namespace, name string) (*api.Pod, error)) error {
 	returnCh := make(chan error, 1)
```
amshuman-kr:
```suggestion
	returnCh := make(chan error, 1)
	defer close(returnCh)
```

ggaurav10: If this channel is closed before all the go routines are done writing to this channel, it could lead to panic. I suggest we leave this channel while returning, and let it be taken care of by the GC, as is done [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/drain/drain.go#L354) from where most of the drain logic is imported:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/drain/drain.go#L354

WDYT?

pkg/driver/driver_packet.go
prashanth26: I think it has been again removed from the master by this https://github.com/gardener/machine-controller-manager/pull/273/files. Maybe we need to rebase again?

pkg/driver/driver_packet.go
prashanth26: I think it has been again removed from the master by this https://github.com/gardener/machine-controller-manager/pull/273/files. Maybe we need to rebase again?

pkg/controller/drain.go
ggaurav10: Also, as mentioned in [The Go Programming Language](https://www.gopl.io/):
> You needn't close every channel when you've finished with it. It's only necessary to close a channel when it is important to tell the receiving goroutines that all data have been sent. A channel that the garbage collector determinies to be unreachable will have its resources reclaimed whether or not it is closed.

amshuman-kr: It might be ok to not close the channel, but the point is why would any go routines be running after the outer func has returned? Shouldn't we make sure they don't? That is the actual review comment. Just like the last time ;-)

pkg/controller/drain.go
```diff
@@ -289,6 +314,10 @@ func (o *DrainOptions) evictPod(pod api.Pod, policyGroupVersion string) error {
 
 // deleteOrEvictPods deletes or evicts the pods on the api server
 func (o *DrainOptions) deleteOrEvictPods(pods []api.Pod) error {
+	// Closing stop channel makes sure that all go routines started later are stopped.
+	stopCh := make(chan bool, 1)
```
amshuman-kr:
```suggestion
	stopCh := make(chan struct{})
```

pkg/controller/drain.go
```diff
@@ -373,7 +405,340 @@ func (o *DrainOptions) evictPods(pods []api.Pod, policyGroupVersion string, getP
...
+				// Found PVC; append and exit
+				pvs = append(pvs, pvc.Spec.VolumeName)
+				break
+			}
+		}
+	}
+	return pvs, nil
+}
+
+func (o *DrainOptions) waitForDetach(volumeIDs []string, nodeName string, stopCh chan bool) error {
```
amshuman-kr:
```suggestion
func (o *DrainOptions) waitForDetach(volumeIDs []string, nodeName string, stopCh <-chan struct{}) error {
```

pkg/controller/drain.go
```diff
@@ -373,7 +405,340 @@ func (o *DrainOptions) evictPods(pods []api.Pod, policyGroupVersion string, getP
...
+			// Found PV; append and exit
+			pvSpecs = append(pvSpecs, pv.Spec)
+			break
+		}
+	}
+
+	return o.Driver.GetVolNames(pvSpecs)
+}
+
+func (o *DrainOptions) evictPodInternal(pod *corev1.Pod, policyGroupVersion string, getPodFn func(namespace, name string) (*api.Pod, error), returnCh chan error, stopCh chan bool) {
```
amshuman-kr:
```suggestion
func (o *DrainOptions) evictPodInternal(pod *corev1.Pod, policyGroupVersion string, getPodFn func(namespace, name string) (*api.Pod, error), returnCh chan error, stopCh <-chan struct{}) {
```

pkg/controller/drain.go
```diff
@@ -373,7 +405,340 @@ func (o *DrainOptions) evictPods(pods []api.Pod, policyGroupVersion string, getP
...
+
+	_, err = o.waitForDelete(podArray, Interval, time.Duration(math.MaxInt64), true, getPodFn, stopCh)
+	if err == nil {
+		returnCh <- nil
+	} else {
+		returnCh <- fmt.Errorf("error when waiting for pod %q terminating: %v", pod.Name, err)
+	}
+}
+
+func (o *DrainOptions) deletePods(pods []api.Pod, getPodFn func(namespace, name string) (*api.Pod, error), stopCh chan bool) error {
```
amshuman-kr:
```suggestion
func (o *DrainOptions) deletePods(pods []api.Pod, getPodFn func(namespace, name string) (*api.Pod, error), stopCh <-chan struct{}) error {
```

pkg/controller/drain.go
```diff
@@ -387,14 +752,20 @@ func (o *DrainOptions) deletePods(pods []api.Pod, getPodFn func(namespace, name
 			return err
 		}
 	}
-	_, err := o.waitForDelete(pods, Interval, globalTimeout, false, getPodFn)
+	_, err := o.waitForDelete(pods, Interval, globalTimeout, false, getPodFn, stopCh)
 	return err
 }
 
-func (o *DrainOptions) waitForDelete(pods []api.Pod, interval, timeout time.Duration, usingEviction bool, getPodFn func(string, string) (*api.Pod, error)) ([]api.Pod, error) {
+func (o *DrainOptions) waitForDelete(pods []api.Pod, interval, timeout time.Duration, usingEviction bool, getPodFn func(string, string) (*api.Pod, error), stopCh chan bool) ([]api.Pod, error) {
```
amshuman-kr:
```suggestion
func (o *DrainOptions) waitForDelete(pods []api.Pod, interval, timeout time.Duration, usingEviction bool, getPodFn func(string, string) (*api.Pod, error), stopCh <-chan struct{}) ([]api.Pod, error) {
```

ggaurav10:
> How are we planning to address the unit/integration tests? In this PR or another one? In either case, we should not make a release without some tests (especially, integration tests).

Let's raise a new PR for test cases.

hardikdr:
Okay then, merging this one.

PR #276, opened 2019-06-04
Fix unit-tests

hardikdr:
Unit-tests seems to be flaking because current taint-support does not guarantee the order of taints. This PR fixes a couple of tests which are assuming the ordered taints.

Fixes #

**Release note**:

```improvement operator

```

### DIFF
--- a/pkg/controller/deployment_rolling_test.go
+++ b/pkg/controller/deployment_rolling_test.go
@@ -116,7 +116,7 @@ var _ = Describe("deployment_rolling", func() {
 				for _, expectedNode := range data.expect.nodes {
 					actualNode, err := controller.targetCoreClient.CoreV1().Nodes().Get(expectedNode.Name, metav1.GetOptions{})
 					Expect(err).ToNot(HaveOccurred())
-					Expect(actualNode.Spec.Taints).Should(Equal(expectedNode.Spec.Taints))
+					Expect(actualNode.Spec.Taints).Should(ConsistOf(expectedNode.Spec.Taints))
 				}
 
 			},
--- a/pkg/controller/machine_util_test.go
+++ b/pkg/controller/machine_util_test.go
@@ -17,7 +17,6 @@ package controller
 
 import (
 	"encoding/json"
-	"reflect"
 
 	"github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
 	machinev1 "github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
@@ -82,7 +81,7 @@ var _ = Describe("machine_util", func() {
 				updatedNodeObject, _ := c.targetCoreClient.Core().Nodes().Get(nodeObject.Name, metav1.GetOptions{})
 
 				if data.expect.node != nil {
-					Expect(updatedNodeObject.Spec.Taints).Should(Equal(data.expect.node.Spec.Taints))
+					Expect(updatedNodeObject.Spec.Taints).Should(ConsistOf(data.expect.node.Spec.Taints))
 					Expect(updatedNodeObject.Labels).Should(Equal(data.expect.node.Labels))
 
 					// ignore LastAppliedALTAnnotataion
@@ -1099,7 +1098,7 @@ var _ = Describe("machine_util", func() {
 
 				waitForCacheSync(stop, c)
 
-				Expect(reflect.DeepEqual(testNode.Spec.Taints, expectedNode.Spec.Taints)).To(Equal(true))
+				Expect(testNode.Spec.Taints).Should(ConsistOf(expectedNode.Spec.Taints))
 				Expect(taintsChanged).To(Equal(data.expect.taintsChanged))
 			},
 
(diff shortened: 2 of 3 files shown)


pkg/controller/controller_suite_test.go
```diff
@@ -555,6 +555,25 @@ func waitForCacheSync(stop <-chan struct{}, controller *controller) {
 	)).To(BeTrue())
 }
 
+func EqualTaints(a, b []corev1.Taint) bool {
```
hardikdr: Not sure, if it makes sense to have such function here, open for suggestions.
@ggaurav10 - this is from our conversation yesterday.

ggaurav10: lgtm. Given that array of taints is not ordered, `reflect.DeepEqual()` cannot be used.
Just as a suggestion: WDYT about generic `assert.Equal()`.
https://github.com/stretchr/testify#assert-package

The PR also works though.

hardikdr: Ah yes, I think it could be used, but wondering if it has a dependency on the `testing` package, not sure if we can use otherwise.

pkg/controller/controller_suite_test.go
```diff
@@ -555,6 +555,25 @@ func waitForCacheSync(stop <-chan struct{}, controller *controller) {
...
+	if len(a) != len(b) {
+		return false
+	}
+	for _, va := range a {
+		found := false
+		for j := range b {
+			if va == b[j] {
```
ggaurav10: Does this comparison work for `Taint.TimeAdded` too?

pkg/controller/controller_suite_test.go
amshuman-kr: `testing` package is already depended on by `ginkgo` and `gomega`.

pkg/controller/controller_suite_test.go
amshuman-kr: 👍 Maybe a `DeepEquals` is better here.

hardikdr: For test cases, we are not setting TimeAdded, and hence it should be fine. 
Although it's a good point,  I am not sure - how would it really work overall with taint feature.

pkg/controller/controller_suite_test.go
amshuman-kr: But since `gomega` is already being used, please use [`ConsistOf`](https://onsi.github.io/gomega/#consistofelement-interface).

hardikdr: Perfect, I missed this. I can use this in all occurrences where taints are being compared. 
thanks for the suggestion.

hardikdr: Here, we want both slices to be exactly equal[ignoring order], so we should check the length separately right?

ggaurav10: How about [`MatchYAML`](https://onsi.github.io/gomega/#matchyamlyaml-interface)?

hardikdr: I just made changes with ConsistOf.  :D

amshuman-kr: > Here, we want both slices to be exactly equal[ignoring order], so we should check the length separately right? or using a separate method could be a better option. , is there a provided-matcher for that as well.

Length check is [built into](https://github.com/onsi/gomega/blob/master/matchers/consist_of.go#L42) `ConsistOf`. So additional length check is not required.

pkg/controller/deployment_rolling_test.go
```diff
@@ -116,7 +116,8 @@ var _ = Describe("deployment_rolling", func() {
 				for _, expectedNode := range data.expect.nodes {
 					actualNode, err := controller.targetCoreClient.CoreV1().Nodes().Get(expectedNode.Name, metav1.GetOptions{})
 					Expect(err).ToNot(HaveOccurred())
-					Expect(actualNode.Spec.Taints).Should(Equal(expectedNode.Spec.Taints))
+					Expect(actualNode.Spec.Taints).Should(HaveLen(len(expectedNode.Spec.Taints)))
```
amshuman-kr: Length check is [built into](https://github.com/onsi/gomega/blob/master/matchers/consist_of.go#L42) `ConsistOf`. So additional length check is not required.

pkg/controller/machine_util_test.go
```diff
@@ -82,7 +81,8 @@ var _ = Describe("machine_util", func() {
 				updatedNodeObject, _ := c.targetCoreClient.Core().Nodes().Get(nodeObject.Name, metav1.GetOptions{})
 
 				if data.expect.node != nil {
-					Expect(updatedNodeObject.Spec.Taints).Should(Equal(data.expect.node.Spec.Taints))
+					Expect(updatedNodeObject.Spec.Taints).Should(HaveLen(len(data.expect.node.Spec.Taints)))
```
amshuman-kr: Same as [above](https://github.com/gardener/machine-controller-manager/pull/276/files#diff-26d5394127cb7967bb5043caeb2ca98cR119).

pkg/controller/machine_util_test.go
```diff
@@ -1099,7 +1099,8 @@ var _ = Describe("machine_util", func() {
 
 				waitForCacheSync(stop, c)
 
-				Expect(reflect.DeepEqual(testNode.Spec.Taints, expectedNode.Spec.Taints)).To(Equal(true))
+				Expect(testNode.Spec.Taints).Should(HaveLen(len(expectedNode.Spec.Taints)))
```
amshuman-kr: Same as [above](https://github.com/gardener/machine-controller-manager/pull/276/files#diff-26d5394127cb7967bb5043caeb2ca98cR119).

Issue #281, opened 2019-06-18
Forceful deletion of Machine


hardikdr:
## Problem
We recently removed the support for forceful deletion of the machines in the cluster. Forceful deletion allowed to skip the drain while deleting the kubernetes-nodes. 

We need to investigate the behavior of MCM while deleting the Shoot-cluster especially when target-APIServer is not reachable. 
- I believe MCM keeps failing/attempting to drain the node, in turn blocking the deletion of the machine even while shoot-deletion. 

## Solution
The probable solution could be to re-check the reasoning behind disabling the forceful deletion and enable it again if possible.

prashanth26:
One reason why we had restored to always calling a drain before deletion of node was to make sure there was no race condition between MCM and Kubernetes in attaching/detaching PVCs attached to a node being deleted.

But to address the problem of drains failing for cluster's whose APIServer is not reachable, would it be safe to ignore calling drain for cluster's whose APIServer is not reachable? Not sure, just think out loud here.

PR #283, opened 2019-06-24
Enhancement/force deletion

prashanth26:
This PR always allows deletion of machines with `force-delete` labels. This avoids failure of deletion calls for clusters where the APIServer is not reachable.

Fixes #https://github.com/gardener/machine-controller-manager/issues/281

I have divided the commits for ease of review

**Release note**:

```improvement operator
Force deletion of machine succeeds even on drain failures
```

### DIFF
--- a/pkg/controller/machine.go
+++ b/pkg/controller/machine.go
@@ -511,26 +511,27 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
 				errBuf,
 			)
 			err = drainOptions.RunDrain()
-			if err != nil {
+			if err == nil {
+				// Drain successful
+				glog.V(2).Infof("Drain successful for machine %q. \nBuf:%v \nErrBuf:%v", machine.Name, buf, errBuf)
 
-				if machine.Labels["force-deletion"] == "True" {
-					// Continue with the deletion of VM without draining
-				} else {
-					// If it's not a force-deletion return error
-					lastOperation := v1alpha1.LastOperation{
-						Description:    "Drain failed - " + err.Error(),
-						State:          v1alpha1.MachineStateFailed,
-						Type:           v1alpha1.MachineOperationDelete,
-						LastUpdateTime: metav1.Now(),
-					}
-					c.updateMachineStatus(machine, lastOperation, machine.Status.CurrentStatus)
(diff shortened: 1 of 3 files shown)


pkg/controller/machine.go
```diff
@@ -512,17 +512,23 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
...
+						LastUpdateTime: metav1.Now(),
+					}
+					c.updateMachineStatus(machine, lastOperation, machine.Status.CurrentStatus)
+
+					// Machine still tries to terminate after drain failure
+					glog.Warningf("Drain failed for machine %q \nBuf:%v \nErrBuf:%v \nErr-Message:%v", machine.Name, buf, errBuf, err)
+					return err
+				}
 			}
 			glog.V(2).Infof("Drain successful for machine %q \nBuf:%v \nErrBuf:%v", machine.Name, buf, errBuf)
```
ggaurav10: Misleading in case drain wasn't successful && it is force-deletion

pkg/controller/machine.go
```diff
@@ -512,17 +512,23 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
 			)
 			err = drainOptions.RunDrain()
 			if err != nil {
```
ggaurav10: why not `if err != nil && machine.Labels["force-deletion"] != "True" {`

pkg/controller/machine.go
prashanth26: Nice catch. Shall fix this.

pkg/controller/machine.go
prashanth26: Makes more sense. Shall make this change.

prashanth26: Made suggested changes - https://github.com/gardener/machine-controller-manager/pull/283/commits/52e1e875559ddd0aeb2c6c74f6dd1d931d54bd57

pkg/controller/machine.go
prashanth26: made suggested changes - https://github.com/gardener/machine-controller-manager/pull/283/commits/52e1e875559ddd0aeb2c6c74f6dd1d931d54bd57

PR #288, opened 2019-07-04
Feature/drain termination grace period

### LINKED ISSUE #262: Serialize the eviction of pods with volumes

amshuman-kr:
Changes the drain logic to honour the individual `pods`' `terminationGracePeriodSeconds` instead of a **fixed** (though configurable) `timeout`.
* For the `pods` without volumes, the timeout would be a bit longer than the maximum of `terminationGracePeriodSeconds` of the `pods` without volume.
* For the `pods` with volumes, the timeout is its own `terminationGracePeriodSeconds` and a **fixed** (but configurable) volume detach timeout for each pod.

This enables the drain logic to honour the `terminationGracePeriod` contract with the individual `pods` which would be important for larger workloads that might require substantial grace periods during termination.

The command-line flags `machine-max-evict-retries` and `machine-pv-detach-timeout` can be used to customize the default behaviour of the number of retries for eviction and wait period for the detaching of volumes after pods are evicted/deleted.

The command-line flag `machine-drain-timeout` has been deprecated (has no effect) and will be removed in a subsequent release.

Fixes #262 

This PR started out as a basis for discussion with @ggaurav10 about the approach to start honouring `terminationGracePeriodSecords` during drain. But now seems to be at a stage for a wider review.

The global timeout for drain has been increased to a default value of `12h` to cover the needs most workloads. In this range the workloads can specify the actual needed grace period in the standard `terminationGracePeriodSeconds` field in the `pod` specification.

**Release note**:

```noteworthy operator
Changes the drain logic to honour the individual `pods`' `terminationGracePeriodSeconds` instead of a fixed(though configurable) `timeout`. For the `pods` without volumes, the timeout would be a bit longer than the maximum of `terminationGracePeriodSeconds` of the `pods` without volume. For the `pods` with volumes, the timeout is its own `terminationGracePeriodSeconds` and a **fixed** (but configurable) volume detach timeout for each pod.

The command-line flags `machine-max-evict-retries` and `machine-pv-detach-timeout` can be used to customize the default behaviour of the number of retries for eviction and wait period for the detaching of volumes after pods are evicted/deleted.

The default value for the existing command-line flag `machine-drain-timeout` has been increased to `12h` which should be enough for most work-loads.
```


prashanth26:
On a quick look over the PR, there is a linting issue while running the check script in the CI pipeline.
```
./cmd/machine-controller-manager/app/options/options.go:36:2: a blank import should be only in a main or test package, or have a comment justifying it
```

pkg/controller/drain.go
```diff
@@ -369,50 +368,57 @@ func filterPodsWithPv(pods []api.Pod) ([]*api.Pod, []*api.Pod) {
...
+func (o *DrainOptions) getGlobalTimeoutForPodsWithoutPV(pods []*api.Pod) time.Duration {
+	var tgpsMax time.Duration
+	for _, pod := range pods {
+		tgps := o.getTerminationGracePeriod(pod)
+		if tgps > tgpsMax {
+			tgpsMax = tgps
+		}
+	}
+
+	return time.Duration(tgpsMax)*time.Second + PodsWithuotPVDrainGracePeriod
```
prashanth26: `getTerminationGracePeriod()` already returns the Period(int) multiplied by seconds right? We are again multiplying it here with seconds? A mistake I am guessing?

pkg/controller/drain.go
```diff
@@ -79,6 +80,13 @@ const (
 	// EvictionSubresource is the kind used for evicting pods
 	EvictionSubresource = "pods/eviction"
 
+	// DefaultMaxEvictRetries is the default value for MaxEvictRetries
+	DefaultMaxEvictRetries = int32(3)
+
+	// PodsWithuotPVDrainGracePeriod defines the grace period to wait for the pods without PV during machine drain.
+	// This is in addition to the maximum terminationGracePeriod amount the pods.
+	PodsWithuotPVDrainGracePeriod = 3 * time.Minute
```
prashanth26: Typo for without
```suggestion
	PodsWithoutPVDrainGracePeriod = 3 * time.Minute
```

pkg/controller/drain.go
```diff
@@ -487,87 +493,110 @@ func filterSharedPVs(pvMap map[string][]string) {
...
 }
 
+func (o *DrainOptions) evictPodsWithPVInternal(attemptEvict bool, pods []*corev1.Pod, volMap map[string][]string,
+	policyGroupVersion string,
+	getPodFn func(namespace, name string) (*api.Pod, error),
+	returnCh chan error) (remainingPods []*api.Pod, fastTrack bool) {
+	var retryPods []*api.Pod
+	for i, pod := range pods {
+		var err error
+		if attemptEvict {
```
prashanth26:
```suggestion
		if attemptEvict && len(policyGroupVersion) > 0 {
```

prashanth26: With regards to handling k8s versions that don't support eviction?

pkg/controller/drain.go
```diff
@@ -700,26 +725,29 @@ func (o *DrainOptions) getVolIDsFromDriver(pvNames []string) ([]string, error) {
...
-			glog.Warningf("The caller function returned. No need to try evicting pods now")
-			return
-		default:
+	nretries := int(o.MaxEvictRetries)
+	for i := 0; ; i++ {
+		if i >= nretries {
+			attemptEvict = false
+		}
+
+		if attemptEvict {
```
prashanth26:
```suggestion
		if attemptEvict && len(policyGroupVersion) > 0 {
```
Same as above.

pkg/controller/drain.go
```diff
@@ -700,26 +725,29 @@ func (o *DrainOptions) getVolIDsFromDriver(pvNames []string) ([]string, error) {
...
-		err = o.evictPod(*pod, policyGroupVersion)
 
 		if err == nil {
 			break
 		} else if apierrors.IsNotFound(err) {
 			glog.V(3).Info("\t", pod.Name, " evicted")
 			returnCh <- nil
 			return
-		} else if apierrors.IsTooManyRequests(err) {
+		} else if attemptEvict && apierrors.IsTooManyRequests(err) {
```
prashanth26: Is this the case hit in case of PDB? Just out of curiosity.

pkg/options/types.go
```diff
@@ -94,9 +94,12 @@ type SafetyOptions struct {
 	// Timeout (in durartion) used while health-check of
 	// a machine before it is declared as failed
 	MachineHealthTimeout metav1.Duration
-	// Timeout (in durartion) used while draining of machine before deletion,
+	// Deprecated. No effect. Timeout (in durartion) used while draining of machine before deletion,
```
prashanth26: typo
```suggestion
	// Deprecated. No effect. Timeout (in duration) used while draining of machine before deletion,
```

pkg/controller/drain.go
prashanth26: Typo for without
```suggestion
	return time.Duration(tgpsMax)*time.Second + PodsWithoutPVDrainGracePeriod
```

pkg/controller/drain.go
```diff
@@ -52,15 +51,17 @@ type DrainOptions struct {
...
-	PvDetachTimeout    time.Duration
-	DeleteLocalData    bool
-	nodeName           string
-	Out                io.Writer
-	ErrOut             io.Writer
-	Driver             driver.Driver
-	pvcLister          corelisters.PersistentVolumeClaimLister
-	pvLister           corelisters.PersistentVolumeLister
+	// Timeout has been deprecated and has no effect.
+	Timeout         time.Duration
```
hardikdr: I am somehow skeptical about not keeping the global-timeout. @ggaurav10 raised a nice point, while offline discussion.

Basically, an uninformed user can halt the rollout or make it really long for certain machines, by keeping extremely long gracefulTimePeriod on Pods. I was hoping more of a solution, which gives lesser control to the user, especially in terms of managing machines.

I can remember having certain counter-arguments over this concern, can you please delineate in case you are aware of any.

amshuman-kr: @hardikdr @ggaurav10 I share the concerns about not having a defence against misconfigured long `terminationGracePeriodSeconds`. But not honouring `terminationGracePeriodSeconds` is also not an option. Otherwise we would be opting not to support some genuine use-cases (some heavy stateful workload).

So, the main question .  is what is the best way to defend. I can think of several ways including a global timeout. Please note: these need not be mutually exclusive.
1. Global timeout
What value would be good? `1/2/6/12/24h`? 
I am a bit reluctant about this because of the side effects on the complications on the code as well as undesirable choices it creates. Also, this goes against fail-fast principle.
1. Validate and enforce upper limit on `terminationGracePeriods` via validating webhooks or some other similar lateral mechanism (`PodPresets`?)
Fails fast. It catches the problem early when the pod is deployed and not when it is evicted and then it leaves a mess like above. For flexibility, the upper limit for `terminationGracePeriods` could be configurable in the `Shoot` resource. It basically acts like a declarative policy. I prefer this the most.
1. Expose granular metrics about`terminationGracePeriodSeconds`, actual termination duration, volume detach duration, successful pod eviction/volume detach within timeout, failed pod eviction/volume detach within timeout etc. at the granularity of both the pods and the node. This can act as diagnostics both for alerts as well as help devops in identifying the problem during support.

There could be more. WDYT?

pkg/controller/drain.go
```diff
@@ -331,15 +338,7 @@ func (o *DrainOptions) deleteOrEvictPods(pods []api.Pod) error {
...
 
-	if len(policyGroupVersion) > 0 {
-		err := o.evictPods(pods, policyGroupVersion, getPodFn, stopCh)
-		if err != nil {
-			glog.Warningf("Pod eviction was timed out, Error: %v. \nHowever, drain will continue to forcefully delete the pods by setting graceful termination period to 0s", err)
-		} else {
-			return nil
-		}
-	}
-	return o.deletePods(pods, getPodFn, stopCh)
```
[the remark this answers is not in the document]
amshuman-kr: Both the code paths are supported in the line [below](https://github.com/gardener/machine-controller-manager/pull/288/files#diff-face99aebd3cd79577003f7df747c863R341).

return o.evictPods(_**len(policyGroupVersion) > 0**_, pods, policyGroupVersion, getPodFn)

pkg/controller/drain.go
ggaurav10: >Global timeout
What value would be good? 1/2/6/12/24h?
I am a bit reluctant about this because of the side effects on the complications on the code as well as undesirable choices it creates. Also, this goes against fail-fast principle.

In case the global timeout (1/2/6/12/24h) is less than the timeout value calculated using `terminationGraecPeriodSeconds`, we would still `fail fast`er.

Although I agree that 2nd option adheres to `fail fast` principle better, however, the reluctance of having complications in the code or making undesirable choices would still exist; that would just be moved to another component. WDYT?

pkg/controller/drain.go
amshuman-kr: Already covered [above](https://github.com/gardener/machine-controller-manager/pull/288/files#diff-face99aebd3cd79577003f7df747c863R341).

pkg/controller/drain.go
amshuman-kr: Already covered [above](https://github.com/gardener/machine-controller-manager/pull/288/files#diff-face99aebd3cd79577003f7df747c863R341).

pkg/controller/drain.go
amshuman-kr: Yes. This was already existing code. I was suspicious about it that this error is only for rate limiting. But @ggaurav10 found documentation that this is the error that is returned if PD is violated.

pkg/options/types.go
amshuman-kr: Fixed.

pkg/controller/drain.go
```diff
@@ -487,87 +493,110 @@ func filterSharedPVs(pvMap map[string][]string) {
...
-		time.Sleep(PodEvictionRetryInterval)
+	} else {
+		remainingPods, _ = o.evictPodsWithPVInternal(false, pods, volMap, policyGroupVersion, getPodFn, returnCh)
 	}
 
-	for j := 0; j < remainingPods; j++ {
+	// Placate the caller by returning the nil status for the remaining pods.
+	for _, pod := range remainingPods {
 		glog.V(4).Info("Returning success for remaining pods")
 		// This is executed when node is not found anymore.
```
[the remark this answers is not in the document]
amshuman-kr: Fixed.

pkg/controller/drain.go
```diff
@@ -487,87 +493,110 @@ func filterSharedPVs(pvMap map[string][]string) {
...
-				// We should anyway return successful eviction for the remaining pods
-				break EvictingPods
-			} else if err != nil {
-				glog.Errorf("Error when waiting for volume to detach from node. Err: %v", err)
-				returnCh <- err
-				continue
-			}
-			glog.V(3).Info("Detached pv for pod: ", pod.Name)
-			returnCh <- nil
+			glog.V(4).Info("Eviction for some pods will be retried after ", PodEvictionRetryInterval)
```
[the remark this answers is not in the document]
amshuman-kr: Fixed.

pkg/controller/drain.go
amshuman-kr: Fixed.

pkg/controller/drain.go
amshuman-kr: Fixed.

pkg/controller/drain.go
amshuman-kr: @ggaurav10 Yes. But by fail fast, I meant when the pod is created (fastest 😉)  rather than when the pod is about to be evicted. But it would be *faster*, yes 😄

amshuman-kr:
@ggaurav10 @prashanth26 @hardikdr Thanks for the review comments. I have addressed them and pushed a commit. Can you PTAL?

amshuman-kr:
Can someone please add the do-no-merge label? We can remove it once all the testing is completed.

pkg/controller/drain.go
prashanth26: Okay. Thanks for the clarification.

cmd/machine-controller-manager/app/options/options.go
```diff
@@ -100,8 +102,9 @@ func (s *MCMServer) AddFlags(fs *pflag.FlagSet) {
...
 	fs.DurationVar(&s.SafetyOptions.MachineHealthTimeout.Duration, "machine-health-timeout", s.SafetyOptions.MachineHealthTimeout.Duration, "Timeout (in durartion) used while re-joining (in case of temporary health issues) of machine before it is declared as failed.")
-	fs.DurationVar(&s.SafetyOptions.MachineDrainTimeout.Duration, "machine-drain-timeout", s.SafetyOptions.MachineDrainTimeout.Duration, "Timeout (in durartion) used while draining of machine before deletion, beyond which MCM forcefully deletes machine.")
-	fs.DurationVar(&s.SafetyOptions.PvDetachTimeout.Duration, "machine-pv-detach-timeout", s.SafetyOptions.PvDetachTimeout.Duration, "Timeout (in duration) used for detach of PV while evicting pods")
+	fs.DurationVar(&s.SafetyOptions.MachineDrainTimeout.Duration, "machine-drain-timeout", controller.DefaultMachineDrainTimeout, "Timeout (in durartion) used while draining of machine before deletion, beyond which MCM forcefully deletes machine.")
+	fs.Int32Var(&s.SafetyOptions.MaxEvictRetries, "machine-max-evict-retries", controller.DefaultMaxEvictRetries, "Maximum number of times evicts would be attempted on a pod before it is forcibly deleted during draining of a machine.")
```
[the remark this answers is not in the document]
amshuman-kr: Other flags use defaults in `make start`. Any particular reason for adding `machine-max-evict-retries` there explicitly?

amshuman-kr:
I have reintroduced the global timeout (defaults to 12h) and also added some unit tests.

@prashanth26 @ggaurav10 Could you please PTAL?

Two cases are currently excluded, I will enable them once I get them working.

Also, I found one of `machine_safety` controller tests failing as below (most likely unrelated to my changes).
```
#machine_safety ##reconcileClusterMachineSafetyAPIServer [It] Control APIServer: UnReachable, Target APIServer: Reachable, Inactive Timer: Started, Pre-Frozen: false = Post-Frozen: false 
/Users/i344628/go/src/github.com/gardener/machine-controller-manager/vendor/github.com/onsi/ginkgo/extensions/table/table_entry.go:46

  Expected
      <bool>: true
  to equal
      <bool>: false

  /Users/i344628/go/src/github.com/gardener/machine-controller-manager/pkg/controller/machine_safety_test.go:723
```

amshuman-kr:
> Two cases are currently excluded, I will enable them once I get them working.

All the 10 tests for drain are now working. Will need to add some negative cases. But will do in a separate PR in the interest of time.

ggaurav10:
> Some minor suggestions, and the following query:
> If I understood right, we keep track of global timeout in memory [here](https://github.com/gardener/machine-controller-manager/pull/288/files#diff-face99aebd3cd79577003f7df747c863R154). If MCM restarts then we would loose the track of time, and the timeout will start from 0 again, right? If yes, should we keep it that way?

What about this?

prashanth26:
@ggaurav10 As discussed offline./

- https://github.com/amshuman-kr/machine-controller-manager/blob/2c1d58cfc314fbd131cd23f01ef83516671a24e0/pkg/controller/machine.go#L488-L490
- https://github.com/amshuman-kr/machine-controller-manager/blob/2c1d58cfc314fbd131cd23f01ef83516671a24e0/pkg/controller/machine.go#L507-L523

So if you see, the timeout is computed based on the difference between `lastUpdateTime` of the status and `drainTimeout`. Hence should be consistent even after restarts.

amshuman-kr:
>>Some minor suggestions, and the following query:
>>If I understood right, we keep track of global timeout in memory here. If MCM restarts then we would loose the track of time, and the timeout will start from 0 again, right? If yes, should we keep it that way?

>What about this?

#### Short Answer
- This was the same behaviour earlier as well, right? Drain time was never persisted in any resources, be it `Machine` or `Node`. If mcm restarted, then all the drains will start from the beginning again. This has not changed in this new logic.
- I used the `drainStartedOn` variable only to calculate the deadline for the `mainContext` accurately to account for the time taken to reach till the `mainContext` is created, that's all.
- So, I think it is ok to let it continue to be that way for the time being.

#### Long Answer
- I do not like the way we do evict/delete and waiting for pod/volume detach right now at all. It is not good for reconciliation iteration to spawn further goroutines. It is especially bad to spawn goroutines from reconcilation just to wait for stuff to finish.
- It would be much better to do the following over multiple cycles of reconciliation with each iteration picking up from where the previous reconciliation left off until either the drain is completed or the timeout happens.
  1. Record the start of drain in status/annotation of `Machine`/`Node`. The start of the first drain attempt could be inferred from the `deletionTimestamp`.
  1. Trigger all the evicts/deletes for the pods not yet evicted/deleted synchronously in the reconciliation iteration. 
  1. Update the status/annotations of `Machine`/`Node` to record the pods still pending eviction/deletion and re-enqueue the resource for retry later.
- I believe this would also reduce a little bit the current spike in load on `kube-apiserver` when machines are drained.
- But I think we should pick this topic separately as an improvement later based on priority.

WDYT?

PR #300, opened 2019-08-02
Split force deletion

prashanth26:

Fixes #

**Release note**:

```improvement operator
Bugfix: Allows force deletion of pods who have no controllers backing them
```

### DIFF
--- a/pkg/controller/drain.go
+++ b/pkg/controller/drain.go
@@ -48,22 +48,23 @@ import (
 
 // DrainOptions are configurable options while draining a node before deletion
 type DrainOptions struct {
-	client             kubernetes.Interface
-	Force              bool
-	GracePeriodSeconds int
-	IgnoreDaemonsets   bool
-	Timeout            time.Duration
-	MaxEvictRetries    int32
-	PvDetachTimeout    time.Duration
-	DeleteLocalData    bool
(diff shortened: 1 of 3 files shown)


pkg/controller/drain.go
```diff
@@ -48,22 +48,23 @@ import (
...
-	nodeName           string
-	Out                io.Writer
-	ErrOut             io.Writer
-	Driver             driver.Driver
-	pvcLister          corelisters.PersistentVolumeClaimLister
-	pvLister           corelisters.PersistentVolumeLister
-	drainStartedOn     time.Time
-	drainEndedOn       time.Time
+	client                       kubernetes.Interface
+	ForceDeletePods              bool
```
hardikdr: I think these `drainOptions` were inspired from here: https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#drain

Shouldn't the behavior be consistent with what's documented there? Eg, Keep the `force` flag and let it be used for force-deleting non-controller-backed pods. We could add a new flag to add diff feature.
- Or is it that way and I missed something?

prashanth26: Considering it is just a variable name change for an unexposed variable. I am going to treat this change as unacceptable.

PR #302, opened 2019-08-07
Regenerated API Types because the latest Version is not checked in

afritzler:
Not sure if this is intended, but `hack/generate-code` on `master` regenerates all the types for me. Is this just my setup?

Environment:
```
go version go1.12.7 darwin/amd64
```

**Release note**:
```improvement operator
NONE
```


prashanth26:
I just cross-checked. Same happened in my setup. Probably one of the vendors was updated and code being generated has some updates. Thanks for the PR.

PR #303, opened 2019-08-07
Removed optional annotation in internal types

afritzler:
Removed `+optional` from internal API types since it is not needed there.

**Release note**:
```improvement operator
NONE
```

### DIFF
--- a/pkg/apis/machine/types.go
+++ b/pkg/apis/machine/types.go
@@ -65,23 +65,18 @@ type MachineList struct {
 type MachineSpec struct {
 
 	// Class contains the machineclass attributes of a machine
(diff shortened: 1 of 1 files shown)


prashanth26:
Hi @afritzler ,

So you mean to say that this is only required in external types right? We can surely remove it from here and not affect us?

afritzler:
Well you could try this out by running the `code-gen` to see that nothing changes.

@rfranzke correct me if I am wrong but the `+optional` is only used in the versioned type definition.

afritzler:
The `optional` annotation is only relevant for the versioned OpenAPI generation.

Issue #305, opened 2019-08-07
Improve logs while draining of machines


amshuman-kr:
### Issue

The current logs related to the drain logic mostly mention only the names of the pods being acted upon. This makes live tracking/monitoring/debugging hard to identify which machine/node is affected.

### Solution

Include the machine/node name in the logs related to the drain logic in addition to the names of the affected pods.

prashanth26:
Need to still log the deletion/eviction/detachment times for Pods/PVs.

prashanth26:
Closing in favor of - https://github.com/gardener/machine-controller-manager/pull/315

PR #309, opened 2019-08-12
Improve logs while draining

hardikdr:
**What this PR does / why we need it**: This PR improves the log while draining - adds node-name wherever needed.

**Release note**:

```improvement operator

```

### DIFF
--- a/pkg/controller/drain.go
+++ b/pkg/controller/drain.go
@@ -564,9 +564,9 @@ func (o *DrainOptions) evictPodsWithPv(attemptEvict bool, pods []*corev1.Pod,
 			// Return success to caller for all non-processed pods so that the caller function can move on.
 			returnCh <- nil
 		} else if attemptEvict {
-			returnCh <- fmt.Errorf("Error evicting pod %s/%s", pod.Namespace, pod.Name)
+			returnCh <- fmt.Errorf("Error evicting pod %s/%s from node %s", pod.Namespace, pod.Name, pod.Spec.NodeName)
 		} else {
-			returnCh <- fmt.Errorf("Error deleting pod %s/%s", pod.Namespace, pod.Name)
+			returnCh <- fmt.Errorf("Error deleting pod %s/%s from node %s", pod.Namespace, pod.Name, pod.Spec.NodeName)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/controller/drain.go
```diff
@@ -603,21 +603,21 @@ func (o *DrainOptions) evictPodsWithPVInternal(attemptEvict bool, pods []*corev1
...
 		} else if err != nil {
-			glog.V(4).Infof("Error when evicting pod: %v/%v. Will be retried. Err: %v", pod.Namespace, pod.Name, err)
+			glog.V(4).Infof("Error when evicting pod: %v/%v from node %v. Will be retried. Err: %v", pod.Namespace, pod.Name, pod.Spec.NodeName, err)
 			retryPods = append(retryPods, pod)
 			continue
 		}
 
 		// Eviction was successful. Wait for pvs for this pod to detach
-		glog.V(3).Info("Eviction/delete succeeded. Waiting for the volumes to detach: ", pod.Name)
+		glog.V(3).Info("Eviction/delete succeeded. Waiting for the volumes to detach: ", pod.Name, pod.Spec.NodeName)
```
prashanth26: We need to capture the time taken for eviction/deletion of the pod here. We might have to introduce a time variable for the same.

pkg/controller/drain.go
```diff
@@ -632,7 +632,7 @@ func (o *DrainOptions) evictPodsWithPVInternal(attemptEvict bool, pods []*corev1
 			returnCh <- err
 			continue
 		}
-		glog.V(3).Info("Detached pv for pod: ", pod.Name)
+		glog.V(3).Info("Detached pv for pod: ", pod.Name, " from node ", pod.Spec.NodeName)
```
prashanth26: We need to capture the time taken for detachment of PV here. We might have to introduce a time variable for the same or reuse the one used for PV eviction logic above.

pkg/controller/drain.go
```diff
@@ -785,14 +785,14 @@ func (o *DrainOptions) evictPodWithoutPVInternal(attemptEvict bool, pod *corev1.
 		if err == nil {
 			break
 		} else if apierrors.IsNotFound(err) {
-			glog.V(3).Info("\t", pod.Name, " evicted")
+			glog.V(3).Info("\t", pod.Name, " evicted from node ", pod.Spec.NodeName)
```
prashanth26: We need to capture the time taken for eviction/deletion of the pod here. We might have to introduce a time variable for the same.

hardikdr:
Agreed, we need more improvements in general for logging including the one mentioned above. I suggest, let's take it in subsequent PRs, and scope this one with current quick change.

prashanth26:
Okay. Shall accept these changes but keep the issue - #305 open to track the same.

PR #311, opened 2019-08-12
Added additional OpenStack credentials for DomainID, TenantID, UserDomainName

afritzler:
This PR adds additional support for
```
DomainID
TenantID
UserDomainName
```
in the OpenStack machine class secret.

This PR is part of https://github.com/gardener/gardener-extensions/issues/154

**Release note**:
```improvement operator
Added additional OpenStack credentials for DomainID, TenantID, UserDomainName
```


pkg/driver/driver_openstack.go
```diff
@@ -356,16 +282,21 @@ func (d *OpenStackDriver) createNeutronClient() (*gophercloud.ServiceClient, err
...
 	}
+
+	// optional OS_USER_DOMAIN_NAME
+	userDomainName := d.CloudConfig.Data[v1alpha1.OpenStackUserDomainName]
+
 	domainName, ok := d.CloudConfig.Data[v1alpha1.OpenStackDomainName]
-	if !ok {
-		return nil, fmt.Errorf("missing %s in secret", v1alpha1.OpenStackDomainName)
+	domainID, ok2 := d.CloudConfig.Data[v1alpha1.OpenStackDomainID]
+	if !ok && !ok2 {
```
prashanth26: So both of them have to be present right? Just confirming.

pkg/driver/driver_openstack.go
```diff
@@ -356,16 +282,21 @@ func (d *OpenStackDriver) createNeutronClient() (*gophercloud.ServiceClient, err
...
+		return nil, fmt.Errorf("missing %s or %s in secret", v1alpha1.OpenStackDomainName, v1alpha1.OpenStackDomainID)
 	}
 	tenantName, ok := d.CloudConfig.Data[v1alpha1.OpenStackTenantName]
-	if !ok {
-		return nil, fmt.Errorf("missing %s in secret", v1alpha1.OpenStackTenantName)
+	tenantID, ok2 := d.CloudConfig.Data[v1alpha1.OpenStackTenantID]
+	if !ok && !ok2 {
```
prashanth26: Same as above.

pkg/driver/driver_openstack.go
afritzler: At least one of them.

afritzler:
Validation would be great. But we only validate the `OpenStackMachineClassSpec` and not the infrastructure secret. Since this change only touches the secret and not the machine spec I currently don't see a point where we could add validation here.

prashanth26:
Okay makes sense. Merging this PR then.

PR #315, opened 2019-08-13
Further improved logs while draining

prashanth26:
Improved logs to display pod/PV drain times

Fixes #309 

**Release note**:

```improvement operator

```

### DIFF
--- a/pkg/controller/drain.go
+++ b/pkg/controller/drain.go
@@ -191,9 +191,9 @@ func (o *DrainOptions) deleteOrEvictPodsSimple() error {
 		if newErr != nil {
 			return newErr
 		}
-		glog.Errorf("There are pending pods when an error occurred: %v \nList of pending pods", err)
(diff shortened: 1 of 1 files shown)


pkg/controller/drain.go
```diff
@@ -303,7 +313,7 @@ func (o *DrainOptions) getPodsForDeletion() (pods []api.Pod, err error) {
 		return []api.Pod{}, errors.New(fs.Message())
 	}
 	if len(ws) > 0 {
-		fmt.Fprintf(o.ErrOut, "WARNING: %s\n", ws.Message())
+		glog.Warningf("%s", ws.Message())
```
hardikdr: Can you please check if o.ErrOut is really required to be removed?

prashanth26: Reverted the changes W.R.T to fmt.Fprintf. As this is used by the logger to log messages.

PR #316, opened 2019-08-14
Deletes nodes object on machine deletion

### LINKED ISSUE #186: Deletion of node object once VM is deleted
## Issue
Currently, the KCM seems to be having a bug where sometimes it is not able to delete the node object when the VM is deleted. 

## Possible solutions
1. Trace back the issue with the community and fix it there on Kubernetes upstream
2. Force MCM for node object deletion once VM deletion returns successfully.

prashanth26:
Deletes node object once the machine object is deleted. This removes any hanging node objects in the cluster. 

Fixes #186 

**Release note**:

```improvement user
Deletes nodes object on machine (object) deletion
```

### DIFF
--- a/pkg/controller/machine.go
+++ b/pkg/controller/machine.go
@@ -566,6 +566,14 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
 			return err
 		}
 
+		// Delete node object
+		err = c.targetCoreClient.CoreV1().Nodes().Delete(nodeName, &metav1.DeleteOptions{})
+		if err != nil && !apierrors.IsNotFound(err) {
+			// If its an error, and anyother error than object not found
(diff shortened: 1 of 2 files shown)


pkg/controller/machine.go
```diff
@@ -566,11 +566,22 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
 			return err
 		}
 
+		// Remove finalizers from machine object
 		c.deleteMachineFinalizers(machine)
```
amshuman-kr: Shouldn't we do this until after node is deleted? To make sure reconciliation can take care of failed node deletion?

prashanth26: Yes. My bad. I shall do it the other way.

prashanth26: Fixed it here - https://github.com/gardener/machine-controller-manager/pull/316/commits/b7b6b395be405ccf6a899bf19cbc88370103884d

PR #320, opened 2019-08-21
Bugfix/delete unhealthy machines

### LINKED ISSUE #306: Draining of node in unknown state
## Issue
When a node is in an unknown state, the waiting for termination of pod/PVs keeps failing. However, the deletion goes through when the drain timeout occurs. (After 12hours)

## Solution
We need to handle the situation for unknown node drains separately. We could have a shorter drain timeout for nodes in an unknown state.

prashanth26:

Fixes #306 

**Release note**:

```noteworthy user
Fixes issues while draining nodes in unknown state.
```
```improvement user
Fixes issues while overriding termination grace periods for pods with larger values.
```

### DIFF
--- a/pkg/controller/drain.go
+++ b/pkg/controller/drain.go
@@ -852,8 +852,8 @@ func (o *DrainOptions) evictPodWithoutPVInternal(attemptEvict bool, pod *corev1.
 	podArray := []*api.Pod{pod}
 
 	timeout := o.getTerminationGracePeriod(pod)
-	if timeout < o.Timeout {
-		glog.V(3).Infof("Overriding large termination grace period (%s) for the pod %s/%s", timeout.String(), pod.Namespace, pod.Name)
+	if timeout > o.Timeout {
+		glog.V(3).Infof("Overriding large termination grace period (%s) for the pod %s/%s and setting it to %s", timeout.String(), pod.Namespace, pod.Name, o.Timeout)
 		timeout = o.Timeout
 	}
 
--- a/pkg/controller/drain.go
+++ b/pkg/controller/drain.go
@@ -857,7 +857,8 @@ func (o *DrainOptions) evictPodWithoutPVInternal(attemptEvict bool, pod *corev1.
 		timeout = o.Timeout
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/controller/drain.go
```diff
@@ -852,12 +852,13 @@ func (o *DrainOptions) evictPodWithoutPVInternal(attemptEvict bool, pod *corev1.
 	podArray := []*api.Pod{pod}
 
 	timeout := o.getTerminationGracePeriod(pod)
-	if timeout < o.Timeout {
-		glog.V(3).Infof("Overriding large termination grace period (%s) for the pod %s/%s", timeout.String(), pod.Namespace, pod.Name)
+	if timeout > o.Timeout {
```
ggaurav10: This check should also be there for pods with PV right?

In function `evictPodsWithPVInternal`, we have a main context with global timeout let's say 12 hours. After successful eviction/deletion, we have another local context with timeout = `o.getTerminationGracePeriod(pod)+o.PvDetachTimeout)` let's say 20 hours

But while waiting for PV detach we only check for the local context and not the global context. This makes the global timeout passed ineffective.

ggaurav10: [Global context](https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/drain.go#L600) should we passed in [this](https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/drain.go#L712) function and it should be checked [here](https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/drain.go#L724)

amshuman-kr: [`Context.WithTimeout`](https://golang.org/pkg/context/#WithTimeout) is supposed to be the same as `WithDeadline(parent, time.Now().Add(timeout))` and [`Context.WithDeadline`](https://golang.org/pkg/context/#WithDeadline) is not supposed to override parent's deadline if the parent's deadline is earlier.

So, technically, no change required. But I see no harm in making it explicit either.

ggaurav10: ah ok. cool then.
/lgtm

PR #322, opened 2019-08-22
Bugfix: Attempt deletion of node only if it isn't blank

prashanth26:

Fixes #

**Release note**:

```improvement operator
Bugfix: Attempt deletion of node only if it is present and isn't blank
```


pkg/controller/machine.go
```diff
@@ -562,6 +562,16 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
 			}
 
 			err = driver.Delete()
+
```
ggaurav10: The error for the above call is not there anymore because of this change.

prashanth26: Good catch, will fix it. Thanks.

PR #323, opened 2019-08-22
Rel-0.21.1-dev: Delete node only if nodeName isn't empty

prashanth26:

Fixes #

**Release note**:

```improvement operator
Bugfix: Delete node only if nodeName isn't empty
```


pkg/controller/machine.go
```diff
@@ -562,6 +562,16 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
 			}
 
 			err = driver.Delete()
+
```
ggaurav10: Error check not there anymore for the above call.

PR #324, opened 2019-08-22
Bugfix: Delete node only if nodeName isn't empty

prashanth26:
Better handling of error conditions

Fixes #

**Release note**:

```improvement operator
Bugfix: Delete node only if nodeName isn't empty
```

### DIFF
--- a/pkg/controller/machine.go
+++ b/pkg/controller/machine.go
@@ -583,12 +583,22 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
 			return err
 		}
 
-		// Delete node object
-		err = c.targetCoreClient.CoreV1().Nodes().Delete(nodeName, &metav1.DeleteOptions{})
-		if err != nil && !apierrors.IsNotFound(err) {
-			// If its an error, and anyother error than object not found
-			glog.Errorf("Deletion of Node Object %q failed due to error: %s", nodeName, err)
-			return err
+		if nodeName != "" {
+			// Delete node object
+			err = c.targetCoreClient.CoreV1().Nodes().Delete(nodeName, &metav1.DeleteOptions{})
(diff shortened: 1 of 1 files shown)


pkg/controller/machine.go
```diff
@@ -562,45 +562,60 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
...
 				err = c.targetCoreClient.CoreV1().Nodes().Delete(nodeName, &metav1.DeleteOptions{})
 				if err != nil && !apierrors.IsNotFound(err) {
+					message := fmt.Sprintf("Deletion of Node Object %q failed due to error: %s", nodeName, err)
+					lastOperation := v1alpha1.LastOperation{
+						Description:    message,
+						State:          v1alpha1.MachineStateFailed,
+						Type:           v1alpha1.MachineOperationDelete,
+						LastUpdateTime: metav1.Now(),
+					}
+					c.updateMachineStatus(machine, lastOperation, machine.Status.CurrentStatus)
```
ggaurav10: machine might have been deleted succesfully [here](https://github.com/gardener/machine-controller-manager/pull/324/files#diff-d8287fe74b5273163c9f6b6c635ad912R564). So this step would fail.

It is much easier to just add check `if nodeName != "" ` before deleting the node in the code before #322 was merged.

ggaurav10: I would suggest revert #322 and just add a the check mentioned above.

prashanth26: Only the VM is deleted above. The machine object still remains below right.

ggaurav10: Yes. Realised this just after posting my comments. :)

prashanth26:
IT is failing due to a known issue. Merging for now.

PR #326, opened 2019-08-23
Add URN field in AzureMachineClass

vpnachev:
With this PR, the azure controller will use the `urn` to define VM images, instead of the quadruple `publisher,offer,sku,version`. 
The goal of this change is to allow image version mapping, so that it will be possible to advertise version `X` as version `Y`, just like rest of the providers are doing.

Fixes #

You could a see discussion about this here https://github.com/gardener/gardener/pull/1284#discussion_r313946117

The change is related to https://github.com/gardener/gardener-extensions/pull/290

**Release note**:

```noteworthy operator
:warning: The `AzureMachineClass` now has a field `Spec.Properties.StorageProfile.ImageReference.URN` that is used to define the OS image for the VMs. This field is replacing the fields `Spec.Properties.StorageProfile.ImageReference.[Publisher|Offer|Sku|Version]` - they are marked as `DEPRECATED` and will be removed soon.
```


vpnachev:
@gardener/mcm-maintainers could you review this PR?

prashanth26:
Hi @vpnachev ,

Thanks for the PR and we really appreciate the tests. However, I was worried about backward compatibility. Isn't it a better idea if we support either or scenarios and validate the same in the validation?

vpnachev:
It is backward compatible, if `urn` is not provided then the `publisher`, `offer`, `sku` and `version` are used to default the urn.
Additionally, the resource is `alpha` so this change should be ok to it.

prashanth26:
> It is backward compatible, if urn is not provided then the publisher, offer, sku and version are used to default the urn.
That's not true right? It isn't backwaard compatible anymore - https://github.com/gardener/machine-controller-manager/pull/326/files#diff-fd5ac12351daf1f5d5b2eafe1ff57d9dR101-R124. Am I missing something?
> Additionally, the resource is alpha so this change should be ok to it.
Agreed. But was just concerned as Gardener could still be using an older version of azureMachineClasses to run old clusters where URN could not be present right?

vpnachev:
Thanks @prashanth26 , I got your concerns - you were right. I was thinking that the defaults are applied(api server extension resources), but the resources are actually CRDs, so there is no defaulting by default. Should I remove the defaulting code, wdyt?



> Agreed. But was just concerned as Gardener could still be using an older version of azureMachineClasses to run old clusters where URN could not be present right? So was just worried about that case.

Basically you are right, with https://github.com/gardener/gardener-extensions/pull/290 provider-azure will support `urn` as well. But it should not matter which one will be updated firstly - MCM or gardener-extensions, because for couple of release they should be compatible.

prashanth26:
> Thanks @prashanth26 , I got your concerns - you were right. I was thinking that the defaults are applied(api server extension resources), but the resources are actually CRDs, so there is no defaulting by default. Should I remove the defaulting code, wdyt?
I am not sure what is the best way to achieve something like deafulting with CRDs. Should we move this logic into the `azure_driver.go` instead? While creating machines?

vpnachev:
Defaulting of the `AzureMachineClasses` objects is removed because it is unused.
The `driver_azure.go` is trying to use the URN field if provided, otherwise it is using the `publisher`, `offer`, `sku` and `version` fields.

PS
> I am not sure what is the best way to achieve something like deafulting with CRDs

A [mutating webhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook) should be used for this.

prashanth26:
> Defaulting of the AzureMachineClasses objects is removed because it is unused.
The driver_azure.go is trying to use the URN field if provided, otherwise it is using the publisher, offer, sku and version fields.

Sounds good to me. The changes look good now.

pkg/driver/driver_azure.go
```diff
@@ -158,6 +160,30 @@ func (d *AzureDriver) getVMParameters(vmName string, networkInterfaceReferenceID
...
 
+func isUrnProvided(d *AzureDriver) bool {
+	return d.AzureMachineClass.Spec.Properties.StorageProfile.ImageReference.URN != nil &&
+		*d.AzureMachineClass.Spec.Properties.StorageProfile.ImageReference.URN != ""
+}
+
+func getAzureImageDetails(d *AzureDriver) (publisher, offer, sku, version string) {
+	if isUrnProvided(d) {
+		splits := strings.Split(*d.AzureMachineClass.Spec.Properties.StorageProfile.ImageReference.URN, ":")
+		if len(splits) != 4 {
```
ggaurav10: Minor: this check shouldn't be needed, as this is taken care of as part of validation.

vpnachev: removed

vpnachev:
The obsolete urn format check is removed.

PR #329, opened 2019-08-29
makes MCM GCP driver to differentiate user-data from startup-script

vlvasilev:
Additional user data on GCP is added by passing key/value pairs. The key is the name of the data and the value is the actual data itself.At the moment no matter what is the type of the data (cloud-config or script) we always set the key to be "user-data". But when GCP see such key it assumes that is dedicated to cloud-init binary so passing a script as a user-data and the corresponding binary is missing the script will to be ran. By making the MCM to differentiate between cloud-config and script now we can set the corresponding key to be either "user-data" or "startup-script". Thus the script will be executed on each boot after the network is up.
Fixes #

**Release note**:

```improvement operator
 MCM differentiates between cloud-config and script as user-data
```

### DIFF
--- a/pkg/driver/driver_gcp.go
+++ b/pkg/driver/driver_gcp.go
@@ -97,12 +97,9 @@ func (d *GCPDriver) Create() (string, string, error) {
 	}
 	instance.Disks = disks
 
-	var metadataItems = []*compute.MetadataItems{
-		{
-			Key:   "user-data",
-			Value: &d.UserData,
-		},
-	}
+	var metadataItems = []*compute.MetadataItems{}
+	metadataItems = append(metadataItems, d.getUserData())
+
 	for _, metadata := range d.GCPMachineClass.Spec.Metadata {
 		metadataItems = append(metadataItems, &compute.MetadataItems{
 			Key:   metadata.Key,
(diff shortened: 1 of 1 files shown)


pkg/driver/driver_gcp.go
```diff
@@ -97,12 +97,9 @@ func (d *GCPDriver) Create() (string, string, error) {
...
 	instance.Disks = disks
 
-	var metadataItems = []*compute.MetadataItems{
-		{
-			Key:   "user-data",
-			Value: &d.UserData,
-		},
-	}
+	var metadataItems = []*compute.MetadataItems{}
+	metadataItems = append(metadataItems, d.getUserData())
```
ggaurav10: Would a more appropriate approach be adding another field for startup-script in the `GCPDriver` struct instead of overloading `UserData`? Could both `user-data` and `startup-script` be used when creating a GCP instance?

hardikdr:
@vlvasilev thanks for the PR.
Can you please provide any active user-story/use-case to bring this change?  And how do we plan to consume it from Gardener?

vlvasilev:
@hardikdr 
Am trying to join a node (with centos 7 OS) to a shoot cluster. The image has not build in cloud-init, so I need to use a script for bootstrapping the download configuration or call it the donwload OperationSystemConfig data. The script is loaded under the name user-data but it is not ran. After looking in the [documentation ](https://cloud.google.com/compute/docs/startupscript) it seams that  the script must be loaded under the "startup-script" name. After renaming it everything went well.

pkg/driver/driver_gcp.go
vlvasilev: It is possible to use both but at the moment there is no use case to using them both. Currently they are using the same data. This is the data from the OperationSystemConfig which is translated to #cloud-config or script. There are neither flow or controller which can deal with "startup-script". So we can consider "startup-script" as an alias for "user-data" until there is mechanism to use both.

vpnachev:
Another use case of this feature is the SuSE JeOS on GCP - currently it hasn't support for `cloud-init`.
There is a PR https://github.com/gardener/gardener-extensions/pull/297 that is going to provide support for `startup` script for bootstrapping the VMs.

pkg/driver/driver_gcp.go
hardikdr: IMO this is specific to the cloud-provider/OS, and better be handled at drivers.

PR #330, opened 2019-09-10
Machine creation retries instead of return failure on failing to GET machine object

### LINKED ISSUE #284: Multiple machine creation issue
## Issue
It was observed that multiple VMs with the same node names tried to register themselves during machine creation. 

## Solution
It was observed that an error case was not handled properly in machine creation. Refer - https://github.com/gardener/machine-controller-manager/blob/master//pkg/controller/machine.go#L364. This line had to continue instead of returning.

prashanth26:

Fixes #284 

I have split the PR into commits to differentiate actual changes, test framework changes and test cases.

**Release note**:

```improvement operator
Machine creation retries instead of return failure on failing to GET machine object
```

### DIFF
--- a/pkg/controller/machine.go
+++ b/pkg/controller/machine.go
@@ -361,7 +361,8 @@ func (c *controller) machineCreate(machine *v1alpha1.Machine, driver driver.Driv
 		// Get the latest version of the machine so that we can avoid conflicts
 		machine, err := c.controlMachineClient.Machines(machine.Namespace).Get(machine.Name, metav1.GetOptions{})
 		if err != nil {
-			return err
+			glog.Warningf("Machine GET failed. Retrying, error: %s", err)
+			continue
 		}
 
 		lastOperation := v1alpha1.LastOperation{
@@ -391,7 +392,7 @@ func (c *controller) machineCreate(machine *v1alpha1.Machine, driver driver.Driv
 		clone.Spec.ProviderID = actualProviderID
 		machine, err = c.controlMachineClient.Machines(clone.Namespace).Update(clone)
 		if err != nil {
-			glog.Warningf("Machine update failed. Retrying, error: %s", err)
+			glog.Warningf("Machine UPDATE failed. Retrying, error: %s", err)
 			continue
 		}
 
@@ -401,7 +402,7 @@ func (c *controller) machineCreate(machine *v1alpha1.Machine, driver driver.Driv
 		clone.Status.CurrentStatus = currentStatus
 		_, err = c.controlMachineClient.Machines(clone.Namespace).UpdateStatus(clone)
 		if err != nil {
-			glog.Warningf("Machine/status update failed. Retrying, error: %s", err)
+			glog.Warningf("Machine/status UPDATE failed. Retrying, error: %s", err)
 			continue
 		}
 		// Update went through, exit out of infinite loop
--- a/pkg/controller/machine_test.go
+++ b/pkg/controller/machine_test.go
@@ -18,6 +18,7 @@ package controller
 import (
 	"errors"
 	"fmt"
+	"math"
 	"time"
 
 	machineapi "github.com/gardener/machine-controller-manager/pkg/apis/machine"
@@ -449,10 +450,11 @@ var _ = Describe("machine", func() {
 
 	Describe("#machineCreate", func() {
 		type setup struct {
-			secrets   []*corev1.Secret
-			aws       []*machinev1.AWSMachineClass
-			openstack []*machinev1.OpenStackMachineClass
-			machines  []*machinev1.Machine
+			secrets             []*corev1.Secret
+			aws                 []*machinev1.AWSMachineClass
+			openstack           []*machinev1.OpenStackMachineClass
+			machines            []*machinev1.Machine
+			fakeResourceActions *customfake.ResourceActions
 		}
 		type action struct {
 			machine        string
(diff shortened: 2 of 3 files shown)


pkg/controller/machine.go
```diff
@@ -361,7 +361,8 @@ func (c *controller) machineCreate(machine *v1alpha1.Machine, driver driver.Driv
 		// Get the latest version of the machine so that we can avoid conflicts
 		machine, err := c.controlMachineClient.Machines(machine.Namespace).Get(machine.Name, metav1.GetOptions{})
 		if err != nil {
-			return err
+			glog.Warningf("Machine GET failed. Retrying, error: %s", err)
```
hardikdr: Could the name of the machine be useful in the logs?

pkg/controller/machine.go
```diff
@@ -391,7 +392,7 @@ func (c *controller) machineCreate(machine *v1alpha1.Machine, driver driver.Driv
 		clone.Spec.ProviderID = actualProviderID
 		machine, err = c.controlMachineClient.Machines(clone.Namespace).Update(clone)
 		if err != nil {
-			glog.Warningf("Machine update failed. Retrying, error: %s", err)
+			glog.Warningf("Machine UPDATE failed. Retrying, error: %s", err)
```
hardikdr: Could the name of the machine be useful in the logs?

pkg/controller/machine.go
```diff
@@ -401,7 +402,7 @@ func (c *controller) machineCreate(machine *v1alpha1.Machine, driver driver.Driv
 		clone.Status.CurrentStatus = currentStatus
 		_, err = c.controlMachineClient.Machines(clone.Namespace).UpdateStatus(clone)
 		if err != nil {
-			glog.Warningf("Machine/status update failed. Retrying, error: %s", err)
+			glog.Warningf("Machine/status UPDATE failed. Retrying, error: %s", err)
```
hardikdr: same as above ....

pkg/controller/machine.go
prashanth26: Yes, makes sense. Made suggested changes. Refer - https://github.com/gardener/machine-controller-manager/pull/330/commits/c3b4dec3a8cd10096d8464fc3550999a630571d9.

pkg/controller/machine.go
prashanth26: Yes, makes sense. Made suggested changes. Refer - https://github.com/gardener/machine-controller-manager/pull/330/commits/c3b4dec3a8cd10096d8464fc3550999a630571d9.

pkg/controller/machine.go
prashanth26: Yes, makes sense. Made suggested changes. Refer - https://github.com/gardener/machine-controller-manager/pull/330/commits/c3b4dec3a8cd10096d8464fc3550999a630571d9.

prashanth26:
PTAL - https://github.com/gardener/machine-controller-manager/pull/330/commits/787908e285519755d6e4b1ee4a1c7b5bfe90403.

pkg/controller/machine.go
```diff
@@ -358,10 +358,12 @@ func (c *controller) machineCreate(machine *v1alpha1.Machine, driver driver.Driv
 	glog.V(2).Infof("Created machine: %q, MachineID: %s", machine.Name, actualProviderID)
 
 	for {
+		machineName := machine.Name
 		// Get the latest version of the machine so that we can avoid conflicts
 		machine, err := c.controlMachineClient.Machines(machine.Namespace).Get(machine.Name, metav1.GetOptions{})
 		if err != nil {
-			return err
+			glog.Warningf("Machine GET failed for %q. Retrying, error: %s", machineName, err)
+			continue
```
amshuman-kr: Would [`k8s.io/client-go/util/retry`](https://godoc.org/k8s.io/client-go/util/retry) be of use here?

pkg/controller/machine.go
```diff
@@ -391,7 +393,7 @@ func (c *controller) machineCreate(machine *v1alpha1.Machine, driver driver.Driv
 		clone.Spec.ProviderID = actualProviderID
 		machine, err = c.controlMachineClient.Machines(clone.Namespace).Update(clone)
 		if err != nil {
-			glog.Warningf("Machine update failed. Retrying, error: %s", err)
+			glog.Warningf("Machine UPDATE failed for %q. Retrying, error: %s", machineName, err)
 			continue
```
amshuman-kr: Same as [above](https://github.com/gardener/machine-controller-manager/pull/330/files#diff-d8287fe74b5273163c9f6b6c635ad912R366).

pkg/controller/machine.go
```diff
@@ -401,7 +403,7 @@ func (c *controller) machineCreate(machine *v1alpha1.Machine, driver driver.Driv
 		clone.Status.CurrentStatus = currentStatus
 		_, err = c.controlMachineClient.Machines(clone.Namespace).UpdateStatus(clone)
 		if err != nil {
-			glog.Warningf("Machine/status update failed. Retrying, error: %s", err)
+			glog.Warningf("Machine/status UPDATE failed for %q. Retrying, error: %s", machineName, err)
 			continue
```
amshuman-kr: Same as [above](https://github.com/gardener/machine-controller-manager/pull/330/files#diff-d8287fe74b5273163c9f6b6c635ad912R366).

pkg/controller/machine.go
prashanth26: Yes, @amshuman-kr. The retry loop could be used here. However, Martin had mentioned that we should avoid loops like these in the controller and should return an error and retry the entire reconciliation from scratch. I plan to get rid of all these loops once the OOT is over. So this change is to fix any issues only until then. 

So I could introduce the retry loop but then we would be getting rid of it in the future. It's your call.

PR #331, opened 2019-09-11
Use tzdata package instead of zoneinfo.zip

ialidzhikov:
- Use tzdata package instead of zoneinfo.zip (based on @adracus work in gardener/gardener#891)
- Update base image to alpine:3.10.

Fixes #

**Release note**:

```improvement operator
`tzdata` package is now used instead of `assets/zoneinfo.zip` to make all timezones available.
```
```improvement operator
Base docker image is updated to `alpine:3.10`.
```


prashanth26:
Thanks for the quick approval @jia-jerry .

PR #335, opened 2019-09-17
Add support for zoned Azure machines

### LINKED ISSUE #332: Azure: Add Availability Zone support
To support also `Machines` in Availability zones on Azure the Machine-Controller-Manager need to be able to deploy, bootstrap and manage Azure machines which are not assigned to an Azure AvailabilitySet.

Basically the `AzureMachineClass` resource need a field to configure a zone.
All `Machines` which reference an `AzureMachineClass` with a zone configuration should be deployed into the zone. MachineClasses which have a zone configuration shouldn't have an AvailabilitySet configuration.
This is an example how the `AzureMachineClass` could look like.
```yaml
apiVersion: machine.sapcloud.io/v1alpha1
kind: AzureMachineClass
metadata:
  name: test-machineclass-zone
  namespace: test
spec:
  location: westeurope
  resourceGroup: test
  secretRef: ...
  subnetInfo: ...
  tags: ...
  properties:
    zone: 1
    hardwareProfile: ...
    osProfile: ...
    storageProfile: ...
```

We still want to support `Machines` in an AvailabilitySet. That means the existing configuration needs to be supported too.
```yaml
apiVersion: machine.sapcloud.io/v1alpha1
kind: AzureMachineClass
metadata:
  name: test-machineclass-av-set
  namespace: test
spec:
  location: westeurope
  resourceGroup: test
  secretRef: ...
  subnetInfo: ...
  tags: ...
  properties:
    availabilitySet:
      id: /subscriptions/my-subscription-id/resourceGroups/test/providers/Microsoft.Compute/availabilitySets/test-workers
    hardwareProfile: ...
    osProfile: ...
    storageProfile: ...
```

In any case it should't be possible to use Availability Zone and AvailabilitySet in parallel. Probably we should throw an error when a user tries to create a `AzureMachineClass` with both configured or if a `Machine` reference such a `AzureMachineClass`.

Is related to: https://github.com/gardener/gardener-extensions/issues/135

cc @AndreasBurger, @MSSedusch, @rfranzke

dkistner:
Azure Machines can be deployed into a Zone OR an AvailabilitySet.
Both in parallel is not possible and will lead to an error.

To support zoned Azure Shoot clusters the MCM need to deploy and manage Azure machines in availability zones.

Fixes #332

**Release note**:

```improvement user
MCM supports now Azure machines deployed into availability zones. You can either deploy a machine into a zone or an AvailabilitySet.
```

### DIFF
--- a/pkg/driver/driver_azure.go
+++ b/pkg/driver/driver_azure.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"strconv"
 	"strings"
 	"sync"
 
@@ -150,11 +151,16 @@ func (d *AzureDriver) getVMParameters(vmName string, networkInterfaceReferenceID
 					},
 				},
 			},
-			AvailabilitySet: &compute.SubResource{
-				ID: &d.AzureMachineClass.Spec.Properties.AvailabilitySet.ID,
-			},
 		},
 		Tags: tagList,
+	}
+
+	if d.AzureMachineClass.Spec.Properties.Zone != nil {
+		VMParameters.Zones = &[]string{strconv.Itoa(*d.AzureMachineClass.Spec.Properties.Zone)}
+	} else if d.AzureMachineClass.Spec.Properties.AvailabilitySet != nil {
+		VMParameters.VirtualMachineProperties.AvailabilitySet = &compute.SubResource{
+			ID: &d.AzureMachineClass.Spec.Properties.AvailabilitySet.ID,
+		}
 	}
 
 	return VMParameters
--- a/kubernetes/machine_classes/azure-machine-class.yaml
+++ b/kubernetes/machine_classes/azure-machine-class.yaml
@@ -41,5 +41,6 @@ spec:
           publicKeys:
             path: "/path/to/public-key/in/machine" # Path to public key
             keyData: "public-key-data" # Public key data
-    availabilitySet:
+    zone: 1 # Zone which the node will be deployed into. Either a zone or an AvailabilitySet can be specified.
+    availabilitySet: # AvailabilitySet which the node will be assigned to. Either a zone or an AvailabilitySet can be specified.
       id: "/subscriptions/subscription-id/resourceGroups/resource-group-name/providers/Microsoft.Compute/availabilitySets/availablity-set-name" # ID of availability set to attach the machine to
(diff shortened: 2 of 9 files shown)


pkg/driver/driver_azure.go
```diff
@@ -329,6 +338,14 @@ func (d *AzureDriver) createVMNicDisk() (*compute.VirtualMachine, error) {
...
 	)
 
+	if d.AzureMachineClass.Spec.Properties.Zone == nil && d.AzureMachineClass.Spec.Properties.AvailabilitySet == nil {
+		return nil, errors.New("Machine need to be assigned to AvailabilityZone or an AvailabilitySet")
+	}
+
+	if d.AzureMachineClass.Spec.Properties.Zone != nil && d.AzureMachineClass.Spec.Properties.AvailabilitySet != nil {
+		return nil, errors.New("Machines assigned to AvailabilityZone and AvailabilitySet in parallel are not supported")
+	}
+
```
ggaurav10: These checks should be moved to `validation` package [here](https://github.com/gardener/machine-controller-manager/blob/rel-0.21.0/pkg/apis/machine/validation/azuremachineclass.go)

dkistner: Thanks wasn't aware of that. I will move the checks to the validation.

dkistner: I have updated the PR.
Unfortunately it is not possible to raise an error which reference multiple fields therefore I note `zone` and `availabilitySet` in the same field reference. Is that ok or do prefer a dedicated error per field? Which mean we would get two validation errors for one error situation.

pkg/driver/driver_azure.go
```diff
@@ -150,13 +151,20 @@ func (d *AzureDriver) getVMParameters(vmName string, networkInterfaceReferenceID
...
-			},
 		},
 		Tags: tagList,
 	}
 
+	if d.AzureMachineClass.Spec.Properties.Zone != nil {
+		VMParameters.Zones = &[]string{strconv.Itoa(*d.AzureMachineClass.Spec.Properties.Zone)}
+	}
+
+	if d.AzureMachineClass.Spec.Properties.AvailabilitySet != nil {
```
prashanth26: Hi @dkistner , 

One small change. Can we move this to `else if`. As with validation, this shouldn't be a problem. However, if anyone is just importing this package it could be a problem.

dkistner: yep makes sense done!

hardikdr:
thanks for the @dkistner, looks good. 
Really appreciate :)

PR #340, opened 2019-10-04
GCP VM's Public IP attachment is now optional

### LINKED ISSUE #339: Remove public IP's from gcp worker
Currently only gcp vm's get a public IP which is then restricted with a firewall rule that denies all traffic to this nodes except ports 443 and 80.

With the new kubernetes version 1.16.0 an new conformance test was introduced that tests if a service with a `NodePort` can be reached with its internal and external IP (only when a external IP is available). 
This tests fails on gcp as only 443 and 80 port is allowed and therefore Gardener is not currently not k8s conformant.

To unify the setup across all cloudproviders and make Gardener Clusters K8s conformant we have to remove the public IP form gcp workers.

cc @DockToFuture

prashanth26:

Fixes #339 

**Release note**:

```noteworthy operator
⚠️ GCP VM's Public IP attachment is now optional. Default behaviour is that public IP attachment is enabled. To disable this, refer to [this example here](https://github.com/gardener/machine-controller-manager/blob/master/kubernetes/machine_classes/gcp-machine-class.yaml#L27)
```

### DIFF
--- a/kubernetes/machine_classes/gcp-machine-class.yaml
+++ b/kubernetes/machine_classes/gcp-machine-class.yaml
@@ -24,7 +24,8 @@ spec:
   - key: my-key
     value: my-value
   networkInterfaces:
-  - network: network-name # Network name to attach the instance to
+  - disableExternalIP: false # If external IP is not required for the NIC, set this to true
+    network: network-name # Network name to attach the instance to
     subnetwork: sub-net-name # Subnet name to attach the instance to
   scheduling:
     automaticRestart: true # Automatic restart of instance
--- a/pkg/apis/machine/types.go
+++ b/pkg/apis/machine/types.go
@@ -1029,9 +1029,9 @@ type GCPMetadata struct {
 
 // GCPNetworkInterface describes network interfaces for GCP
 type GCPNetworkInterface struct {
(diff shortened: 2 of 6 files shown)


prashanth26:
Good catch ! I have addressed your concern here - https://github.com/gardener/machine-controller-manager/pull/340/commits/0562f649af0e74d22f7810db59a1e65180b21e2e. Also have updated the release notes accordingly.

kubernetes/machine_classes/gcp-machine-class.yaml
```diff
@@ -26,6 +26,7 @@ spec:
   networkInterfaces:
   - network: network-name # Network name to attach the instance to
     subnetwork: sub-net-name # Subnet name to attach the instance to
+    disableExternalIP: true # If external IP is not required for the NIC, set this to true
```
hardikdr: I'd suggest, we either keep this `false` in the documentation or rather mention in the comment about the consequences of node-joining.
- There is a chance of bad UX otherwise if someone tries MCM independently.

prashanth26: Done.

PR #341, opened 2019-10-09
Closed open channels

### LINKED ISSUE #337: CPU profiling of MCM
## Issue 
CPU usage on MCM seems to be increasing over time. We need figure out the cause for it.

prashanth26:

Fixes #337 

**Release note**:

```improvement operator
Fixes leaks in Go Routines leading to CPU wastage
```

### DIFF
--- a/pkg/controller/drain.go
+++ b/pkg/controller/drain.go
@@ -414,6 +414,7 @@ func (o *DrainOptions) getGlobalTimeoutForPodsWithoutPV(pods []*api.Pod) time.Du
 
 func (o *DrainOptions) evictPods(attemptEvict bool, pods []api.Pod, policyGroupVersion string, getPodFn func(namespace, name string) (*api.Pod, error)) error {
 	returnCh := make(chan error, len(pods))
+	defer close(returnCh)
 
 	if o.ForceDeletePods {
 		podsToDrain := make([]*api.Pod, len(pods))
--- a/pkg/controller/machineset.go
+++ b/pkg/controller/machineset.go
@@ -619,7 +619,7 @@ func getMachineKeys(machines []*v1alpha1.Machine) []string {
 	return machineKeys
 }
 
-func (c *controller) prepareMachineForDeletion(targetMachine *v1alpha1.Machine, machineSet *v1alpha1.MachineSet, wg *sync.WaitGroup, errCh *chan error) {
+func (c *controller) prepareMachineForDeletion(targetMachine *v1alpha1.Machine, machineSet *v1alpha1.MachineSet, wg *sync.WaitGroup, errCh chan<- error) {
 	defer wg.Done()
 
 	// Machine is already marked as 'to-be-deleted'
@@ -641,7 +641,7 @@ func (c *controller) prepareMachineForDeletion(targetMachine *v1alpha1.Machine, 
 		machineKey := MachineKey(targetMachine)
 		glog.V(2).Infof("Failed to delete %v, decrementing expectations for %v %s/%s", machineKey, machineSet.Kind, machineSet.Namespace, machineSet.Name)
 		c.expectations.DeletionObserved(machineSetKey, machineKey)
-		*errCh <- err
+		errCh <- err
 	}
 
 	// Force trigger deletion to reflect in machine status
@@ -669,9 +669,8 @@ func (c *controller) terminateMachines(inactiveMachines []*v1alpha1.Machine, mac
 	defer close(errCh)
 
 	wg.Add(numOfInactiveMachines)
-
 	for _, machine := range inactiveMachines {
-		go c.prepareMachineForDeletion(machine, machineSet, &wg, &errCh)
+		go c.prepareMachineForDeletion(machine, machineSet, &wg, errCh)
 	}
 	wg.Wait()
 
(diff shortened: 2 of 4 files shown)


pkg/controller/machineset.go
```diff
@@ -664,6 +666,8 @@ func (c *controller) terminateMachines(inactiveMachines []*v1alpha1.Machine, mac
 		numOfInactiveMachines = len(inactiveMachines)
 		errCh                 = make(chan error, numOfInactiveMachines)
 	)
+	defer close(errCh)
```
swapnilgm: Just from readability perspective, I would refactor it as:
```suggestion
	wg.Add(numOfInactiveMachines)
	for _, machine := range inactiveMachines {
		go c.prepareMachineForDeletion(machine, machineSet, &wg, &errCh)
	}
	wg.Wait()
	close(errCh)
	err, ok := <-errCh
	if !ok {
	 return nil
	}
	return err
	
```

pkg/controller/drain.go
```diff
@@ -414,6 +414,7 @@ func (o *DrainOptions) getGlobalTimeoutForPodsWithoutPV(pods []*api.Pod) time.Du
 
 func (o *DrainOptions) evictPods(attemptEvict bool, pods []api.Pod, policyGroupVersion string, getPodFn func(namespace, name string) (*api.Pod, error)) error {
 	returnCh := make(chan error, len(pods))
+	defer close(returnCh)
```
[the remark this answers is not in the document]
ggaurav10: Yes, the closing of this channel is not handled properly in the go routines that follow. These go routines don't check whether the caller has/has not closed the channel before writing to them. However, this [`for`](https://github.com/gardener/machine-controller-manager/pull/341/files/735d9736cbfcbc42b67d4a4bd69c70e36c9f14bd#diff-face99aebd3cd79577003f7df747c863R443-R449) loop makes sure that all go routines have already returned, after which the channel is closed. So we are safe.
This could be a potential issue if those go routines themselves have a bug.

pkg/controller/machineset.go
amshuman-kr: > ```go
> close(errCh)
> err, ok := <-errCh
> ```

Won't work :-)

swapnilgm: It will work, https://play.golang.org/p/Oz3XDGZpUM8
Readability is perspective based. Your choice.

amshuman-kr: 👍 You are right. My bad.

My comments were not based on readability. I do prefer the error check standard pattern over the switch check.

amshuman-kr: BTW regarding readability, I think we don't really need the call to `prepareMachineForDeletion` as goroutines in the first place. I think we are being over-aggressive with asynchronous calls. But that is for another day.

ggaurav10: Closing a channel before reading from it won't work:
https://play.golang.org/p/F1Z7z0Zs5a3
In @swapnilgm's go playground it works because the channel is read before the go routine could close it. It is also a potential race.

ggaurav10: Even in @swapnilgm's suggested changes, there could be a race. The channel could be closed before any of the go routines complete/scheduled. I prefer the `defer` way of closing the channel.

ggaurav10: Just saw: `wg.Wait()` would take care of the potential race. However, closing the channel before reading would not work.

swapnilgm: Closing of channel is developers reponsibility. Developer has to take care of race condition and all. Here, as you already noticed `wg.Wait()` takes care of it. 

> However, closing the channel before reading would not work.

The closing of channel before reading is totally fine by golng design. Its been used many places in golang based project. Nothing wrong in it. Only writing to channel is not allowed after closing the channel.

ggaurav10: what about this: https://play.golang.org/p/F1Z7z0Zs5a3
Doesn't seem to work here. Am I missing something?

swapnilgm: Ofcourse its race condition. Its your responsibility to handle it, but you missed it. If the go routine you had created won the race, then it would been paniced.
check : https://play.golang.org/p/2J7FOtazkXd . I simply added `sleep` so that go routing gets executed first post closing of channel.

pkg/controller/drain.go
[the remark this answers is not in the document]
prashanth26: So I shall keep it as it is for now.

pkg/controller/machineset.go
swapnilgm: Writing  to unbuffered channel  is blocking call. Try making it buffered channel : https://play.golang.org/p/2J7FOtazkXd

prashanth26: I am keeping the defer logic as it is. However, I am changing the receiving channel to write only as pointed out by Swapnil. Refer - https://github.com/gardener/machine-controller-manager/pull/341/commits/9c50392cdb691bfbafc51a6dcb8496256f64e4a6.

prashanth26:
CPU utilization before changes --> Exponential CPU increase observed

Screen Shot 2019-10-10 at 2 12 20 PM

CPU utilization after changes --> Constant CPU usage observed

Screen Shot 2019-10-10 at 2 13 48 PM

prashanth26:
@swapnilgm Are your concerns addressed?

PR #343, opened 2019-10-11
Added multi stage docker build

afritzler:
This PR adds a multi stage build `Dockerfile` to this project and removes the necessity to run a local build prior to a `make docker-images` run.

**Release note**:
```improvement developer
Added multi stage docker build
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,19 @@
-FROM alpine:3.10
+#############      builder                                  #############
+FROM golang:1.13.1 AS builder
+
+WORKDIR /go/src/github.com/gardener/machine-controller-manager
+COPY . .
+
+RUN .ci/build
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,7 +1,19 @@
...
+RUN .ci/build
+
+#############      base                                     #############
+FROM alpine:3.10 as base
 
-COPY bin/rel/machine-controller-manager /machine-controller-manager
+RUN apk add --update bash curl tzdata
 WORKDIR /
+
+#############      gardener-extension-hyper                 #############
```
ggaurav10: This should be "machine-controller-manager", right?

afritzler: Right :)

afritzler: Should be fixed now.

PR #347, opened 2019-10-30
Delete infra resources before returning error on create

### LINKED ISSUE #345: Azure improvements
## Issue

There were several VMs create instead of a few on Azure. Looked like an infra issue. However, in order to prevent such future scenarios, the following changes are to be made on MCM

- [x] Better error handling on the failure of VM creation. Delete any existing resource before returning error on creation.
- [ ] Check all code paths for reference objects being reused.

prashanth26:
Deletes any infra resources before returning error on machine create

Fixes #345 

**Release note**:

```improvement operator
Deletes any infra resources before returning error on machine create
```

### DIFF
--- a/pkg/driver/driver_azure.go
+++ b/pkg/driver/driver_azure.go
@@ -334,6 +334,7 @@ func (d *AzureDriver) createVMNicDisk() (*compute.VirtualMachine, error) {
 		vnetResourceGroup = resourceGroupName
 		subnetName        = d.AzureMachineClass.Spec.SubnetInfo.SubnetName
 		nicName           = dependencyNameFromVMName(vmName, nicSuffix)
+		diskName          = dependencyNameFromVMName(vmName, diskSuffix)
 	)
 
 	clients, err := d.setup()
@@ -346,6 +347,9 @@ func (d *AzureDriver) createVMNicDisk() (*compute.VirtualMachine, error) {
 		vnetResourceGroup = *d.AzureMachineClass.Spec.SubnetInfo.VnetResourceGroup
 	}
 
+	/*
+		Subnet fetching
+	*/
 	// Getting the subnet object for subnetName
 	subnet, err := clients.subnet.Get(
 		ctx,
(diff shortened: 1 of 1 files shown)


ggaurav10:
what should happen if MCM crashes/restarts while creating VM or other infra resources. Should we handle that case with this PR or separately.

amshuman-kr:
> what should happen if MCM crashes/restarts while creating VM or other infra resources. Should we handle that case with this PR or separately.

We should address it. But I can't think of a good way apart from keeping track of the individual sub-resources (nic, disk etc.) in persistence (CRD status).

prashanth26:
> what should happen if MCM crashes/restarts while creating VM or other infra resources. Should we handle that case with this PR or separately?

In case of Azure we use `CreateOrUpdate()` calls. Hence if MCM crashes during the creation, it will just update the previously created resources and create any new ones. But yes, in other infrastructures we should handle this. What I have in mind is to fix this in out of tree.

prashanth26:
> > what should happen if MCM crashes/restarts while creating VM or other infra resources. Should we handle that case with this PR or separately.
> 
> We should address it. But I can't think of a good way apart from keeping track of the individual sub-resources (nic, disk etc.) in persistence (CRD status).

What I had in mind, was tagging these resources with machine names and checking their existence before actually creating them.

pkg/driver/driver_azure.go
```diff
@@ -359,48 +363,90 @@ func (d *AzureDriver) createVMNicDisk() (*compute.VirtualMachine, error) {
...
+			glog.Errorf("Error occurred during resource clean up: %s", deleteErr)
+		}
+
 		return nil, onARMAPIErrorFail(prometheusServiceNIC, err, "NIC.CreateOrUpdate failed for %s", *NICParameters.Name)
 	}
+
+	// Wait until NIC is created
 	err = NICFuture.WaitForCompletionRef(ctx, clients.nic.Client)
 	if err != nil {
+		// Since machine creation failed, delete any infra resources created
```
hardikdr: Is this comment here appropriate or is it intended for NIC ?

prashanth26: Since this was a create machine method, i had mentioned machine creation failure. Since the method deletes any infra not just NIC right?

pkg/driver/driver_azure.go
```diff
@@ -359,48 +363,90 @@ func (d *AzureDriver) createVMNicDisk() (*compute.VirtualMachine, error) {
...
 	// Creating NICParameters for new NIC creation request
 	NICParameters := d.getNICParameters(vmName, &subnet)
+
+	// NIC creation request
 	NICFuture, err := clients.nic.CreateOrUpdate(ctx, resourceGroupName, *NICParameters.Name, NICParameters)
 	if err != nil {
+		// Since machine creation failed, delete any infra resources created
+		deleteErr := clients.deleteVMNicDisk(ctx, resourceGroupName, vmName, nicName, diskName)
+		if deleteErr != nil {
+			glog.Errorf("Error occurred during resource clean up: %s", deleteErr)
```
hardikdr: Should we return an error here?

Context:
While discussing offline with @prashanth26, we realized not returning the error could lead to an undesired situation of one extra machine created. Basically we try to delete all the resources of creation of any of them fail. If the deletion of the resources also fails, which is very likely, we may end up with orphan resources.
- We'd then expect Safety-controller to clean up, but personally I'd prefer to not rely on safety-net for features to work, it should only act while unknown behaviors from infra is observed.

_Also returning the error might not be the complete solution_, as deletion-flow would still delete the machine-object as the `ProviderID` field could be empty, this could again lead to orphan-resources.  

Overall we need to brainstorm a little for a better deletion logic for machines specifically for Azure, it can be tracked [here](https://github.com/gardener/machine-controller-manager/issues/350).

We don't have to convolute the purpose of this PR, we can solve it separately at the issue tagged [above](https://github.com/gardener/machine-controller-manager/issues/350).

PR #356, opened 2019-11-20
OpenStack: Ensure VMs are deleted after unsuccessful create operation

### LINKED ISSUE #355: OpenStack: compute instance with error state is not removed
This is a follow-up for the #286 PR.

It looks like the compute instance create error is not handled properly in https://github.com/gardener/machine-controller-manager/blob/9c45b1fe89758caa8fa8181996058c9c2bacee00/pkg/controller/machine.go#L342

Probably the error handling logic has to ensure that the VM is removed, if it is in error state just after the creation. Or, if this is applicable only to OpenStack, the VM remove logic has to be embedded inside the https://github.com/gardener/machine-controller-manager/blob/9c45b1fe89758caa8fa8181996058c9c2bacee00/pkg/driver/driver_openstack.go#L112 error handler, like I tried to implement in [docker machine](https://github.com/docker/machine/pull/4755/commits/b8bb147e14be53d078b29963345d846622ee9563)

/cc @afritzler  @hardikdr @prashanth26

kayrus:

New logic ensures that the machine resources after the failed create process are purged. If there are no VMs, the `Delete()` method will return nil.

In addition in OpenStack environment, the `Delete()` method will wait until the machine resource is actually deleted, before creating a new machine. This can avoid quota limit exceptions.

Fixes #355 

**Release note**:

```improvement operator
- improvement: OpenStack: Ensure VMs are deleted after unsuccessful create operation
```

### DIFF
--- a/pkg/controller/machine.go
+++ b/pkg/controller/machine.go
@@ -354,6 +354,13 @@ func (c *controller) machineCreate(machine *v1alpha1.Machine, driver driver.Driv
 			LastUpdateTime: metav1.Now(),
 		}
 		c.updateMachineStatus(machine, lastOperation, currentStatus)
+
+		// make sure that the failed machine doesn't exist
+		if e := driver.Delete(); e != nil {
+			message := fmt.Sprintf("Error deleting machine %s (%s) after unsuccessful create attempt: %s", machine.Name, e.Error(), err.Error())
+			return errors.New(message)
+		}
+
 		return err
 	}
 	glog.V(2).Infof("Created machine: %q, MachineID: %s", machine.Name, actualProviderID)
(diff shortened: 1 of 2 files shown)


pkg/controller/machine.go
```diff
@@ -354,6 +354,13 @@ func (c *controller) machineCreate(machine *v1alpha1.Machine, driver driver.Driv
 			LastUpdateTime: metav1.Now(),
 		}
 		c.updateMachineStatus(machine, lastOperation, currentStatus)
+
+		// make sure that the failed machine doesn't exist
+		if e := driver.Delete(); e != nil {
```
prashanth26: IMO I don't think we should be doing this here. I think this should be handled at the driver to trigger such deletion. Because all the MCM should care is if creation went through or not and if not error out. Creation logic at MCM shouldn't care about delete on failure. I think it can be better handled at the driver itself. Something like this - https://github.com/gardener/machine-controller-manager/blob/master/pkg/driver/driver_azure.go#L408-L413

kayrus: Fixed

PR #365, opened 2019-12-02
Bugfix: Remove dependency on MachineClass while deleting MachineDeployment/MachineSets

### LINKED ISSUE #357: Remove dependency on MachineClass while deleting MachineDeployment
## Issue
Currently, we block the deletion of the MachineDeployment if referred MachineClass is missing. I guess it makes sense to avoid/block enqueueing the MachineDeployment while creation if referred MachiineClass is missing, which we already do. But for any reason, if referred MachineClass is deleted, and it can be ensured that no Machines are lying around referring to it, probably it should be safe to delete such a MachineDeployment.

## Probable Solution
Allow deletion of MachineDeployment, if `Spec.Replicas=0`, no child objects [machine-objects, in case the controller crashed immediately after] exists. Open for suggestions.

/cc @prashanth26

prashanth26:

Fixes #357 

**Release note**:

```improvement operator
Bugfix: Remove dependency on MachineClass while deleting MachineDeployment/MachineSets
```

### DIFF
--- a/pkg/controller/deployment.go
+++ b/pkg/controller/deployment.go
@@ -464,10 +464,12 @@ func (dc *controller) reconcileClusterMachineDeployment(key string) error {
 		return nil
 	}
 
-	// Validate MachineClass
-	_, secretRef, err := dc.validateMachineClass(&deployment.Spec.Template.Spec.Class)
-	if err != nil || secretRef == nil {
-		return err
+	if deployment.DeletionTimestamp == nil {
+		// Validate MachineClass if the machineDeployment is not triggerred for deletion
+		_, secretRef, err := dc.validateMachineClass(&deployment.Spec.Template.Spec.Class)
+		if err != nil || secretRef == nil {
+			return err
+		}
 	}
 
 	// Resync the MachineDeployment after 10 minutes to avoid missing out on missed out events
--- a/pkg/controller/machineset.go
+++ b/pkg/controller/machineset.go
@@ -462,14 +462,14 @@ func (c *controller) reconcileClusterMachineSet(key string) error {
 		return nil
 	}
 
-	// Validate MachineClass
-	_, secretRef, err := c.validateMachineClass(&machineSet.Spec.Template.Spec.Class)
-	if err != nil || secretRef == nil {
-		return err
-	}
-
-	// Manipulate finalizers
 	if machineSet.DeletionTimestamp == nil {
+		// Validate MachineClass
+		_, secretRef, err := c.validateMachineClass(&machineSet.Spec.Template.Spec.Class)
+		if err != nil || secretRef == nil {
+			return err
+		}
+
+		// Manipulate finalizers
 		c.addMachineSetFinalizers(machineSet)
 	}
 


hardikdr:
> Also, if `DeletionTimestamp != nil` and there are machines (with VMs), shouldn't we block the deletion of `MachineDeployment`?

I think this is only about removing the check of MachineClass while deletion, else remaining behavior should remain the same, eg. machine-deployment should not be deleted if there are machine objects around yet referring to it.  @prashanth26 can confirm may be.

> What about same logic in machineset and machine

Not sure. Probably we could delete the machineset and machinedeployment if corresponding MachineClass is missing.  But we might not want to delete the machine-object if `Spec.ProviderID`  is set and MachineClass is missing. 
- In the case of a machine object, there is a possibility of actual VM existing, but MachineClass being deleted unintentionally.

prashanth26:
> What about same logic in [machineset](https://github.com/gardener/machine-controller-manager/blob/rel-0.21.0/pkg/controller/machineset.go#L466) and [machine](https://github.com/gardener/machine-controller-manager/blob/rel-0.21.0/pkg/controller/machine.go#L141)

I think machine set we could add, but for the machine object I was thinking with OOT we would require the machineClass to compute the providerID for AWS. Because we would be using machine object name and spec to determine the providerID. Hence it would better if we left it for machine objects.

prashanth26:
> > Also, if `DeletionTimestamp != nil` and there are machines (with VMs), shouldn't we block the deletion of `MachineDeployment`?
> 
> I think this is only about removing the check of MachineClass while deletion, else remaining behavior should remain the same, eg. machine-deployment should not be deleted if there are machine objects around yet referring to it. @prashanth26 can confirm may be.

Hardik is right, the existing logic already takes care of logic like these. Refer - https://github.com/gardener/machine-controller-manager/blob/d160f3043f0ce1a49953aa026610e7ee9fa522f0/pkg/controller/deployment.go#L513-L524. This is just an exclusion to allow deletion of machineDeployments without machineClasses.

prashanth26:
> What about same logic in [machineset](https://github.com/gardener/machine-controller-manager/blob/rel-0.21.0/pkg/controller/machineset.go#L466) and [machine](https://github.com/gardener/machine-controller-manager/blob/rel-0.21.0/pkg/controller/machine.go#L141)

I could add this as well if everyone feels the same.

ggaurav10:
Sounds reasonable. Without the finalisers, how do we make sure that when deleting a machine object, machine spec/class is available? If the machine class is deleted before the machine object,  what are the options?

prashanth26:
> Without the finalisers, how do we make sure that when deleting a machine object, machine spec/class is available?

Without the finalizers ideally, no operation would've been performed on this machineDeployment object. There would be no machineSets created for it.

> If the machine class is deleted before the machine object, what are the options?

IMO if machineClass isn't there for a machine,  we would have to manually re-create the machineClass for deletion to proceed. We could eliminate this by using only the providerID field on machine objects, but I think it might get messy with OOT (to use only machineNames), and even if we wanted to run checks in the future to check VM properties before deleting the machine.

prashanth26:
Hi All,

I have updated the logic to work for machine sets as well. Are there any pending changes?

PR #366, opened 2019-12-03
Add csi volume fetching for AliCloud

jia-jerry:
AliCloud Seed clusters uses CSI volume plugin. Need to consider it.
Fixes #

**Release note**:

```improvement operator
When nodes are rolling updated, pods with Persistent volume attached can now be evicted correctly/serially before the node is drained.
```

### DIFF
--- a/pkg/driver/driver_alicloud.go
+++ b/pkg/driver/driver_alicloud.go
@@ -292,12 +292,12 @@ func (c *AlicloudDriver) GetVolNames(specs []corev1.PersistentVolumeSpec) ([]str
 	names := []string{}
 	for i := range specs {
 		spec := &specs[i]
-		if spec.FlexVolume == nil || spec.FlexVolume.Options == nil {
-			// Not an aliCloud volume
-			continue
-		}
-		if name, ok := spec.FlexVolume.Options["volumeId"]; ok {
-			names = append(names, name)
+		if spec.FlexVolume != nil && spec.FlexVolume.Options != nil {
+			if name, ok := spec.FlexVolume.Options["volumeId"]; ok {
+				names = append(names, name)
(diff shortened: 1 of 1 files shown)


prashanth26:
CC: @ggaurav10

pkg/driver/driver_alicloud.go
```diff
@@ -292,12 +292,12 @@ func (c *AlicloudDriver) GetVolNames(specs []corev1.PersistentVolumeSpec) ([]str
...
-			// Not an aliCloud volume
-			continue
-		}
-		if name, ok := spec.FlexVolume.Options["volumeId"]; ok {
-			names = append(names, name)
+		if spec.FlexVolume != nil && spec.FlexVolume.Options != nil {
+			if name, ok := spec.FlexVolume.Options["volumeId"]; ok {
+				names = append(names, name)
+			}
+		} else if spec.CSI != nil && spec.CSI.VolumeHandle != "" {
```
ggaurav10: Should we also check if the field `spec.CSI.driver` matches the driver name(s) relevant to Alicloud?
https://kubernetes-csi.github.io/docs/drivers.html#production-drivers

Background: Typically other cloud providers such as AWS, Azure etc, have specific fields in the [`PersistentVolumeSource`](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L177) struct, those fields were used in the corresponding drivers (eg. [AWS driver](https://github.com/gardener/machine-controller-manager/blob/rel-0.21.0/pkg/driver/driver_aws.go#L318)).

jia-jerry: By now, only  *diskplugin.csi.alibabacloud.com* will be considered.

PR #373, opened 2019-12-13
Avoid force-deleting if previous drain has failed

hardikdr:
**What this PR does / why we need it**: This PR skips the force-deletion of the machine when the previous drain has failed.

Fixes #

**Release note**:

```improvement operator
Avoided force-deleting the machine if previous drain has failed.
```

### DIFF
--- a/pkg/controller/drain.go
+++ b/pkg/controller/drain.go
@@ -175,6 +175,7 @@ func (o *DrainOptions) RunDrain() error {
 	}()
 
 	if err := o.RunCordonOrUncordon(true); err != nil {
+		glog.Errorf("Drain Error: Cordoninig of node failed with error: %v", err)
 		return err
 	}
 
@@ -942,7 +943,7 @@ func (o *DrainOptions) RunCordonOrUncordon(desired bool) error {
 	}
 	unsched := node.Spec.Unschedulable
 	if unsched == desired {
-		glog.V(3).Info("Already desired")
+		glog.V(3).Infof("Already desired state of cordoning or uncordoning for node: %q", node.Name)
 	} else {
 		clone := node.DeepCopy()
 		clone.Spec.Unschedulable = desired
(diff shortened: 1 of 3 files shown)


pkg/controller/drain.go
```diff
@@ -942,7 +943,7 @@ func (o *DrainOptions) RunCordonOrUncordon(desired bool) error {
 	}
 	unsched := node.Spec.Unschedulable
 	if unsched == desired {
-		glog.V(3).Info("Already desired")
+		glog.V(3).Infof("Already desired state of cordoning or uncordoning for node: %q", node.Name)
```
prashanth26:
```suggestion
		glog.V(3).Infof("Scheduling state for node %q is already in desired state", node.Name)
```

pkg/controller/machine.go
```diff
@@ -503,7 +503,7 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
 			timeOut := metav1.Now().Add(-timeOutDuration).Sub(machine.Status.CurrentStatus.LastUpdateTime.Time)
 			timeOutOccurred = timeOut > 0
 
-			if forceDeleteLabelPresent || timeOutOccurred || lastDrainFailed {
+			if forceDeleteLabelPresent || timeOutOccurred {
 				// To perform forceful machine drain/delete either one of the below conditions must be satified
 				// 1. force-deletion: "True" label must be present
 				// 2. Deletion operation is more than drain-timeout minutes old
```
ggaurav10: The third point needs to be removed. The following comment in this `if` block also needs to be updated accordingly:
```
glog.V(2).Infof("Force deletion has been triggerred for machine %q due to Label:%t, timeout:%t, lastDrainFailed:%t",...
```

Also, we don't need the variable `lastDrainFailed` anymore. We can leave it for logging purposes though.

hardikdr: thanks, good catch, fixed.

PR #375, opened 2019-12-16
Activate healthz earlier

zuzzas:

Currently, due to failed leader election (in the case of multiple machine-controller-managers), the liveness probe will constantly restart a container, which is confusing at best and may trip a monitoring system. Also, it makes no sense to not enable liveness probe immediately. Liveness probe should start to fail once a program transitions into a state, in which it won't be able to continue operating correctly, and it needs to be shut down externally.

**Release note**:

```noteworthy operator
Healthz endpoint is immediately initialized with a healty (200) response
```


prashanth26:
Hi @zuzzas ,

We appreciate both the PRs. The changes look good to me. I am waiting for the integration tests to pass on our pipeline, post which I shall merge the PR. 

Regards,
Prashanth

PR #378, opened 2019-12-21
Support multiple Disks/Security Groups on AWS

rewiko:

Machine-controller-manager do not support multiple disks and security groups on the instance. Allowing multiple disks separate IOPS, disk space for docker, kubelet volumes and attaching multiple security groups.

Fixes #

**Release note**:

Support multiple disks on AWS and keep it backward compatible with a single disk without mention of `DeviceName`.
Support multiple security groups in AWS.


pkg/apis/machine/validation/awsmachineclass.go
```diff
@@ -116,18 +116,34 @@ func validateAWSClassSpecTags(tags map[string]string, fldPath *field.Path) field
...
+			allErrs = append(allErrs, field.Required(idxPath.Child("ebs.volumeType"), "Please mention a valid ebs volume type"))
 		} else if blockDevices[0].Ebs.VolumeType == "io1" && blockDevices[0].Ebs.Iops <= 0 {
-			allErrs = append(allErrs, field.Required(fldPath.Child("ebs.iops"), "Please mention a valid ebs volume iops"))
+			allErrs = append(allErrs, field.Required(idxPath.Child("ebs.iops"), "Please mention a valid ebs volume iops"))
 		}
 	}
+
+	if rootPartitionCount > 1 {
+		allErrs = append(allErrs, field.Required(fldPath, "Only one '/root' partition can be specified"))
+	} else if rootPartitionCount == 0 && len(blockDevices) > 1 {
```
rewiko: `len(blockDevices) > 1` allow backward compatibility when a single disk is provided without `DeviceName`

hardikdr:
@rewiko thanks for the PR, will take a look soon.

pkg/apis/machine/validation/awsmachineclass.go
```diff
@@ -116,18 +117,49 @@ func validateAWSClassSpecTags(tags map[string]string, fldPath *field.Path) field
 
 func validateBlockDevices(blockDevices []machine.AWSBlockDeviceMappingSpec, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
-	if len(blockDevices) > 1 {
-		allErrs = append(allErrs, field.Required(fldPath.Child(""), "Can only specify one (root) block device"))
-	} else if len(blockDevices) == 1 {
-		if blockDevices[0].Ebs.VolumeSize <= 0 {
-			allErrs = append(allErrs, field.Required(fldPath.Child("ebs.volumeSize"), "Please mention a valid ebs volume size"))
+
+	if 0 == len(blockDevices) {
```
ggaurav10: Do we need this new check? Do we know what is the default behaviour in case no block devices are mentioned?

pkg/apis/machine/validation/awsmachineclass.go
ggaurav10: Should this be added as comment in the code also

pkg/apis/machine/validation/awsmachineclass.go
rewiko: I've tested, if you don't provide any block device, AWS will create one for you and mount as root partition. 
I will remove the condition.

rewiko:
I've addressed your comment and added tests on both files.
I've also added support to attach multiples security groups.

pkg/driver/driver_aws.go
```diff
@@ -142,6 +117,77 @@ func (d *AWSDriver) Create() (string, string, error) {
...
+	for _, disk := range blockDevices {
+
+		deviceName := disk.DeviceName
+		if disk.DeviceName == "/root" || len(blockDevices) == 1 {
+			deviceName = *rootDeviceName
+		}
+		volumeSize := disk.Ebs.VolumeSize
+		volumeType := disk.Ebs.VolumeType
+		deleteOnTermination := disk.Ebs.DeleteOnTermination
+		iops := disk.Ebs.Iops
```
ggaurav10: This variable is not used anywhere. Shouldn't the build fail here?

rewiko: Yes just pushed the fix :)

PR #384, opened 2020-01-10
added MachineClass for incremental out-of-tree migration

MartinWeindel:
This PR adds the new MachineClass type for the out-of-tree (OOT) machine controller manager providers.
The infrastructure providers are still located together in the [gardener-extensions](https://github.com/gardener/gardener-extensions) repository.
If the migration to OOT should happen separately for each infra-provider, some of the current specific machineclass types are referenced alongside the new machineclass type in the worker controllers.
The provider specific machineclass types are already removed in the mcm branch `cmi-client` and can therefore not be used as dependency for the gardener-extensions project. 
This is only an immediate solution until either all infra-providers are migrated to OOT or the gardnerer-extensions repository is splitted up.


hardikdr:
Hi Martin, thanks for the PR.  :) 
The necessary changes required to adapt the master branch with out-of-tree is being worked upon at the [cmi-client](https://github.com/gardener/machine-controller-manager/tree/cmi-client) branch.

MartinWeindel:
@hardikdr Hardik, I'm aware that out-of-tree is being worked on in the cmi-branch. But this does not solve the problem I described with migrating individual infrastructure providers in the gardener-extensions repository. Either you have to include the old machine classes in the cmi-branch or with a solution like that I purpose here.

hardikdr:
I see, sorry I misunderstood it earlier. We have issues with CI at the moment, we shall fix it soon and merge the PR soon, thanks.

PR #388, opened 2020-01-14
Added Cinder based Custom Root Disks for OpenStack Vms

### LINKED ISSUE #386: Custom Root Disk Size for OpenStack Machines
Currently the OpenStack flavor determines the root disk size of machine. By using a custom block device mapping it is possible to boot machines based of Cinder backed volumes.

afritzler:
This PR extends the OpenStack MachineClass to support custom root disk size. It does this by adding a Cinder based root disk volume to the VM created via MCM.

To configure the root disk size the manifest looks like the following

```shell
apiVersion: machine.sapcloud.io/v1alpha1
kind: OpenStackMachineClass
metadata:
  name: test-openstack # Name of OpenStack machine class goes here
  namespace: default # Namespace in which the machine class is to be deployed
spec:
  ...
  rootDiskSize: 100 # 100 GB 
```

The creation policy for the Cinder volume is `DeleteOnTermination` meaning that the volume gets removed when the machine is terminated.

Fixes #386 

**Release note**:
```improvement operator
Added Cinder based root disk support with customisable disk size 
```

### DIFF
--- a/pkg/apis/machine/v1alpha1/types.go
+++ b/pkg/apis/machine/v1alpha1/types.go
@@ -742,6 +742,7 @@ type OpenStackMachineClassSpec struct {
 	Networks         []OpenStackNetwork      `json:"networks,omitempty"`
 	SecretRef        *corev1.SecretReference `json:"secretRef,omitempty"`
 	PodNetworkCidr   string                  `json:"podNetworkCidr"`
+	RootDiskSize     int                     `json:"rootDiskSize,omitempty"` // in GB
 }
 
 type OpenStackNetwork struct {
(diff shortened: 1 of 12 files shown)


pkg/apis/machine/v1alpha1/types.go
```diff
@@ -742,6 +742,7 @@ type OpenStackMachineClassSpec struct {
 	Networks         []OpenStackNetwork      `json:"networks,omitempty"`
 	SecretRef        *corev1.SecretReference `json:"secretRef,omitempty"`
 	PodNetworkCidr   string                  `json:"podNetworkCidr"`
+	RootDiskSize     int                     `json:"rootDiskSize,omitempty"` // in GB
```
ialidzhikov: Why `resource.Quantity` is not used?
g/g uses `resource.Quantity` for `volume.size` in the Shoot spec - ref https://github.com/gardener/gardener/blob/ab4ddfc16c0709b0d301ee8713f61fad7fcbeb6f/pkg/apis/core/v1beta1/types_cloudprofile.go#L136-L137

PR #395, opened 2020-01-23
Azure: Allow to manage machine with identity

dkistner:
Allow to create and manage Azure virtual machines which have an user assigned managed identity assigned to.

Azure machines can now be deployed with an attached user assigned managed identity, which can be used to authenticate to other Azure services like ACR.

Fixes #

**Release note**:

```improvement user
Azure machines can now be deployed with an attached user assigned managed identity.
```


pkg/driver/driver_azure.go
```diff
@@ -163,6 +163,15 @@ func (d *AzureDriver) getVMParameters(vmName string, networkInterfaceReferenceID
 		}
 	}
 
+	if d.AzureMachineClass.Spec.Properties.IdentityID != nil {
```
ggaurav10: ` && *d.AzureMachineClass.Spec.Properties.IdentityID != ""`?

dkistner: Thanks I have adapted this.

dkistner:
Thanks @ggaurav10. Should be done.

pkg/openapi/openapi_generated.go
```diff
@@ -9834,7 +9798,7 @@ func GetOpenAPIDefinitions(ref common.ReferenceCallback) map[string]common.OpenA
 					Properties: map[string]spec.Schema{
 						"replicas": {
 							SchemaProps: spec.SchemaProps{
-								Description: "Replicas is the number of desired replicas. This is a pointer to distinguish between explicit zero and unspecified. Defaults to 0. More info: https://kubernetes.io/docs/concepts/workloads/controllers/replicationcontroller#what-is-a-replicationcontroller",
+								Description: "Replicas is the number of desired replicas. This is a pointer to distinguish between explicit zero and unspecified. Defaults to 1. More info: https://kubernetes.io/docs/concepts/workloads/controllers/replicationcontroller#what-is-a-replicationcontroller",
```
ggaurav10: This shouldn't change with this PR. Is it a rebasing issue?

dkistner:
Hmm I just called `hack/generate-code` and this change was produced by the script. I have build this based on the master. Is there something which needs to be fixed on the upstream master? If yes give me a hint when I need to rebase and re run the code generation.

hardikdr:
Sure @dkistner, We haven't yet been able to look into it. We shall soon fix it and let you know.

hardikdr:
One of the recent PR dint run `./hack/generatecode`, which brought these changes into this PR.
We can merge this one.

pkg/openapi/openapi_generated.go
hardikdr: This got in as `hack/generatecode` was not run previously. Though we need to check on the open-api generate-code path, as in reality it gets defaulted to `0`, but not the concern of this PR.

PR #397, opened 2020-01-29
Added support for Data Disks in Azure and AliCloud

### LINKED ISSUE #396: Data Disks in AliCloud and Azure
Currently, MCM supports multiple (non-root) devices in AWS and GCP. Other IaaS providers (eg. Azure, AliCloud) also support this ability using Data Disks. However, MCM does not expose this option in relevant machine classes and drivers. 

Data Disks can be useful in k8s nodes for several scenarios:
* OS Disk Encryption has limited support in some IaaS providers, whereas data disk encryption is widely supported. Sensitive ephemeral data can be conveniently encrypted on data disks.  https://github.com/gardener/gardener/issues/1743
* Assurance of IOPS for workloads that utilize a dedicated data disk
* memory expansion

MCM should expose this ability.

guydc:

Enable MCM support for Data Disks in AliCloud and Azure. Aligns MCM’s capability to create and attach multiple devices across all major IaaS providers, for use cases such as memory expansion, resource isolation and encryption.  

Use a test suite for all driver tests. 

Fixes #396 
Part of gardener/gardener#1743

**Release note**:

```improvement developer
Added support for multiple Data Disks in Azure and AliCloud. 
```


hardikdr:
thanks @guydaichs for the PR, we shall review it soon.

pkg/apis/machine/validation/awsmachineclass.go
```diff
@@ -132,12 +132,19 @@ func validateBlockDevices(blockDevices []machine.AWSBlockDeviceMappingSpec, fldP
 
 	validVolumeTypes := []string{"gp2", "io1", "st1", "sc1", "standard"}
 
+	// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/device_naming.html
+	const dataDeviceNameFmt string = `^/dev/(sd[a-z]|xvd[a-c][a-z]?)$`
+	var dataDeviceNameRegexp = regexp.MustCompile("^" + dataDeviceNameFmt + "$")
```
ggaurav10: Do `^` and `$` still need to be there if `dataDeviceNameFmt` already has them?

pkg/driver/driver_alicloud.go
```diff
@@ -75,6 +76,11 @@ func (c *AlicloudDriver) Create() (string, string, error) {
 		request.InternetMaxBandwidthOut = requests.NewInteger(*c.AlicloudMachineClass.Spec.InternetMaxBandwidthOut)
 	}
 
+	if ok := c.AlicloudMachineClass.Spec.DataDisks != nil; ok {
```
ggaurav10: Should this line be replaced with just:
`c.AlicloudMachineClass.Spec.DataDisks != nil`
Also, since we are doing a `nil` check, do we need the following check also:
`len(c.AlicloudMachineClass.Spec.DataDisks) > 0`

pkg/driver/driver_azure.go
```diff
@@ -99,6 +99,9 @@ func (d *AzureDriver) getVMParameters(vmName string, networkInterfaceReferenceID
 		tagList[idx] = to.StringPtr(element)
 	}
 
+	azureDataDisks := d.AzureMachineClass.Spec.Properties.StorageProfile.DataDisks
+	dataDisks := d.generateDataDisks(vmName, azureDataDisks)
```
ggaurav10: Do we need a `nil` and `len()` check here also just like in the case of Alicloud? Please check previous comment.

hardikdr:
Thanks @guydaichs for PR/feature. As this does not fall under our existing CI scope[multiple-disks], we shall test it out separately soon, and update the thread here. 

@jia-jerry @dkistner Can you please take a brief look for the Alicloud and Azure part respectively :-)

hardikdr:
cc @rewiko  
As you had enabled support for multiple-disks recently in other PR.

prashanth26:
Looks like there is an issue here - 
pkg/apis/machine/validation/azuremachineclass.go:144: Sprintf format %s has arg lun of wrong type int32

dkistner:
I will try to check the azure specific part soon.

guydaichs:
@rewiko  - I added a verification for block device names in `awsmachineclass`, based on the recommended naming documented by [AWS](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/device_naming.html). I changed a few unit tests that were using a device named `/dev/vdga`. Is this ok from your point of view?

pkg/apis/machine/validation/awsmachineclass.go
```diff
@@ -132,12 +132,19 @@ func validateBlockDevices(blockDevices []machine.AWSBlockDeviceMappingSpec, fldP
...
+
 	// if blockDevices is empty, AWS will automatically create a root partition
 	for i, disk := range blockDevices {
 		idxPath := fldPath.Index(i)
 
 		if disk.DeviceName == "/root" {
 			rootPartitionCount++
+		} else if len(blockDevices) > 1 && !dataDeviceNameRegexp.MatchString(disk.DeviceName) {
+			// if there are multiple devices, non-root devices are expected to adhere to AWS naming conventions
+			allErrs = append(allErrs, field.Invalid(idxPath.Child("deviceName"), disk.DeviceName, utilvalidation.RegexError(dataDeviceNameFmt, "/dev/sdf")))
```
rewiko:
```suggestion
			allErrs = append(allErrs, field.Invalid(idxPath.Child("deviceName"), disk.DeviceName, utilvalidation.RegexError(fmt.Sprintf("Device name given: %s does not match the expected pattern", disk.DeviceName), dataDeviceNameFmt)))
```

pkg/apis/machine/validation/awsmachineclass_test.go
```diff
@@ -378,7 +371,7 @@ var _ = Describe("AWSMachineClass Validation", func() {
 					},
 				},
 				{
```
rewiko: I think tests should be added for the device name regex validation.

rewiko:
LGTM just a few comments

jia-jerry (changes requested):

pkg/apis/machine/validation/alicloudmachineclass.go
```diff
@@ -63,6 +64,34 @@ func validateAlicloudMachineClassSpec(spec *machine.AlicloudMachineClassSpec, fl
...
+			if _, keyExist := names[dataDisk.Name]; keyExist {
+				names[dataDisk.Name]++
+			} else {
+				names[dataDisk.Name] = 1
+			}
+
+			if dataDisk.Name == "" {
+				allErrs = append(allErrs, field.Required(idxPath.Child("name"), "Data Disk name is required"))
+			}
+			if dataDisk.Size <= 0 {
```
jia-jerry: For AliCloud, data disk size should be larger than 20 and less than 32768 GB.

pkg/driver/driver_azure.go
```diff
@@ -167,14 +171,35 @@ func (d *AzureDriver) getVMParameters(vmName string, networkInterfaceReferenceID
...
 }
 
+func (d *AzureDriver) generateDataDisks(vmName string, azureDataDisks []v1alpha1.AzureDataDisk) []compute.DataDisk {
+	var dataDisks []compute.DataDisk
+	for _, azureDataDisk := range azureDataDisks {
+		dataDiskName := dependencyNameFromVMNameAndDependency(getAzureDataDiskName(azureDataDisk), vmName, dataDiskSuffix)
+		dataDiskLun := azureDataDisk.Lun
+		dataDiskSize := azureDataDisk.DiskSizeGB
+		dataDisk := compute.DataDisk{
+			Lun:     &dataDiskLun,
```
dkistner: Can we simply enumerate the `lun` starting with `0`? This will be only called when a new machine is created. In all other cases we would need to determine a free `lun` first because k8s could already attach a data disk which allocate the `lun`.

pkg/apis/machine/v1alpha1/types.go
```diff
@@ -992,6 +993,15 @@ type AzureOSDisk struct {
 	CreateOption string                     `json:"createOption,omitempty"`
 }
 
+type AzureDataDisk struct {
+	Name         string                     `json:"name,omitempty"`
+	Lun          int32                      `json:"lun,omitempty"`
+	Caching      string                     `json:"caching,omitempty"`
+	ManagedDisk  AzureManagedDiskParameters `json:"managedDisk,omitempty"`
```
dkistner: Can we make the `ManagedDisk` attribute flat? We want to specify only the volume type here? So why not simply renaming it `volumeType` and make it of type string or better `compute.StorageAccountTypes`.

pkg/driver/driver_azure.go
```diff
@@ -167,14 +171,35 @@ func (d *AzureDriver) getVMParameters(vmName string, networkInterfaceReferenceID
...
+			Name:    &dataDiskName,
+			Caching: compute.CachingTypes(azureDataDisk.Caching),
+			ManagedDisk: &compute.ManagedDiskParameters{
+				StorageAccountType: compute.StorageAccountTypes(azureDataDisk.ManagedDisk.StorageAccountType),
+			},
+			DiskSizeGB:   &dataDiskSize,
+			CreateOption: compute.DiskCreateOptionTypes(azureDataDisk.CreateOption),
```
dkistner: Do we need other CreateOptions than `Empty`? I think `Attach` and `FromImage` would require addional information like which existing volume should be attached or which image should be used to create the volume?

pkg/driver/driver_azure.go
```diff
@@ -167,14 +171,35 @@ func (d *AzureDriver) getVMParameters(vmName string, networkInterfaceReferenceID
 		VMParameters.Identity = &compute.VirtualMachineIdentity{
 			Type: compute.ResourceIdentityTypeUserAssigned,
 			UserAssignedIdentities: map[string]*compute.VirtualMachineIdentityUserAssignedIdentitiesValue{
-				*d.AzureMachineClass.Spec.Properties.IdentityID: &compute.VirtualMachineIdentityUserAssignedIdentitiesValue{},
```
dkistner: Have you tested this?

pkg/apis/machine/v1alpha1/types.go
```diff
@@ -992,6 +993,15 @@ type AzureOSDisk struct {
 	CreateOption string                     `json:"createOption,omitempty"`
 }
 
+type AzureDataDisk struct {
+	Name         string                     `json:"name,omitempty"`
+	Lun          int32                      `json:"lun,omitempty"`
+	Caching      string                     `json:"caching,omitempty"`
```
dkistner: Need to check for chaching, but afaik it is only supported for premium storage. 
BTW: The same for os disk. Need to double check.

dkistner (changes requested):

pkg/apis/machine/v1alpha1/types.go
guydaichs: I couldn't find a limitation to premium storage in the [docs](https://docs.microsoft.com/en-us/learn/modules/caching-and-performance-azure-storage-and-disks/3-enable-and-configure-azure-vm-disk-cache-by-using-the-azure-portal), and I was able to configure caching for standard disks. 
image

pkg/driver/driver_azure.go
guydaichs: unintended, i'll revert this.

pkg/driver/driver_azure.go
guydaichs: well, it depends on the use case. If MCM is used purely for declarative VM management (outside of k8s), it might  be useful to configure luns explicitly. For example, the recommended way for [identifying devices within the VM is by lun](https://docs.microsoft.com/en-us/azure/virtual-machines/troubleshooting/troubleshoot-device-names-problems#identify-disk-luns). This is also consistent with the device-naming behavior in the AWS machine controller.

pkg/apis/machine/validation/alicloudmachineclass.go
```diff
@@ -63,6 +64,34 @@ func validateAlicloudMachineClassSpec(spec *machine.AlicloudMachineClassSpec, fl
...
+		names := map[string]int{}
+		for i, dataDisk := range spec.DataDisks {
+			idxPath := fldPath.Child("dataDisks").Index(i)
+			if _, keyExist := names[dataDisk.Name]; keyExist {
+				names[dataDisk.Name]++
+			} else {
+				names[dataDisk.Name] = 1
+			}
+
+			if dataDisk.Name == "" {
```
jia-jerry: Also need to check Name string pattern. ```The name must be 2 to 128 characters in length, and can contain letters, digits, periods (.), underscores (_), colons (:), and hyphens (-). It must start with a letter.``` Should be better checked when the shoot is created, btw.

jia-jerry (changes requested):

pkg/driver/driver_azure.go
dkistner: Hmm ok, you mean machines which are managed as Kubernetes resources but not intended to be k8s nodes. Makes sense for me. @prashanth26, @hardikdr WDYT?

Could we then make the `lun` optional? In case no `lun` is set then enumerate and starting with 0? WDYT?

pkg/apis/machine/v1alpha1/types.go
dkistner: I have also double checked this. Caching is possible also standard storage. Thanks!

dkistner: Could we make caching optional and default to `None`?

pkg/driver/driver_azure.go
guydaichs: Well, this input is used both for disk creation, deletion and on-error cleanup flows. so, lun ids will need to be generated in several scenarios. If it was possible to default this value before storage, handling unset luns would be much cleaner. To avoid over complication, I suggest that luns would either be exclusively set or unset.  WDYT ?

pkg/apis/machine/validation/alicloudmachineclass.go
guydaichs: The actual name is generated from the device name, instance name and some suffix. The machine name is safe, but I can't assume much about its length. To be a bit safer, I can enforce that data disk name length should be <64. Regarding the shoot validation - I can add it in the ac-provider extension. WDYT?

jia-jerry: Sure. Thank you.

hardikdr (commented):
Thanks for the nice PR @guydaichs

guydaichs:
> How are the luns managed by the PVC/PV controller in k8s ? Is it possible that, kubernetes allocates the same lun value for a pvc-disk and we set the same lun value for the data-disk here, what will be the behaviour then 

@hardikdr - It looks like the attachment process in azure finds an available lun : https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/vendor/k8s.io/legacy-cloud-providers/azure/azure_controller_common.go#L280. 

My assumptions are:
* MCM will provision and attach data-disks before any PV is attached. 
* If an existing PV is (re-)attached to a node, an available lun would be selected on-demand. 
* If the data-disk definition for an azure machine will change, the node would be destroyed and re-provisioned. 
 
Do you think that we need an additional safety mechanism?

pkg/driver/driver_azure.go
hardikdr: I think I am fine exclusively setting/unsetting it for the scope of this PR.
Though, we might want to in the future allow 
1. optional lun, if not specified start from 0, as pointed by @dkistner .
2. And check on the free lun before allocating, in case we decide to move towards managing machines not necessarily being k8s nodes.

dkistner: > How are the luns managed by the PVC/PV controller in k8s ?

The azure k8s driver is fetching the machine spec from Azure to check for free `luns` before attaching a volume. Therefore it shouldn't be possible that k8s allocate the same `lun` for data disk which is attached to machine by mcm.

My suggestion was to make the `lun` in the spec optional. If set ok then use that one on the machine (if invalid machine creation will of course fail). If not set start to enumerate the lun beginning with 0.

hardikdr (commented):
Overall looks good to me,  I have few minor comments inline, thanks for the much-needed feature.

pkg/driver/driver_azure.go
hardikdr: >My suggestion was to make the lun in the spec optional. If set ok then use that one on the machine (if invalid machine creation will of course fail). If not set start to enumerate the lun beginning with 0.

Understood thanks. Having optional lun would be indeed nice. 
Though, the current state doesn't seem to be hurting, leaving the decision to @guydaichs to change the behavior.

hardikdr:
>Do you think that we need an additional safety mechanism?

Looks good to me, but you might want to take a call on the suggestion made [here](https://github.com/gardener/machine-controller-manager/pull/397#discussion_r402962049) by @dkistner , if you feel to make the `lun` optional, and enumerate from zero if not specified.

hardikdr:
Overall, thanks a lot to @guydaichs for the PR, and thanks to everyone for reviewing it.

I'd suggest following, to further complete the review and go ahead merging it.

@guydaichs  Can you please address the remaining the review-comments on the PR. Also, `Resolve Conversations` wherever you feel it's taken care of already.  There is also a conflicting file, you might want to consider.
  - Overall, an apology from my side for the delayed review.

@dkistner  @jia-jerry  Can you please follow-up (resolve the conversations if taken care of or concern is not valid anymore) and approve the PR from the Azure's and Alicloud's perspective respectively when it's ready.  Thanks a lot in advance.

pkg/apis/machine/v1alpha1/types.go
```diff
@@ -992,6 +993,14 @@ type AzureOSDisk struct {
 	CreateOption string                     `json:"createOption,omitempty"`
 }
 
+type AzureDataDisk struct {
```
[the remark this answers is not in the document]
hardikdr: Let's drop this suggestion for now, not very critical from the functionality point of view. We shall try to align the APIs with that of SDK later separately if required.

pkg/apis/machine/validation/awsmachineclass_test.go
guydaichs: Yes, naming conventions also depend on virtualization types, and are documented here: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/device_naming.html. 

As discussed, for the scope of this PR, I think that it's sufficient to validate that the device name matches the `/dev/[sd/xvd]*` regex, and that there is no duplication of device names. In case of device name incompatibility with the VM type, I think that it's reasonable to fail on an AWS SDK error. 

The AWS EBS CSI Driver uses the `/dev/xvd` prefix : https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/cloud/devicemanager/manager.go#L29, and generates a suffix taking into consideration existing device names: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/cloud/devicemanager/allocator.go#L52. I expect that the aws provider extensions for gardener would generate device names using a logic similar to the EBS CSI Driver.

pkg/driver/driver_azure.go
guydaichs: Considering that the Azure Disk Deletion API is based on resource names https://docs.microsoft.com/en-us/rest/api/compute/disks/delete, it's important to establish unique and stable names for data disks. One option is to use the already-unique lun attribute in the naming convention, to achieve this uniqueness. If there was an option to mutate the machine class and set a value for unspecified luns once, that would fine. But we don't have this ability in MCM yet, and I'm concerned that generating this unique identifier in several flows can lead to inconsistency. For the scope of this PR, I prefer to keep the MCM simple and  require this value to be set.

pkg/driver/driver_alicloud.go
```diff
@@ -75,6 +76,11 @@ func (c *AlicloudDriver) Create() (string, string, error) {
 		request.InternetMaxBandwidthOut = requests.NewInteger(*c.AlicloudMachineClass.Spec.InternetMaxBandwidthOut)
 	}
 
+	if c.AlicloudMachineClass.Spec.DataDisks != nil && len(c.AlicloudMachineClass.Spec.DataDisks) > 0 {
+		dataDiskRequests := c.generateDataDiskRequests(c.AlicloudMachineClass.Spec.DataDisks)
```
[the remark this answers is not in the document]
guydaichs: Yes, both disks are tagged with the instance tags. Below is an example of an instance created with a single data disk: 
```json
{
  "TotalCount": 2,
  "Disks": {
    "Disk": [
      {
        "Size": 100,
        "Encrypted": true,
        "Tags": {
          "Tag": [
            {
              "TagKey": "kubernetes.io/cluster/shoot--dev--ac-cb3",
              "TagValue": "1"
            },
            {
              "TagKey": "kubernetes.io/rol e/worker/shoot--dev--ac-cb3",
              "TagValue": "1"
            }
          ]
        },
        "DeleteWithInstance": true,
        "DiskName": "shoot--dev--ac-cb3-cpu-worker-eu-central-1a-7c6477bdb7-hjw8l-dv1-data-disk",
        "DiskId": "d-gw8foxh3f9cp0ed0cee4"
      },
      {
        "Size": 75,
        "Encrypted": false,
        "Tags": {
          "Tag": [
            {
              "TagKey": "kubernetes.io/cluster/shoot--dev--ac-cb3",
              "TagValue": "1"
            },
            {
              "TagKey": "kubernetes.io/role/worker/shoot--dev--ac-cb3",
              "TagValue": "1"
            }
          ]
        },
        "DeleteWithInstance": true,
        "DiskName": "",
        "DiskId": "d-gw8c4p6uken0bp2nol18"
      }
    ]
  }
}
```

pkg/driver/driver_alicloud.go
```diff
@@ -106,6 +112,33 @@ func (c *AlicloudDriver) Create() (string, string, error) {
...
+		}
+
+		if disk.DeleteWithInstance != nil {
+			dataDiskRequest.DeleteWithInstance = strconv.FormatBool(*disk.DeleteWithInstance)
+		} else {
+			dataDiskRequest.DeleteWithInstance = strconv.FormatBool(true)
+		}
+
+		if disk.Category == "DiskEphemeralSSD" {
+			dataDiskRequest.DeleteWithInstance = ""
```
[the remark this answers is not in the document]
guydaichs: Ephemeral SSDs are using the physical host's local storage: https://www.alibabacloud.com/help/doc-detail/35241.htm. Their termination-with-instance is actually ignored, as seen in the terraform provider implementation: https://github.com/terraform-providers/terraform-provider-alicloud/blob/master/alicloud/resource_alicloud_instance.go#L990. 

Anyway, It's not really a use case in the context of gardener right now, as only mainstream volumes (cloud_efficiency, cloud_ssd) are viable options of worker volumes.

kubernetes/machine_classes/aws-machine-class.yaml
```diff
@@ -25,6 +25,15 @@ spec:
     namespace: default  # Namespace
     name: test-secret # Name of the secret
   blockDevices:
-    - ebs:
+    - name: /root
```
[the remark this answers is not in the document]
guydaichs: I'm only extending the example based on multiple-volume support introduced by @rewiko in a previous change. Yes, the change was backwards-compatible. If a single volume is provided it would be treated as a the root volume and the name would be obtained from AMI (machine-class provided name ignored).

pkg/apis/machine/validation/awsmachineclass_test.go
hardikdr: Sure, sounds good to me.

pkg/driver/driver_azure.go
hardikdr: Overall, sounds good to me. MCM could potentially lose the allocated-lun and storing it at MachineClass dynamically, could actually bring complexity, and probably not in the scope of this PR. 
@dkistner closing the comment, please feel free to re-open if you think otherwise :)

kubernetes/machine_classes/azure-machine-class.yaml
```diff
@@ -29,6 +29,21 @@ spec:
         caching: "None" # Caching Strategy (None/ReadOnly/ReadWrite)
         diskSizeGB: 50 # Size of disk to be created in GB
         createOption: "FromImage" # Create option for disk (Empty/Attach/FromImage)
+      DataDisks:
```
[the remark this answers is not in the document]
hardikdr: Resolved.

pkg/apis/machine/v1alpha1/types.go
hardikdr: @guydaichs , this has been taken care of, right ?

pkg/driver/driver_azure.go
hardikdr: I believe, this is also resolved, can you please confirm @guydaich ?

hardikdr: I believe, this is also resolved, can you please confirm @guydaich @dkistner ?

guydaichs: resolved

pkg/apis/machine/validation/alicloudmachineclass.go
guydaichs: length and regex validation on the device name was added.

pkg/apis/machine/validation/alicloudmachineclass.go
```diff
@@ -63,6 +64,34 @@ func validateAlicloudMachineClassSpec(spec *machine.AlicloudMachineClassSpec, fl
 		allErrs = append(allErrs, field.Required(fldPath.Child("keyPairName"), "KeyPairName is required"))
 	}
 
+	if spec.DataDisks != nil {
+		names := map[string]int{}
+		for i, dataDisk := range spec.DataDisks {
+			idxPath := fldPath.Child("dataDisks").Index(i)
+			if _, keyExist := names[dataDisk.Name]; keyExist {
```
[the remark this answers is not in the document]
guydaichs: validations were added

pkg/apis/machine/v1alpha1/types.go
guydaichs: done

pkg/driver/driver_azure.go
guydaichs: done

pkg/driver/driver_azure.go
guydaichs: confirmed

pkg/apis/machine/validation/alicloudmachineclass.go
guydaichs: done

pkg/apis/machine/types.go
```diff
@@ -1106,6 +1117,16 @@ type AlicloudSystemDisk struct {
...
 }
 
+// AlicloudDataDisk describes DataDisk for Alicloud.
+type AlicloudDataDisk struct {
+	Name               string
+	Category           string
+	Description        string
+	Encrypted          bool
+	Size               int
+	DeleteWithInstance bool
```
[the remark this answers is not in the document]
guydaichs: done

kubernetes/machine_classes/aws-machine-class.yaml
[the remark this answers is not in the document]
hardikdr: I see, thanks.

pkg/driver/driver_alicloud.go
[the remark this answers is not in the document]
hardikdr: thanks for checking.

pkg/driver/driver_alicloud.go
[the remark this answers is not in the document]
hardikdr: I see I would then assume it's deleted with the instance termination. Hope that's the correct understanding.

pkg/apis/machine/validation/awsmachineclass.go
guydaichs: fixed

pkg/driver/driver_alicloud.go
guydaichs: both checks added

pkg/driver/driver_azure.go
guydaichs: done.

pkg/apis/machine/validation/alicloudmachineclass.go
```diff
@@ -63,6 +66,43 @@ func validateAlicloudMachineClassSpec(spec *machine.AlicloudMachineClassSpec, fl
 		allErrs = append(allErrs, field.Required(fldPath.Child("keyPairName"), "KeyPairName is required"))
 	}
 
+	const dataDiskNameFmt string = `[a-zA-Z0-9\.\-_:]+`
```
jia-jerry, line 69: it should starts with a letter. The regexp looks like `[a-zA-Z][a-zA-Z0-9\.\-_:]+`

jia-jerry (changes requested):
one minor issue

pkg/driver/driver_azure.go
dkistner: Hmm I try to understand the concerns. The data disks are exclusively bound to the machine and can't be reused by other machines. That mean the disks are also bound to the lifecycle of the machine. 

On creation of the machine the data disks will be also created and at this point in time all luns free on the machine, so we could start with 0 lun and enumerate for other data disks.
During lifetime k8s will attach new disks, but k8s (or someone else) will also check for free luns before attaching the disk which is anyways the safest way to attach new disks.
If the MachineClass spec will change (e.g. adding a new data disks) then we would need to roll the machines and new data disks, which contain the name of the new machine in their name, will be created, right?
On deletion of the machine the attached data disks will be deleted together with machine.

Name generation for the data disk shouldn't be also a big deals as we already have a `name` for each data disk. In combination with machine name it should be unique. We only need to validate that the data disk names are unique.

So what's the concern? Losing a disk (e.g. when something unexpected like mcm crashes happen during the deletion process) which would lead to an orphan disk? That could be handled by the safety controller and by adding some tags to the data disk which contain the machine name. So if the machine is gone and belonging data disks are still exists then it could find and delete them.

Don't get me wrong. The proposed approach is ok for the moment, but from my point of view `lun` is an internal detail of the machine and for connivence we should not force the user to specify it. Users should simply need to say I want a data disk, with this name, size and storage type (+ maybe I need also caching). But I understand this is more effort to archive.

hardikdr: I overall agree with the suggestion to keep the lun optional. But I also feel, having luns as part of the disk-names makes the uniqueness stronger. As we don't want to add the complexity of updating the machineclass with generated-luns, let's keep it the current way for now. We can always have an another look at it in the future if we see a need.

dkistner (approved):
/lgtm

We will tackle the things with optional luns for Azure later in a subsequent issue and pr.
Thanks @guydaichs :)

jia-jerry (approved):
/lgtm

hardikdr:
@ggaurav10 @prashanth26  @rewiko  Can you please take one final look at PR, and approve if possible?

ggaurav10 (approved):

hardikdr (approved):

rewiko (approved):
Nice PR @guydaichs :)

prashanth26 (approved):

Issue #393, opened 2020-01-20
Probable Azure improvements for MCM.


hardikdr:
# Issue
We recently faced an issue with the azure provider, where all of the VM-operations in the given Availibility-set was stuck due to the specific disk-detachment issue in one of the VMs.

## What was observed?
- A data-disk was deleted successfully. The same deleted disk was found attached to the other VM. This situation brings the entire AV-set into the inconsistent state, eventually leading to the failure of all the other operations in the same AV-set.

## RCA from Azure:
```
From the backend logs I can see the following scenario that caused the issue and the VM in a failed state after deleting the disk :
 
at 2020-01-10 16:13:50.682, see disk successfully detached
at 2020-01-10 18:21:59.167, disk is successfully deleted
 
at 2020-01-10 18:24:53.160, see attach disk operation for the disk: shoot-garden-abc-dynamic-pvc-abcd-1234-1234
but it failed as the disk cannot be found (expected as already deleted)
this is the operation to put VM in failed state
Why the script you are using is trying to reattach the disk after it was deleted? I can see request number 3 coming from the application : mapping to xyz-account
```

## Quick workaround:
This issue was resolved by recreating the deleted-but-attached disk with the same name and size. The newly created disk re-attached itself to the VM, which eventually led to the success of other stuck operations in AV-set.

# Solution
Based on the RCA from Azure, we need to investigate further the azure-operations we do with disks/VMs. This could also be potentially related to the CCM/KCM issue, as MCM never tries to attach the disk back to the VM.

hardikdr:
cc @ialidzhikov @prashanth26 @vpnachev @amshuman-kr

MSSedusch:
One of our support engineers suggested to update the CRP (Compute Resource Provider) API version to a more recent version to avoid issues with failed Availability Sets. Can you please update the RP versions to

	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute"
	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2019-11-01/network"
	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-05-01/resources"

https://github.com/gardener/machine-controller-manager/blob/master/pkg/driver/driver_azure.go#L33-L35

prashanth26:
We identified MCM restarts during machine creation/deletion (on azure) could lead to orphan VMs.
1. Make a list call to verify any remaining infra before deletion
[...]

PR #406, opened 2020-02-06
Azure: Update Azure SDK + List for orphan dependant resources

dkistner:
Update of the Azure SDK to use more recent version of the Azure service sdks should include improvements to avoid or at least reduce blocking machines which are assigned to availability sets. See https://github.com/gardener/machine-controller-manager/issues/393#issuecomment-581469452

This PR contains a more recent version of the Azure `compute` service sdk, but not yet the recommended one as this is not yet released. We are clarifying if the used one already contain the required improvements.

Furthermore we MCM will now list for orphan dependent resources after their belonging machine has been deleted and try to remove those which still exists. See https://github.com/gardener/machine-controller-manager/issues/393#issuecomment-591945623

cc @MSSedusch 

**Status**
- [x] Updated Azure Go SDK for `compute`, `networking` and `resources`
- [x] Add list call and deletion logic for orphan dependant resources after vm deletion has been completed
- [x] Double check if the referenced Azure Go SDK for `compute` contains the required improvements

Fixes partly #393 

**Release note**:

```improvement operator
MCM will now check after Azure vm deletion if there are still existing dependant resources of the respective vm available and try to delete them to prevent orphan resources.
```


dkistner:
@prashanth26 I have integrated a list call for orphan dependent machine resources after the machine is deleted. If after machine deletion still dependant resources available the it will try to delete them.

@MSSedusch: Do we already know if the used version of the Azure compute go-sdk (`2019-07-01`) incorporate the required improvements?

prashanth26:
@dkistner - Is this PR still WIP or reviewable?

dkistner:
@prashanth26 Sorry I forgot to remove the WIP... Yes if you have time would be nice to have a review.

hardikdr:
@MSSedusch  Can you please take a brief look at the PR, mainly from PoV of your comment [here](https://github.com/gardener/machine-controller-manager/issues/393#issuecomment-581469452) ?

hardikdr:
> TODO: Check if we can use the required version of the compute provider.
@dkistner I am curious, what does this TODO mean in the commit? Is it anything significant enough to be tracked separately?

MSSedusch:
@hardikdr still checking with Azure Compute team if the fix is part of Compute API version 2019-07-01. Will keep this thread updated.

hardikdr:
>@hardikdr still checking with Azure Compute team if the fix is part of Compute API version 2019-07->01. Will keep this thread updated.

Sure, thanks for the update.

dkistner:
> > TODO: Check if we can use the required version of the compute provider.

> @dkistner I am curious, what does this TODO mean in the commit? Is it anything significant enough to be tracked separately?

@hardikdr It is a reminder to check if the included version of the azure compute sdk contians the improvements we are looking for. Lets wait until @MSSedusch confirm. Thanks!

MSSedusch:
@dkistner I had a chat with the Compute team last week and they confirmed that the issue is fixed with Compute API version 2019-07-01

dkistner:
Nice, that's the version we are using. @hardikdr, @prashanth26 Could you please have another look?

hardikdr:
Looks good to me, I'll wait for Prashanth to review once.

pkg/driver/driver_azure.go
```diff
@@ -795,6 +811,42 @@ func (clients *azureDriverClients) deleteVM(ctx context.Context, resourceGroupNa
...
+	nicResults, err := clients.nic.List(ctx, resourceGroupName)
+	if err != nil {
+		return err
+	}
+	for _, nic := range nicResults.Values() {
+		if nic.Name == nil || !strings.Contains(*nic.Name, vmName) {
+			continue
+		}
+		klog.V(2).Infof("Found orphan NIC %q belonging to deleted machine %q", *nic.Name, vmName)
+		if err := clients.deleteNIC(ctx, resourceGroupName, *nic.Name); err != nil {
```
ggaurav10: Should we ignore the `NotFound` error just in case the resource was deleted after the `List` call but before `Delete` call could be made.

pkg/driver/driver_azure.go
```diff
@@ -795,6 +811,42 @@ func (clients *azureDriverClients) deleteVM(ctx context.Context, resourceGroupNa
...
+	diskResults, err := clients.disk.ListByResourceGroup(ctx, resourceGroupName)
+	if err != nil {
+		return err
+	}
+	for _, disk := range diskResults.Values() {
+		if disk.Name == nil || !strings.Contains(*disk.Name, vmName) {
+			continue
+		}
+		klog.V(2).Infof("Found orphan disk %q belonging to deleted machine %q", *disk.Name, vmName)
+		if err := clients.deleteDisk(ctx, resourceGroupName, *disk.Name); err != nil {
```
ggaurav10: Same as above

pkg/driver/driver_azure.go
```diff
@@ -24,15 +24,16 @@ import (
...
 	"github.com/gardener/machine-controller-manager/pkg/metrics"
 	"github.com/prometheus/client_golang/prometheus"
 	corev1 "k8s.io/api/core/v1"
 
-	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute"
-	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2018-11-01/network"
-	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources"
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2019-11-01/network"
+	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-05-01/resources"
```
prashanth26: Should we be using a version later than `2019-07-01` here? Or it doesn't matter?

MSSedusch: In general it is always a good idea to use the latest one available but for the bug fix, only the Compute API version is relevant.

dkistner: I just used the mentioned azure service sdks.
We could also introduce now the most recent ones? WDYT?

pkg/driver/driver_azure.go
dkistner: Hmm should we do that already in the `deleteNIC` function? We could check if the future status is succeeded and if the status code == 404 and then return no error?
Same for `deleteDisk` funtion.
WDYT?

ggaurav10: yes. Sounds good. Better to be handled at the deepest level.

pkg/driver/driver_azure.go
MSSedusch: sounds good to me.

pkg/driver/driver_azure.go
dkistner: I just saw we have already a `notFound` function to determine if an error is a 404. I simply used this one.

pkg/driver/driver_azure.go
dkistner: Same here, see comment above.

pkg/driver/driver_azure.go
dkistner: Let's go one with this version of the azure go sdks as they contain the required fixes. Upgrading to the latest ones seems to be more effort because it seems that there are some things have changed. Will do that in a subsequent pr.

dkistner:
Hmm the test failed, but seems to be unrelated to any Azure changes.
The gcp test failed.
```
E0319 12:04:19.769909   13688 controllermanager.go:329] unable to get all supported resources from server: unable to retrieve the complete list of server APIs: metrics.k8s.io/v1beta1: the server is currently unable to handle the request
...
			----- Test failure logs -----------
Integration test for the following cloud providers have failed. Kindly re-check your PR.
	- gcp
```

PR #408, opened 2020-02-11
Bugfix: set deleteOnTermination to true by default and tag volumes

hardikdr:
**What this PR does / why we need it**: Setting the `deleteOnTermination` to true by default and tag the volumes the same as instances.

Fixes #

**Release note**:

```noteworthy operator
Bugfix: Set deleteOnTermination to true by default for volumes. Disks that are created with the instance are deleted with instance termination. 
```


prashanth26:
Tests need fixing, otherwise looks good.

pkg/driver/driver_aws.go
```diff
@@ -167,20 +177,23 @@ func (d *AWSDriver) generateBlockDevices(blockDevices []v1alpha1.AWSBlockDeviceM
...
-				DeleteOnTermination: aws.Bool(deleteOnTermination),
-				Encrypted:           aws.Bool(encrypted),
-				VolumeSize:          aws.Int64(volumeSize),
-				VolumeType:          aws.String(volumeType),
+				Encrypted:  aws.Bool(encrypted),
+				VolumeSize: aws.Int64(volumeSize),
+				VolumeType: aws.String(volumeType),
 			},
 		}
+		if deleteOnTermination != nil {
```
rfranzke: This isn't needed, is it?

hardikdr: This is just to be sure, and could otherwise result in unexpected behavior from AWS.

PR #412, opened 2020-02-14
OpenStack: Print API request/response debug, when verbosity is set to 6

kayrus:

Setting the verbosity to 6 allows to print detailed OpenStack API requests/responses.

**Release note**:

```improvement operator
Openstack: Print API request/response debug when verbosity is set to 6
```

cc @rfranzke . It might be an idea to introduce an interface to enable verbose logging dynamically, because if you change the manifest for one particular shoot, it will be reconciled with the defaults. Any thoughts on how to do this?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,11 +14,10 @@ require (
 	github.com/emicklei/go-restful v2.9.6+incompatible // indirect
 	github.com/go-ini/ini v1.36.0 // indirect
 	github.com/go-openapi/spec v0.19.2
-	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
 	github.com/golang/groupcache v0.0.0-20180513044358-24b0969c4cb7 // indirect
 	github.com/googleapis/gnostic v0.2.0 // indirect
 	github.com/gophercloud/gophercloud v0.7.0
-	github.com/gophercloud/utils v0.0.0-20190527093828-25f1b77b8c03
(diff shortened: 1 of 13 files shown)


kayrus:
@mvladev Is it possible to use `/configz` http handler to dynamically modify the klog verbosity settings?

prashanth26:
> It might be an idea to introduce an interface to enable verbose logging dynamically, because if you change the manifest for one particular shoot, it will be reconciled with the defaults. Any thoughts on how to do this?

Last I checked this is not possible. Refer - https://github.com/google/glog/issues/452.

kayrus:
@prashanth26 fine. BTW: glog is not used here, but klog.

hardikdr:
I would be really interested in knowing if there is a good way to dynamically update the log levels. Do you have suggestions, around how can that be achieved with klog ?

kayrus:
> As we don't have CI configured for OpenStack, I am curious if you got a chance to test this PR, if works as expected ?

@dkistner @dguendisch @rfranzke can you test this change against openstack environment?

> I would be really interested in knowing if there is a good way to dynamically update the log levels. Do you have suggestions, around how can that be achieved with klog ?

I don't think this is impossible and I already have an idea, but I'd focus on dynamic logic within a separate PR

hardikdr:
>I don't think this is impossible and I already have an idea, but I'd focus on dynamic logic within a separate PR

Sure, thanks. 
I'll then go ahead and merge this one soon, later I will try to find some time to test the openstack provider separately before next mcm release.

PR #414, opened 2020-02-17
Add gcp local ssd disk support

tennix:

This PR enables user to add local ssd disks on GCP, the local ssd disk supports two interfaces: `NVME` and `SCSI`. See details in the below documents:
* https://cloud.google.com/compute/docs/disks/local-ssd#create_local_ssd
* https://pkg.go.dev/google.golang.org/api@v0.17.0/compute/v1?tab=doc#AttachedDisk

I've tested this on GCP and created 4 local NVME ssd disks. Note a GCE instance can only attach at most 8 local ssd disks. So there may need a validation for the total number of local ssd disks.

**Release note**:

```improvement operator
Add GCP local ssd disk support
```


hardikdr:
@tennix Thanks a lot for the PR, we'll take a look soon.

prashanth26:
@hardikdr @ggaurav10 - Do you see any concerns here?

pkg/driver/driver_gcp.go
```diff
@@ -84,16 +84,29 @@ func (d *GCPDriver) Create() (string, string, error) {
...
-				DiskSizeGb:  disk.SizeGb,
-				DiskType:    fmt.Sprintf("zones/%s/diskTypes/%s", zone, disk.Type),
-				Labels:      disk.Labels,
-				SourceImage: disk.Image,
-			},
-		})
+		var attachedDisk compute.AttachedDisk
+		if disk.Type == "SCRATCH" {
+			attachedDisk = compute.AttachedDisk{
+				AutoDelete: disk.AutoDelete,
```
ggaurav10: This default to `false`, right?
@hardikdr @prashanth26 - Wouldn't it lead to similar issues that we recently saw of orphan disks when a VM is deleted.

pkg/driver/driver_gcp.go
```diff
@@ -84,16 +84,29 @@ func (d *GCPDriver) Create() (string, string, error) {
...
+				Type:       disk.Type,
+				Interface:  disk.Interface,
+				InitializeParams: &compute.AttachedDiskInitializeParams{
+					DiskType: fmt.Sprintf("zones/%s/diskTypes/%s", zone, "local-ssd"),
+				},
+			}
+		} else {
+			attachedDisk = compute.AttachedDisk{
+				AutoDelete: disk.AutoDelete,
```
ggaurav10: same as above

pkg/driver/driver_gcp.go
prashanth26: Good catch. What would be the best way to handle this? Default the values to true on MCM or Gardener maybe?

ggaurav10: IMO, defaulting should be done in MCM, keeping it mind that it can be used independently of Gardener

prashanth26: @tennix - Can we default this flag to true in the MCM?

pkg/driver/driver_gcp.go
prashanth26: @tennix - Can we default this flag to true in the MCM?

pkg/driver/driver_gcp.go
tennix: I don't think it's related to this. Unlike to persistent disks, the local ssd disks will be deleted once the instance is deleted. But I can change the default value to `true` if you like.

pkg/driver/driver_gcp.go
tennix: For persistent disks, is it better to default to false so if the VM is deleted there is still a chance to recover the data in persistent disk especially for production environment?

pkg/driver/driver_gcp.go
hardikdr: I'd actually prefer to delete these disks by default with the VM, unless specified in machinieclass. Considering the fact that we constantly create/delete the machines with autoscaler, we may end up with many such orphan disks, and we don't have active automation to keep track of these disks later.

hardikdr: According to the documentation the local-ssds are expected to be deleted with the VM termination. - https://cloud.google.com/compute/docs/disks#localssds . 
- @tennix please correct if that's not the case.

Though, to save us from the possible changes at gcp's behaviour, I would prefer to set it to `true`. 
- The issue actually is, the API type for `AutoDelete` is defined as `bool` and leaves no good way to check if it's set at MachineClass or not. 
- @tennix Please feel free to change the type to `*bool` from `bool`, that'll help you in checking `AutoDelete==nil`(implies not set at MachineClass), and default to true.

pkg/driver/driver_gcp.go
hardikdr: > @tennix - Can we default this flag to true in the MCM?

Actually, at the moment, Gardener sets the `AutoDelete` to `true` for all the GCPMachineClasses, hence the current codepath should not impact us immediately.

>For persistent disks, is it better to default to false so if the VM is deleted there is still a chance to recover the data in persistent disk especially for production environment?

I would actually suggest to set the `AutoDelete=true` even for the persistent disks, and let the user change the behaviour if needed via MachineClass. 
- The change to the api type to `*bool` would be useful here as well.

pkg/driver/driver_gcp.go
tennix: Okay, I'll change default `AutoDelete` to `true`.

hardikdr:
@prashanth @ggaurav10 can you please take a quick look and approve if fine with it?

pkg/apis/machine/v1alpha1/types.go
```diff
@@ -1108,10 +1108,11 @@ type GCPMachineClassSpec struct {
 
 // GCPDisk describes disks for GCP.
 type GCPDisk struct {
-	AutoDelete bool              `json:"autoDelete"`
+	AutoDelete *bool             `json:"autoDelete"`
```
prashanth26, line 1111: Need to make this change on - `pkg/apis/machine/types.go` as well right? And then run regenerate clients.

tennix: Okay, I've addressed this. PTAL again @prashanth26

prashanth26: Thanks.

PR #416, opened 2020-02-17
Skip eviction during drain if ForceDeletePods.

### LINKED ISSUE #415: PDB violation blocks drain during forced machine deletion
### Context

The intended original behaviour of drain during hibernation and cluster deletion is that all the pods in the node will be evicted in parallel and PDBs (if any) would be violated. The `ForceDeletePods` flag is set by gardenet/worker-pool extension during hibernation/ cluster deletion to inform the drain logic to [take this path](https://github.com/gardener/machine-controller-manager/blob/2e4c6cf44d4e47e94cadefed47cb9ddcf00dc68f/pkg/controller/drain.go#L423). The `attemptEvict flag` still is [being passed](https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/drain.go#L369) based on [whether apiserver supports the eviction API](https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/drain.go#L360). This [flag is reset](https://github.com/gardener/machine-controller-manager/blob/2e4c6cf44d4e47e94cadefed47cb9ddcf00dc68f/pkg/controller/drain.go#L823) after `MaxEvictRetries` to finally force a deletion. This was intended as a way to be nice to the pods even in this scenario.

### Problem
This was fine when the `MaxEvictRetries` value was [small earlier](https://github.com/gardener/machine-controller-manager/blob/6b7e1f350ae4bb5a5a40f70af424b586e0fb2892/pkg/controller/drain.go#L89). But after we [increased it to match the drain timeout](https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/drain.go#L116), the logic gets no chance to  do the force deletion until the drain timeout happens.
 
So, it is a bug that got introduced by the change to the `MaxEvictRetries` value recently (which was required because of PDB violations during regular machine deletion).

### Proposed Solution
This can be fixed by one of the following approaches if `ForceDeletePods` option is enabled.
1. Having a smaller `MaxEvictRetries` value
2.  always passing the attemptEvict as a false

I think option 2 would be simpler.

amshuman-kr:

Fixes #415 

No eviction is attempted if `ForceDeletePods` is enabled. Direct pod deletion is done instead. The unit tests have been adjusted for the new behaviour. The rest of the logic remains the same.

**Release note**:

```improvement operator
Skip eviction during drain if ForceDeletePods is enabled.
```


prashanth26:
Irrespective of the decision of the discussions on #415 . We can still merge this right?

amshuman-kr:
> Irrespective of the decision of the discussions on #415 . We can still merge this right?

Yes. And we should raise a PR on gardener to set the `ForceDeletePods` flag during hibernation.

Issue #421, opened 2020-03-02
Enhance machine-deletion flow to avoid leaking resources


hardikdr:
## Issues:
We identified the case where the following set of events happening exactly at the same time could cause MCM to leak a resource. 

T1.  New machine-objects are created, maybe via scale-up.

T2. MCM processes fresh machine-objects by making a cloud call of CreateVMs().
   - `Machine.Status` and `Spec.ProviderID` is set only after this Create call returns. 

T2. Before CreateVM() call returns, MCM is brought down and re-created.

T2. At the same time, machine-objects are deleted may be via shoot-deletion/scale-down.

Earlier MCM used to block the deletion of the machine-objects if status/providerID is missing but with the PR #364 we changed it for good. Now machine-set controller allows the deletion of the machine-object without status/providerID.

## Improvement: 
We need to safeguard the MCM with the possibility that a machine-object without status/providerID might be backed by a real VM. We should basically make a listVM() call before deleting such fresh machine objects, as we might not have providerID , we could filter via machine-name tags.

hardikdr:
Closing this in favor of #432

PR #432, opened 2020-03-19
Improve machine creation and deletion flow by listing the machines before actual creation/deletion

### LINKED ISSUE #423: Restore ACID properties on MCM
## Issue

* Currently the MCM loses state on MCM restarts. 
* Also the Machine and machine-set controller have races while creating/deleting.

## Solution

Reenable transaction ACID properties for machine creation/deletion.

hardikdr:
**What this PR does / why we need it**: This PR addresses the cases when an unfortunate restart of the MCM can make it lose the state. MCM now lists the VMs in the cloud, before actually creating or deleting the machines. 
- This PR also enables MCM to adopt the dynamically node-object in case it is not populated on the machine-object while creation. 

Fixes #423 

**Release note**:

```improvement operator
Improves the machine-creation and machine-deletion flow by listing and verifying the state of the machines before actual creation and deletion. 
```
```improvement operator
MCM now dynamically maps the node-objects with machines if `Status.Node` is not set. 
```

### DIFF
--- a/pkg/controller/machine.go
+++ b/pkg/controller/machine.go
@@ -266,14 +266,39 @@ func (c *controller) getMachineFromNode(nodeName string) (*v1alpha1.Machine, err
 }
 
 func (c *controller) updateMachineState(machine *v1alpha1.Machine) (*v1alpha1.Machine, error) {
+	nodeName := machine.Status.Node
 
-	if machine.Status.Node == "" {
-		// There are no objects mapped to this machine object
-		// Hence node status need not be propogated to machine object
-		return machine, nil
+	if nodeName == "" {
+		// Check if any existing node-object can be adopted.
+		nodeList, err := c.nodeLister.List(labels.Everything())
+		if err != nil {
+			klog.Errorf("Could not list the nodes due to error: %v", err)
+			return machine, err
+		}
+		for _, node := range nodeList {
+			nID, mID := decodeMachineID(node.Spec.ProviderID), decodeMachineID(machine.Spec.ProviderID)
+			if nID == mID {
+				klog.V(2).Infof("Adopting the node object %s for machine %s", node.Name, machine.Name)
+				nodeName = node.Name
+				clone := machine.DeepCopy()
+				clone.Status.Node = nodeName
+				clone, err = c.controlMachineClient.Machines(clone.Namespace).UpdateStatus(clone)
+				if err != nil {
+					klog.Errorf("Could not update status of the machine-object %s due to error %v", machine.Name, err)
+					return machine, err
+				}
+				break
+			}
+		}
+		// Couldnt adopt any node-object.
+		if nodeName == "" {
+			// There are no objects mapped to this machine object
+			// Hence node status need not be propogated to machine object
+			return machine, nil
+		}
 	}
 
-	node, err := c.nodeLister.Get(machine.Status.Node)
+	node, err := c.nodeLister.Get(nodeName)
 	if err != nil && apierrors.IsNotFound(err) {
 		// Node object is not found
 
@@ -325,7 +350,7 @@ func (c *controller) updateMachineState(machine *v1alpha1.Machine) (*v1alpha1.Ma
 		clone.Labels = make(map[string]string)
 	}
 
-	if _, ok := clone.Labels["node"]; !ok {
+	if n := clone.Labels["node"]; n == "" {
 		clone.Labels["node"] = machine.Status.Node
 		machine, err = c.controlMachineClient.Machines(clone.Namespace).Update(clone)
 		if err != nil {
@@ -344,6 +369,7 @@ func (c *controller) updateMachineState(machine *v1alpha1.Machine) (*v1alpha1.Ma
 
 func (c *controller) machineCreate(machine *v1alpha1.Machine, driver driver.Driver) error {
 	klog.V(2).Infof("Creating machine %q, please wait!", machine.Name)
+	var actualProviderID, nodeName string
 
 	err := c.addBootstrapTokenToUserData(machine.Name, driver)
 	if err != nil {
@@ -362,7 +388,18 @@ func (c *controller) machineCreate(machine *v1alpha1.Machine, driver driver.Driv
 		c.updateMachineStatus(machine, lastOperation, currentStatus)
 		return err
 	}
-	actualProviderID, nodeName, err := driver.Create()
+	// Before actually creating the machine, we should once check and adopt if the virtual machine already exists.
+	VMList, err := driver.GetVMs("")
+	for providerID, machineName := range VMList {
+		if machineName == machine.Name {
+			klog.V(2).Infof("Adopted an existing VM %s for machine object %s.", providerID, machineName)
+			actualProviderID = providerID
+		}
+	}
+	if actualProviderID == "" {
+		actualProviderID, nodeName, err = driver.Create()
+	}
+
 	if err != nil {
 		klog.Errorf("Error while creating machine %s: %s", machine.Name, err.Error())
 		lastOperation := v1alpha1.LastOperation{
@@ -385,7 +422,7 @@ func (c *controller) machineCreate(machine *v1alpha1.Machine, driver driver.Driv
 
 		return err
 	}
-	klog.V(2).Infof("Created machine: %q, MachineID: %s", machine.Name, actualProviderID)
+	klog.V(2).Infof("Created/Adopted machine: %q, MachineID: %s", machine.Name, actualProviderID)
 
 	for {
 		machineName := machine.Name
@@ -614,6 +651,23 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
 			}
 
 			err = driver.Delete(machineID)
+		} else {
+			klog.V(2).Infof("Missing ProviderID when deleting the machine object. %s might have been deleted too soon.", machine.Name)
+			// As MachineID is missing, we should check once if actual VM was created but MachineID was not updated on machine-object.
+			// We list VMs and check if any one them map with the given machine-object.
+			VMList, _ := driver.GetVMs("")
+			for providerID, machineName := range VMList {
+				if machineName == machine.Name {
+					klog.V(2).Infof("Deleting the VM %s backing the machine-object %s.", providerID, machine.Name)
+					err = driver.Delete(providerID)
+					if err != nil {
+						klog.Errorf("Error deleting the VM %s backing the machine-object %s due to error %v", providerID, machine.Name, err)
+						// Not returning error so that status on the machine object can be updated in the next step if errored.
+					} else {
+						klog.V(2).Infof("VM %s backing the machine-object %s is deleted successfully", providerID, machine.Name)
+					}
+				}
+			}
 		}
 
 		if err != nil {
@@ -951,4 +1005,10 @@ func shouldReconcileMachine(machine *v1alpha1.Machine, now time.Time) bool {
 	// TODO add more cases where this will be false
 
 	return true
+}
+
+// decodeMachineID is a generic way of decoding the Spec.ProviderID field of node-objects.
+func decodeMachineID(id string) string {
+	splitProviderID := strings.Split(id, "/")
+	return splitProviderID[len(splitProviderID)-1]
 }
--- a/pkg/controller/machine.go
+++ b/pkg/controller/machine.go
@@ -390,6 +390,10 @@ func (c *controller) machineCreate(machine *v1alpha1.Machine, driver driver.Driv
 	}
 	// Before actually creating the machine, we should once check and adopt if the virtual machine already exists.
 	VMList, err := driver.GetVMs("")
+	if err != nil {
+		klog.Errorf("Failed to list VMs before creating machine %q %+v", machine.Name, err)
+		return err
+	}
 	for providerID, machineName := range VMList {
 		if machineName == machine.Name {
 			klog.V(2).Infof("Adopted an existing VM %s for machine object %s.", providerID, machineName)
@@ -655,7 +659,11 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
 			klog.V(2).Infof("Missing ProviderID when deleting the machine object. %s might have been deleted too soon.", machine.Name)
 			// As MachineID is missing, we should check once if actual VM was created but MachineID was not updated on machine-object.
 			// We list VMs and check if any one them map with the given machine-object.
-			VMList, _ := driver.GetVMs("")
+			VMList, err := driver.GetVMs("")
+			if err != nil {
+				klog.Errorf("Failed to list VMs while deleting the machine %q %v", machine.Name, err)
+				return err
+			}
 			for providerID, machineName := range VMList {
 				if machineName == machine.Name {
 					klog.V(2).Infof("Deleting the VM %s backing the machine-object %s.", providerID, machine.Name)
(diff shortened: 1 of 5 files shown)


pkg/controller/machine.go
```diff
@@ -614,6 +651,23 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
...
+			// As MachineID is missing, we should check once if actual VM was created but MachineID was not updated on machine-object.
+			// We list VMs and check if any one them map with the given machine-object.
+			VMList, _ := driver.GetVMs("")
+			for providerID, machineName := range VMList {
+				if machineName == machine.Name {
+					klog.V(2).Infof("Deleting the VM %s backing the machine-object %s.", providerID, machine.Name)
+					err = driver.Delete(providerID)
+					if err != nil {
+						klog.Errorf("Error deleting the VM %s backing the machine-object %s due to error %v", providerID, machine.Name, err)
+						// Not returning error so that status on the machine object can be updated in the next step if errored.
```
prashanth26: I didn't follow this. Won't the machine object be deleted in the next step and this orphan VM continues to be there? (unless the orphan VM controller acts on it). And incase of shoot deletion this might still lead to situations where VM persists once the machine object is deleted?

pkg/controller/machine.go
```diff
@@ -266,14 +266,39 @@ func (c *controller) getMachineFromNode(nodeName string) (*v1alpha1.Machine, err
...
-		return machine, nil
+	if nodeName == "" {
+		// Check if any existing node-object can be adopted.
+		nodeList, err := c.nodeLister.List(labels.Everything())
+		if err != nil {
+			klog.Errorf("Could not list the nodes due to error: %v", err)
+			return machine, err
+		}
+		for _, node := range nodeList {
+			nID, mID := decodeMachineID(node.Spec.ProviderID), decodeMachineID(machine.Spec.ProviderID)
```
prashanth26: Is this required? I think comparing both the complete providerIDs should be enough right?

pkg/controller/machine.go
```diff
@@ -362,7 +388,18 @@ func (c *controller) machineCreate(machine *v1alpha1.Machine, driver driver.Driv
...
+		if machineName == machine.Name {
+			klog.V(2).Infof("Adopted an existing VM %s for machine object %s.", providerID, machineName)
+			actualProviderID = providerID
+		}
+	}
+	if actualProviderID == "" {
+		actualProviderID, nodeName, err = driver.Create()
+	}
+
 	if err != nil {
```
ggaurav10: This now ends up checking `err` for the added call `VMList, err := driver.GetVMs("")` instead of `actualProviderID, nodeName, err = driver.Create()`.
Error check needs to be added separately for the added call.

pkg/controller/machine.go
```diff
@@ -614,6 +651,23 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
 			}
 
 			err = driver.Delete(machineID)
+		} else {
+			klog.V(2).Infof("Missing ProviderID when deleting the machine object. %s might have been deleted too soon.", machine.Name)
+			// As MachineID is missing, we should check once if actual VM was created but MachineID was not updated on machine-object.
+			// We list VMs and check if any one them map with the given machine-object.
+			VMList, _ := driver.GetVMs("")
```
ggaurav10: Same as above. Better not to ignore any error here.

pkg/controller/machine.go
```diff
@@ -362,7 +388,18 @@ func (c *controller) machineCreate(machine *v1alpha1.Machine, driver driver.Driv
...
+	// Before actually creating the machine, we should once check and adopt if the virtual machine already exists.
+	VMList, err := driver.GetVMs("")
+	for providerID, machineName := range VMList {
+		if machineName == machine.Name {
+			klog.V(2).Infof("Adopted an existing VM %s for machine object %s.", providerID, machineName)
+			actualProviderID = providerID
+		}
+	}
+	if actualProviderID == "" {
+		actualProviderID, nodeName, err = driver.Create()
```
ggaurav10: In case VM already exists, we don't get to this logic, and we end up updating `clone.Status.Node = ""` since `nodeName = ""`.
Is this supposed to be taken care of in the next `updateMachineState()` call? It could take 10 minutes according to the deferred `enqueueMachineAfter` call, right.
Same goes for `syncMachineNodeTemplates` call.

pkg/controller/machine.go
prashanth26: Good catch.

pkg/controller/machine.go
hardikdr: >Is this required? I think comparing both the complete providerIDs should be enough right?

Yes, this seems to be required as `ProviderID` at machine object and node-object have slightly different formatting based on providers. And content is not exactly the same, but the last part of the IDs are always uuid and same. 

> we don't need to make the call decodeMachineID(machine.Spec.ProviderID) in every iteration.
This can be done just once before starting the iterations

Node-call has to be inside and machine-call could be outside the loop. Considering this gets executed only in rare cases when node-object-name is missing, I kept both of them inside for better readability, hope that's fine.

prashanth26: > Yes, this seems to be required as ProviderID at machine object and node-object have slightly different formatting based on providers. And content is not exactly the same, but the last part of the IDs are always uuid and same.

When a machine is trying to adopt a node, the machine and node should belong to the same provider and the complete providerID should match right? This decode might not be required right?

prashanth26: If it's not the case, can you provide me an example where you have seen differently?

hardikdr: Yes, of course, they are the same providers. But I observed changes in the ProviderIDs at machine-object and node-objects, for instance, 
- aws:///eu-central-1/i-0123442 - at machine object.
- aws:///eu-central-1a/i-0123442 - at node object. 

This can happen, as at machine we calculate the providerID via driver-implementation but providerID on node-object comes from the kubelet. We should probably use the same kubelet function everywhere.

pkg/controller/machine.go
```diff
@@ -266,14 +266,39 @@ func (c *controller) getMachineFromNode(nodeName string) (*v1alpha1.Machine, err
...
+			return machine, err
+		}
+		for _, node := range nodeList {
+			nID, mID := decodeMachineID(node.Spec.ProviderID), decodeMachineID(machine.Spec.ProviderID)
+			if nID == mID {
+				klog.V(2).Infof("Adopting the node object %s for machine %s", node.Name, machine.Name)
+				nodeName = node.Name
+				clone := machine.DeepCopy()
+				clone.Status.Node = nodeName
+				clone, err = c.controlMachineClient.Machines(clone.Namespace).UpdateStatus(clone)
```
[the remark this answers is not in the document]
hardikdr: >We need to update clone.Labels["node"], .status.LastOperation and .status.CurrentStatus also here.

`clone.Labels["node"]` will be updated [here](https://github.com/gardener/machine-controller-manager/blob/859ef7504d2ff3efa715bfd6c84317c3267de835/pkg/controller/machine.go#L353) in the same function. 

I couldn't understand why `LastOperation` or `CurrentStatus` needs to be updated here, can you please elaborate. Basically, normal machine-controller flow should apply, this part of the code should execute only when `status.Node` is missing for any reason- possibly when MCM restarts during machineCreate() and updating the node-name to machine-object failed. As at the moment, our only source of node-name is from the `driver.Create()`, the output of which might be lost, we decided to adopt the existing node-object via providerID.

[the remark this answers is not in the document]
hardikdr: >In case this change is also supposed to take care of https://github.com/gardener/machine-controller-manager/issues/423#issuecomment-597628323, then mID will not be available as mentioned in #421

I couldn't understand this, can you please explain a little more. Basically the specific changes here of adopting-node are for the creation-flow.

pkg/controller/machine.go
```diff
@@ -325,7 +350,7 @@ func (c *controller) updateMachineState(machine *v1alpha1.Machine) (*v1alpha1.Ma
 		clone.Labels = make(map[string]string)
 	}
 
-	if _, ok := clone.Labels["node"]; !ok {
+	if n := clone.Labels["node"]; n == "" {
```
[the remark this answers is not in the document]
hardikdr: >The right way is to depend on the second (ok) field. I have read this somewhere. I think we should revert this back.

`if _, ok := clone.Labels["node"]; !ok` would tell us if key exists or not, but I wanted to explicitly check if key doesn't exist but also not set to ""(empty string). As the code [here](https://github.com/gardener/machine-controller-manager/blob/859ef7504d2ff3efa715bfd6c84317c3267de835/pkg/controller/machine.go#L466-L469) sets the label of `node: ""` on machine-object if MCM restarts immediately after `driver.Create()`, which can slip through if we revert the code back to relying on second(ok) argument.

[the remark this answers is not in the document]
hardikdr: >we should also check for: clone.Labels["node"] != machine.Status.Node

Yes, we should have such a check irrespective of the creation/deletion flow, not sure if it falls under the scope of this PR though.

pkg/controller/machine.go
hardikdr: Yes, this is a very good point. Basically, if the nodeName is mising then both node-label and `Status.Node` will be set to "" unfortunately, because we dont have a way to get the nodeName at that time. We reconcile the machine-objects on events[add/update] from node-object as well, it means both label and `Status.Node` will adopt the node-name on the next reconcilliation of the machine-object, which doesn't have to be 10mins but till the time node really joins.

pkg/controller/machine.go
hardikdr: Hm, intersting catch. So basically if `GetVMs` fails, it's better to err back than creating the new machine. I'll make a change here to handle the error, thanks.

pkg/controller/machine.go
hardikdr: Yes, it makes sense. Thanks for catching this, I shall handle the error here.

pkg/controller/machine.go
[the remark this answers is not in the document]
ggaurav10: we should still be checking for 2nd `ok` argument i believe. If the crash happens immediately after `driver.Create()` then the `"node"` key would not even exist in `clone.Labels`. And this call itself will fail.
It might be worth testing the scenario by intensionally crashing/panicking MCM just after `driver.Create()` returns.

pkg/controller/machine.go
hardikdr: If `err` is set here by `driver.Delete()`, then the check in the next block `err != nil`should return the same `err` after setting it on the machine-status.

pkg/controller/machine.go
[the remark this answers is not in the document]
hardikdr: Afaik, if the `key` doesn't exist in the map[string]string, then the expression `k = label[key]` should return "", and not fail/panic.

pkg/controller/machine.go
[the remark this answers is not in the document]
ggaurav10: May be i need help with understanding the scenario.
Consider the following case:
T1: machine is created
T2. MCM processes fresh machine-objects by making a cloud call of CreateVMs().
- Machine.Status and Spec.ProviderID is set only after this Create call returns.

T2. Before CreateVM() call returns, MCM is brought down/crashed
T3. Node joins
T4. MCM comes up after several retries (misses the node add event)
T5. `updateMachineState` is called by machine reconcile
- Does't see provider ID in spec

T6. `machineCreate` is called, and since machine exists, the provider ID is updated, and machine phase is set to "Pending"
T7. Machine is enqueued after 10 min
T7. Machine health timeout of 10 min (or less) expires and the machine is marked as failed

Is this a scenario that could happen?

pkg/controller/machine.go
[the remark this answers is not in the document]
ggaurav10: I was mainly concerned about [this](https://github.com/gardener/machine-controller-manager/blob/859ef7504d2ff3efa715bfd6c84317c3267de835/pkg/controller/machine.go#L466-L467) check which is missing now.
If clone.Labels == nil, then the following assignment will panic, right?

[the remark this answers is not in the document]
ggaurav10: The labels may already be present if the machine is created by machineset controller, but if a machine alone is created then we might be in trouble. WDYT?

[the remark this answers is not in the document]
hardikdr: I think even if the map is set to nil, it should still not panic. I think it should only return the "" value back if the key is not found.

pkg/controller/machine.go
prashanth26: Okay if this is the case. The current implementation makes sense. However, we need to update our driver create to return the right parameter then as per the kubelet.

pkg/controller/machine.go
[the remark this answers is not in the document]
prashanth26:
```suggestion
	if nodeLabel, exists := clone.Labels["node"]; !exists &&  nodeLabel== "" {
```
How about we check both scenarios explicitly. Just to be safe? I hope the condition i have written is what you have explained.

pkg/controller/machine.go
prashanth26: Okay, then it's alright. I didn't understand the comment earlier.

pkg/controller/machine.go
hardikdr: done

pkg/controller/machine.go
hardikdr: done

pkg/controller/machine.go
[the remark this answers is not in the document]
hardikdr: This is certainly a very good point and probably deserves more testing/experiments, as I  am not sure of the exact outcome of the above sequence. Would you be fine if we try to separately address this, outside from current PR?

pkg/controller/machine.go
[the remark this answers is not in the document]
hardikdr: If `node: ""`  label exists on the machine-object, it will not pass the above-suggested check, right? 

We could probably use `!exists ||  nodeLabel== ""` but it seems to have same impact as the already written check.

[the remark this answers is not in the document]
ggaurav10: >  think even if the map is set to nil, it should still not panic. I think it should only return the "" value back if the key is not found.
Yes, but the next assignment `clone.Labels["node"] = machine.Status.Node` panics in this case.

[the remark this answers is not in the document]
prashanth26: > Yes, but the next assignment clone.Labels["node"] = machine.Status.Node panics in this case.

I don't think it should panic considering he is initializing the map a step before. Here - https://github.com/gardener/machine-controller-manager/blob/859ef7504d2ff3efa715bfd6c84317c3267de835/pkg/controller/machine.go#L350. 

> but I wanted to explicitly check if key doesn't exist but (I am assuming but you mean and right?) also not set to ""(empty string)

If that's the case i think it would be 
```suggestion
	if nodeLabel, exists := clone.Labels["node"]; !exists && nodeLabel== "" {
```

Or maybe i understood it wrongly. You can take a call on the condition. But i would suggest you to use the exists/ok value also for the condition check. Refer - https://www.digitalocean.com/community/tutorials/understanding-maps-in-go#checking-existence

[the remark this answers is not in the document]
ggaurav10: > I don't think it should panic considering he is initializing the map a step before. Here -

ah. missed that it was already there 2 lines above.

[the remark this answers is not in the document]
prashanth26: Had a call with hardik and is resolved from my side.

pkg/controller/machine.go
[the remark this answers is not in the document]
ggaurav10: As discussed over call, we will address these concerns separately.

pkg/controller/machine.go
ggaurav10: As discussed over call, we can address these concerns separately.

PR #436, opened 2020-03-22
Support VM Image ID for Azure

vpnachev:
Support VM Image ID for Azure.
With this change VMs can be created not only from URN, but also from image ID.
This is necessary when the VM images are not available in the marketplace and URN cannot be used. 

Fixes #

**Release note**:

```improvement operator
Azure VMs now can be created with image ID.
```


vpnachev:
This PR is now ready for review.

hardikdr:
@dkistner Can you please take a quick look at the PR, if possible?

dkistner:
Yes sorry, will have a look today.

PR #441, opened 2020-03-30
Support multiple network interfaces in the AWS driver

zuzzas:

There is an adequate and thought out level of support for multiple network interfaces in AWSMachineClass types and validations. It seems that the only thing missing is the actual driver implementation?

As for the "source-dest check" options: we've got some peculiar network routing designs in some of our Kubernetes clusters.

I've removed subnetID string slice conversion and corresponding test cases. There is a simple AWS SDK [function](https://docs.aws.amazon.com/sdk-for-go/api/aws/#StringSlice) that requires neither testing nor error checking, it's a simple type conversion.

**Release note**:

```improvement user
Support multiple network interfaces in the AWS driver
```


hardikdr:
thanks a lot for the PR @zuzzas.

hardikdr:
@afritzler @kayrus @dguendisch could you please also take a look if possible.

pkg/driver/driver_aws.go
```diff
@@ -97,25 +97,32 @@ func (d *AWSDriver) Create() (string, string, error) {
...
 
-	securityGroupIDs, err := d.generateSecurityGroups(d.AWSMachineClass.Spec.NetworkInterfaces[0].SecurityGroupIDs)
-	if err != nil {
-		return "Error", "Error", err
+	var networkInterfaceSpecs []*ec2.InstanceNetworkInterfaceSpecification
+	for i, netIf := range d.AWSMachineClass.Spec.NetworkInterfaces {
+		spec := &ec2.InstanceNetworkInterfaceSpecification{
+			Groups:                   aws.StringSlice(netIf.SecurityGroupIDs),
+			DeviceIndex:              aws.Int64(int64(i)),
+			AssociatePublicIpAddress: &netIf.AssociatePublicIPAddress,
```
hardikdr, line 105: Can we please explicitly set `AssociatePublicIpAddress` to false if not specified in the MachineClass ? Unintentionally creating a public-address could have severe security-impact.
- I see, we are using `bool` type, please feel free to convert it to `*bool`. You can then check if it's `nil` (implies not set in MachineClass), simply set it to `false`. 
- This would also help us to not introduce the breaking API-change for Gardener, as it consumes MCM out of box.

pkg/driver/driver_aws.go
```diff
@@ -97,25 +97,32 @@ func (d *AWSDriver) Create() (string, string, error) {
...
+			DeleteOnTermination:      &netIf.DeleteOnTermination,
```
hardikdr, line 106: `DeleteOnTermination` will default to `false` if not specified in the MachineClass, which can potentially leave orphan resources. Can we please set `DeleteOnTermination` to `true` if not specified in the MachineClass ?
- We could also avoid setting this field at all, if not specified in the MachineClass. Please have a look at [ref](https://github.com/gardener/machine-controller-manager/blob/master/pkg/driver/driver_aws.go#L180-L195). This may also apply to `AssociatePublicIpAddress` above.

pkg/driver/driver_aws.go
```diff
@@ -97,25 +97,32 @@ func (d *AWSDriver) Create() (string, string, error) {
 		return "Error", "Error", err
 	}
 
-	securityGroupIDs, err := d.generateSecurityGroups(d.AWSMachineClass.Spec.NetworkInterfaces[0].SecurityGroupIDs)
-	if err != nil {
-		return "Error", "Error", err
+	var networkInterfaceSpecs []*ec2.InstanceNetworkInterfaceSpecification
+	for i, netIf := range d.AWSMachineClass.Spec.NetworkInterfaces {
+		spec := &ec2.InstanceNetworkInterfaceSpecification{
```
hardikdr, line 102: If I understand it correct, the only essential input here is the SecurityGroups and SubnetIDs. Do we really need other fields of `DeviceIndex`, `AssociatePublicIpAddress`, `DeleteOnTermination` and `Description` ? We can simply avoid setting them altogether otherwise.
Introducing these fields with golang-defaults could imply breaking API changes, and possibly unintentional behviour. :) 
- If you see these fields are useful, can we please follow a pattern where we set them only if they are specified in the MachineClass. We will have to convert these fields to a pointer, eg `bool` to `*bool` in that case.
- Please check the pattern followed here for [ref](https://github.com/gardener/machine-controller-manager/blob/master/pkg/driver/driver_aws.go#L180-L195) for DeleteOnTermination.

pkg/driver/driver_aws.go
```diff
@@ -127,16 +134,17 @@ func (d *AWSDriver) Create() (string, string, error) {
...
-}
-
-func (d *AWSDriver) generateSecurityGroups(sgIDs []string) ([]*string, error) {
-
-	var securityGroupIDs []*string
-	for _, sg := range sgIDs {
-		securityGroupIDs = append(securityGroupIDs, aws.String(sg))
+	_, err = svc.ModifyInstanceAttribute(&ec2.ModifyInstanceAttributeInput{
+		InstanceId:      runResult.Instances[0].InstanceId,
+		SourceDestCheck: &ec2.AttributeBooleanValue{Value: d.AWSMachineClass.Spec.SourceDestCheck},
```
hardikdr, line 139: Same comment as above, can we please avoid executing this code-path if `SourceDestCheck` is not set in the MachineClass. 
- It could be, we exeute `ModifyInstanceAttribute` only if the `SourceDeskCheck` is set.

zuzzas: And there we've got some peculiar design decision on the AWS go-sdk team's part. Namely, the `AttributeBooleanValue` thing. To skip e2e testing and possible breaking I've guarded it with the `if type !=nil ` idiom.

pkg/driver/driver_aws.go
zuzzas: >I see, we are using bool type, please feel free to convert it to *bool. You can then check if it's nil (implies not set in MachineClass), simply set it to false.
I've converted the MachineClass' type, but I strongly believe that there is no need for `if type != nil` idiom in this case. You see, if we pass these parameters as-is, then it would exactly the same as not passing them at all since Go's pointer's empty value is `nil`.

pkg/driver/driver_aws.go
zuzzas: Created the default.

pkg/driver/driver_aws.go
zuzzas: The same as https://github.com/gardener/machine-controller-manager/pull/441#discussion_r407408752.


>`DeviceIndex`

It's required and not specified anywhere in the MachineClass. Simply an AWS API kink.

hardikdr: Sure, it should be fine to keep it this way then, thanks for updating the types.

pkg/driver/driver_aws.go
hardikdr: Sure, thanks.

pkg/driver/driver_aws.go
hardikdr: Thanks.

pkg/driver/driver_aws.go
hardikdr: Sure, thanks for the change, it looks good now.

hardikdr:
Thanks for making the changes. 
There is a conflicting file, can you please resolve it.

hardikdr:
>The Create call is not atomic anymore. If MCM crashes/restarts after creating instance, and before modifying the source/destination check attribute, it would never be updated by MCM, since, after PR #432, the created VM will be adopted as it is whenever MCM comes back up

That's a very good point, thanks for pointing out.  
@zuzzas  is it possible to set this field while machine creation itself?

zuzzas:
Unfortunately, some cloud provider's APIs have non-atomic APIs.

Here's what I propose:

1. Remove `source-dest-check` from this PR;
2. Create [an issue](https://github.com/gardener/machine-controller-manager/issues/444) detailing problems with non-atomic APIs.

zuzzas:
@hardikdr the bot is trying to overthrow you.

hardikdr:
>@gardener-ci3 gardener-ci3 added needs/ok-to-test and removed reviewed/ok-to-test labels 

This is expected. 
Bot removes the label, with integration tests starting in parallel. If that was the reason you thought it's overthrowing. :)

PR #442, opened 2020-04-01
Add "auth-extra-groups" field to bootstrap token

zuzzas:

I'd like to access Kubernetes objects in some way or another using the bootstrap token. It's possible to do just that by providing the [`auth-extra-groups`](https://kubernetes.io/docs/reference/access-authn-authz/bootstrap-tokens/#bootstrap-token-secret-format) field (this PR) and creating a Role/RoleBinding. 

The format of the configuration is open to discussion. Perhaps, it would be better to move this to a per-MachineClass Secret?

**Release note**:

```improvement user
Provide a way to specify "auth-extra-groups" field in created bootstrap tokens.
```

### DIFF
--- a/cmd/machine-controller-manager/app/controllermanager.go
+++ b/cmd/machine-controller-manager/app/controllermanager.go
@@ -272,6 +272,7 @@ func StartControllers(s *options.MCMServer,
 			recorder,
 			s.SafetyOptions,
 			s.NodeConditions,
+			s.BootstrapTokenAuthExtraGroups,
 		)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


hardikdr:
I am generally fine either way. Setting it per machine-class secret could be useful to differentiate the auth-groups per machine-classes as well. 
I'd leave it to you to decide the approach :)

zuzzas:
I'd like to leave as it is. It serves us perfectly since we are using MCM without Gardener (for now). If a need arises, it can easily be refactored.

PR #450, opened 2020-04-21
Skip drain on NotReady Nodes

### LINKED ISSUE #448: Avoid draining of NotReady nodes
Avoid draining of nodes in `NotReady` state. 

**Why is this needed**:
Draining of nodes in a `NotReady` state sometimes ends up waiting for up to drain timeout as the Kubelet might be stuck. So in cases where the node isn't ready, a force delete might be preferred.

zuzzas:

Sometimes (mainly due to incorrect requests/limits in our case) a Node's kubelet might become unavailable. Let's skip drain completely on these Nodes.

Fixes #448 

[This](https://dev.to/duske/how-kubernetes-handles-offline-nodes-53b5) article is of use while trying to understand Node NotReady status. I've come to the conclusion that checking current Node Conditions should be enough,

**Release note**:

```operator user
The drain is completely skipped if the node is NotReady for the timeout period. Default timeout is 5 mins. 
```

### DIFF
--- a/pkg/controller/drain.go
+++ b/pkg/controller/drain.go
@@ -58,6 +58,7 @@ type DrainOptions struct {
 	PvDetachTimeout              time.Duration
 	DeleteLocalData              bool
 	nodeName                     string
+	nodeNotReadyDuration         time.Duration
 	Out                          io.Writer
 	ErrOut                       io.Writer
 	Driver                       driver.Driver
@@ -123,6 +124,7 @@ func NewDrainOptions(
 	maxEvictRetries int32,
 	pvDetachTimeout time.Duration,
 	nodeName string,
+	nodeNotReadyDuration time.Duration,
 	gracePeriodSeconds int,
 	forceDeletePods bool,
 	ignorePodsWithoutControllers bool,
@@ -134,6 +136,9 @@ func NewDrainOptions(
 	pvcLister corelisters.PersistentVolumeClaimLister,
 	pvLister corelisters.PersistentVolumeLister,
 ) *DrainOptions {
+	if nodeNotReadyDuration == 0 {
+		nodeNotReadyDuration = 5 * time.Minute
+	}
 
 	return &DrainOptions{
 		client:                       client,
@@ -146,6 +151,7 @@ func NewDrainOptions(
 		PvDetachTimeout:              pvDetachTimeout,
 		DeleteLocalData:              deleteLocalData,
 		nodeName:                     nodeName,
+		nodeNotReadyDuration:         nodeNotReadyDuration,
 		Out:                          out,
 		ErrOut:                       errOut,
 		Driver:                       driver,
@@ -179,7 +185,23 @@ func (o *DrainOptions) RunDrain() error {
 		return err
 	}
 
-	err := o.deleteOrEvictPodsSimple()
+	node, err := o.client.CoreV1().Nodes().Get(o.nodeName, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		klog.V(4).Infof("Node %q not found, skipping drain", o.nodeName)
+		return nil
+	} else if err != nil {
+		klog.Errorf("Error getting details for node: %q. Err: %v", o.nodeName, err)
+		return err
+	}
+
+	for _, condition := range node.Status.Conditions {
+		if condition.Type == corev1.NodeReady && condition.Status != corev1.ConditionTrue && (time.Since(condition.LastTransitionTime.Time) > o.nodeNotReadyDuration) {
+			klog.Warningf("Skipping drain for NotReady Node %q", o.nodeName)
+			return nil
+		}
+	}
+
+	err = o.deleteOrEvictPodsSimple()
 	return err
 }
 
--- a/pkg/controller/machine.go
+++ b/pkg/controller/machine.go
@@ -620,6 +620,7 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
 				maxEvictRetries,
 				pvDetachTimeOut,
 				nodeName,
+				5*time.Minute,
 				-1,
 				forceDeletePods,
 				true,
--- a/pkg/controller/machine_test.go
+++ b/pkg/controller/machine_test.go
@@ -812,6 +812,7 @@ var _ = Describe("machine", func() {
 			machine                 string
 			fakeProviderID          string
 			fakeNodeName            string
+			nodeRecentlyNotReady    *bool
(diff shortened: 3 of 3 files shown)


pkg/controller/drain.go
```diff
@@ -179,7 +179,23 @@ func (o *DrainOptions) RunDrain() error {
...
+	if apierrors.IsNotFound(err) {
+		klog.V(4).Info("Node not found: ", o.nodeName)
+		return err
+	} else if err != nil {
+		klog.Errorf("Error getting details for node: %q. Err: %v", o.nodeName, err)
+		return err
+	}
+
+	for _, condition := range node.Status.Conditions {
+		if condition.Type == corev1.NodeReady && condition.Status != corev1.ConditionTrue {
```
rfranzke, line 192: Does it make sense to enhance this check in order to verify that the node is not ready because the kubelet is not running?

hardikdr: I think the `Ready` node-condition represents if kubelet is posting the status or not.
```
  - lastHeartbeatTime: "2020-04-22T04:29:19Z"
    lastTransitionTime: "2020-04-17T14:13:22Z"
    message: kubelet is posting ready status
    reason: KubeletReady
    status: "True"
    type: Ready
```

Or you hinting towards more advanced node-condition/check. We can add new node-condition via NPD say `KubeletHung` and rely on that for force-deletion.

rfranzke: No, that was already what I was looking for, makes sense. Then ignore my comment/question above. Thanks!

rfranzke:
> Can we overwrite the drain-timeout if NotReady is set during drain.? WDYT?

Not sure, what would you set it to? If the kubelet is down/not responding then the drain timeout will always elapse, I would assume. Does it makes sense then?

hardikdr:
I would have set it to a value equal to health-check timeout[~10mins] or even lesser[~5mins]. 

The idea would be to consider the possibility:
- that kubelet might not actually be completely unhealthy and may come back soon., hence give one chance for pods to evict. 
- we do sometimes see node flipping between ready and not-ready,  may be due to unknown networking issues.

rfranzke:
Hm okay, the second point is valid I guess, thanks for bringing it up. But then even a small timeout might not be good idea, or? What about starting with a very small timeout, ~5min?, and if it elapses reevaluate whether the node meanwhile became ready again? Only if it didn't then terminate it forcefully, otherwise drain again with normal timeout?

hardikdr:
> Hm okay, the second point is valid I guess, thanks for bringing it up. But then even a small timeout might not be good idea, or? What about starting with a very small timeout, ~5min?, and if it elapses reevaluate whether the node meanwhile became ready again? Only if it didn't then terminate it forcefully, otherwise drain again with normal timeout?

That would be a pretty reliable solution.

zuzzas:
Ok, let's try this from this angle. We don't really have to track (time) anything by ourselves since there is a handy `lastTransitionTime` field in the Conditions.

pkg/controller/drain.go
```diff
@@ -179,7 +179,23 @@ func (o *DrainOptions) RunDrain() error {
 		return err
 	}
 
-	err := o.deleteOrEvictPodsSimple()
+	node, err := o.client.CoreV1().Nodes().Get(o.nodeName, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		klog.V(4).Info("Node not found: ", o.nodeName)
+		return err
```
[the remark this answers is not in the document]
zuzzas: An oversight on my part. Thanks, fixed!

pkg/controller/drain.go
```diff
@@ -179,7 +185,23 @@ func (o *DrainOptions) RunDrain() error {
 		return err
 	}
 
-	err := o.deleteOrEvictPodsSimple()
+	node, err := o.client.CoreV1().Nodes().Get(o.nodeName, metav1.GetOptions{})
```
zuzzas, line 188: Another question: does the logic belong here? Perhaps, it should be moved higher up into `machine.go`?

hardikdr: I don't have a strong opinion if this should be here or outside. I would be fine keeping it here.

hardikdr:
>Ok, let's try this from this angle. We don't really have to track (time) anything by ourselves since > there is a handy lastTransitionTime field in the Conditions.

The approach looks good, I'll take a further look tomorrow. Mainly want to check, what happens if the node has been `NotReady` for less than 5 mins, and the drain starts. Would it then wait till complete drain-timeout occurs?

hardikdr:
Drain-timeout seems to be still effective if the drain starts before 5mins since the node is NotReady. I think there would be more changes and complexity introduced if we target the perfect solution discussed above.

Though, the current solution is valuable for most cases when NotReady nodes will be skipped on the drain.
Should we merge this as a short term solution and keep the related issue open? 

WDYS?

zuzzas:
@hardikdr 
One of the most prominent cases, where this issue has been hitting us is when Node has a lot of Pods with badly configured resource requests scheduled on it. If they collectively allocate lots of memory, kernel's memory management stalls, and Node becomes completely stuck without a chance of recovery. It is imperative that MCM swiftly removes and replaces it with a healthy one.

Let's call this PR "a bandaid" which fixes one the uglies problems right now. And merge it, of course. :)

pkg/controller/drain.go
```diff
@@ -179,7 +185,23 @@ func (o *DrainOptions) RunDrain() error {
...
+	} else if err != nil {
+		klog.Errorf("Error getting details for node: %q. Err: %v", o.nodeName, err)
+		return err
+	}
+
+	for _, condition := range node.Status.Conditions {
+		if condition.Type == corev1.NodeReady && condition.Status != corev1.ConditionTrue && (time.Since(condition.LastTransitionTime.Time) > o.nodeNotReadyDuration) {
+			klog.Warningf("Skipping drain for NotReady Node %q", o.nodeName)
+			return nil
+		}
```
hardikdr, line 201: Can we return `err` to get it reconciled again if the machine has been `NotReady` for less than 5 mins?  
- Basically, if node is `NotReady` then check 
     - if more than 5 mins: skip drain
     - if less than 5 mins: return err and let it be reconciled again. If it heals itself then we are lucky else after 5 mins worth of reconciles, the machine drain will be skipped.

WDYT ?

Thanks to @ggaurav10 for the suggestion.

hardikdr: Hm, I'll take this comment back, there seems to be another undesirable side-effect of returning `err` here, especially when the node is flipping between ready/not-ready. 

Pointed out by @ggaurav10

PR #451, opened 2020-04-22
Do not adopt Nodes with empty ProviderID field

zuzzas:

Do not adopt Nodes with an empty ProviderID field.

We've hit a nasty cloud-controller-manager bug: manually created master Node has not been assigned a ProviderID. As a result, the machine-controller-manager adopted the master Node, promptly replacing its metadata and removing it after when Machine has been removed.

**Release note**:

```improvement operator
The mcm stopped adopting Nodes with empty ProviderID field
```


zuzzas:
@hardikdr, requesting a review, since you've contributed the Node adoption mechanism.

hardikdr:
Thanks for the PR @zuzzas .
Looks good to me.

PR #454, opened 2020-04-29
Annotate machine-deployment if machine-types are not available in cloud

hardikdr:
This PR adds the annotation at the nodeSpec of the MachineDeployment, whenever a specific pattern is found in the creation-logs.
It is also mainly motivated due to the resource-scarcity at Azure, where certain machine-types are not available. MCM adds the annotation on such MachineDeployment, which becomes the hint for the autoscaler to keep trying to scale-up it up. 

Related to this: https://github.com/gardener/autoscaler/issues/35

Fixes #

**Release note**:

```improvement operator
MCM adds the annotation `machine.sapcloud.io/machine-type-not-available` whenever certain machine-types are not available in Azure and AWS.
```

### DIFF
--- a/pkg/controller/deployment.go
+++ b/pkg/controller/deployment.go
@@ -302,7 +302,7 @@ func (dc *controller) getMachineDeploymentForMachine(machine *v1alpha1.Machine) 
 		// No controller owns this Machine.
 		return nil
 	}
-	if controllerRef.Kind != "MachineDeployment" { //TODO: Remove hardcoded string
+	if controllerRef.Kind != "MachineSet" { //TODO: Remove hardcoded string
 		// Not a Machine owned by a machine set.
 		return nil
 	}
--- a/pkg/controller/machine.go
+++ b/pkg/controller/machine.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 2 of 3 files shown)


pkg/controller/machine.go
```diff
@@ -428,6 +429,27 @@ func (c *controller) machineCreate(machine *v1alpha1.Machine, driver driver.Driv
...
+				annotations := mdCopy.Spec.Template.Spec.NodeTemplateSpec.Annotations
+				if annotations == nil {
+					annotations = make(map[string]string)
+				}
+				annotations[MachineTypeNotAvailableAnnotation] = "True"
+				mdCopy.Spec.Template.Spec.NodeTemplateSpec.Annotations = annotations
+
+				_, errUpdate := c.controlMachineClient.MachineDeployments(machine.Namespace).Update(mdCopy)
+				if errUpdate != nil {
+					klog.Errorf("Error updating the machine-deployment %+v", errUpdate)
```
hardikdr, line 448: Not returning the error here, to keep the error the intact from the cloud-provider, or should we return the errUpdate ?

pkg/controller/deployment.go
```diff
@@ -302,7 +302,7 @@ func (dc *controller) getMachineDeploymentForMachine(machine *v1alpha1.Machine)
 		// No controller owns this Machine.
 		return nil
 	}
-	if controllerRef.Kind != "MachineDeployment" { //TODO: Remove hardcoded string
+	if controllerRef.Kind != "MachineSet" { //TODO: Remove hardcoded string
```
prashanth26, line 305: I understand the change here. But just wondering if it was not leading to a bug earlier?

pkg/controller/machine.go
```diff
@@ -428,6 +429,27 @@ func (c *controller) machineCreate(machine *v1alpha1.Machine, driver driver.Driv
 			klog.Warning(err)
 		}
 
+		// Check if the creation-error logs contain the specific pattern, and annotate the machine-deployment if found.
+		contains, _ := c.containsSpecialPattern(err.Error(), MachineTypeNotAvailableAzure, MachineTypeNotAvailableAWS)
```
prashanth26, line 433: I think we should move this logic into the machine deployment controller. We shouldn't be updating machine deployments in machine reconciliation. The machine deployment controller can look up for these patterns in failedMachines field?

hardikdr: Hm, that's a valid point.
I had actually started implementing this into the deployment controller. But later moved it here. First of all the log-patterns are extremely cloud-provider specific, may keep changing, and providers may want to add more such patterns in the future. I felt the best place for this logic is to be at provider-specific machine-controllers.

pkg/controller/deployment.go
hardikdr: I think this was a bug earlier, fixed with this.

pkg/controller/machine.go
prashanth26: My concern being that, the machine controller on OOT will not have access to machineDeployments. It will not be able to update this. One alternate approach could be this - https://github.com/gardener/autoscaler/pull/37#discussion_r419854838?

hardikdr:
Update: We hold the development here unless we see the need for improvement again. 

Let's re-open later if required. 
/close

PR #457, opened 2020-04-29
[openstack] Config drive option

zuzzas:

Add an option to use config drive while provisioning OpenStack Servers.

**Release note**:

```improvement operator
Added an option to use configDrive in the OpenStackMachineClass
```

### DIFF
--- a/pkg/driver/driver_openstack.go
+++ b/pkg/driver/driver_openstack.go
@@ -25,10 +25,9 @@ import (
 	"runtime"
 	"strings"
 
-	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
-
 	"github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
 	"github.com/gardener/machine-controller-manager/pkg/metrics"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
 	"github.com/prometheus/client_golang/prometheus"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/klog"
@@ -113,6 +112,7 @@ func (d *OpenStackDriver) Create() (string, string, error) {
 	metadata := d.OpenStackMachineClass.Spec.Tags
 	podNetworkCidr := d.OpenStackMachineClass.Spec.PodNetworkCidr
 	rootDiskSize := d.OpenStackMachineClass.Spec.RootDiskSize
+	useConfigDrive := d.OpenStackMachineClass.Spec.UseConfigDrive
 
 	var createOpts servers.CreateOptsBuilder
 	var imageRef string
@@ -170,6 +170,7 @@ func (d *OpenStackDriver) Create() (string, string, error) {
 		Metadata:         metadata,
 		UserData:         []byte(d.UserData),
 		AvailabilityZone: availabilityZone,
+		ConfigDrive:      useConfigDrive,
 	}
 
 	createOpts = &keypairs.CreateOptsExt{
--- a/kubernetes/crds/machine.sapcloud.io_openstackmachineclasses.yaml
+++ b/kubernetes/crds/machine.sapcloud.io_openstackmachineclasses.yaml
@@ -102,6 +102,8 @@ spec:
               additionalProperties:
                 type: string
               type: object
+            useConfigDrive:
+              type: boolean
           required:
           - availabilityZone
           - flavorName
--- a/pkg/apis/machine/types.go
+++ b/pkg/apis/machine/types.go
@@ -638,6 +638,7 @@ type OpenStackMachineClassSpec struct {
 	SecretRef        *corev1.SecretReference
 	PodNetworkCidr   string
 	RootDiskSize     int // in GB
+	UseConfigDrive   *bool
 }
 
 type OpenStackNetwork struct {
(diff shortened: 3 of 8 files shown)


hardikdr:
Thanks for the PR @zuzzas.

hardikdr:
@afritzler @kayrus @dguendisch Would be really nice if you could take a brief look if possible :)

pkg/driver/driver_openstack.go
```diff
@@ -160,6 +161,29 @@ func (d *OpenStackDriver) Create() (string, string, error) {
 
 	metrics.APIRequestCount.With(prometheus.Labels{"provider": "openstack", "service": "nova"}).Inc()
 
+	var useConfigDrive *bool
```
kayrus, line 164: I'd move the logic below into a separate func. However the whole `Create` func already consists of `if then else`. Apart from the above, it's better to define `var useConfigDrive bool` here and use `ConfigDrive:      &useConfigDrive,` in `createOpts definition.

pkg/driver/driver_openstack.go
```diff
@@ -160,6 +161,29 @@ func (d *OpenStackDriver) Create() (string, string, error) {
...
+
+		allSubnetsForMainNet, err := subnets.ExtractSubnets(allSubnetsPagesForMainNet)
+		if err != nil {
+			return "", "", fmt.Errorf("failed to extract subnets for networkID %q: %s", mainNetworkID, err)
+		}
+
+		for _, subnet := range allSubnetsForMainNet {
+			if !subnet.EnableDHCP {
+				trueVal := true
+				useConfigDrive = &trueVal
```
kayrus, line 182:
```suggestion
				useConfigDrive = true
				break
				
```

pkg/driver/driver_openstack.go
```diff
@@ -160,6 +161,29 @@ func (d *OpenStackDriver) Create() (string, string, error) {
 
 	metrics.APIRequestCount.With(prometheus.Labels{"provider": "openstack", "service": "nova"}).Inc()
 
+	var useConfigDrive *bool
+	if len(serverNetworks) > 0 {
```
kayrus, line 165: Does it make sent to loop over all serverNetworks?
```suggestion
	for _, serverNetwork := range serverNetworks {
```

pkg/driver/driver_openstack.go
```diff
@@ -606,3 +631,13 @@ func resourceInstanceBlockDevicesV2(rootDiskSize int, imageID string) ([]bootfro
 	klog.V(2).Infof("[DEBUG] Block Device Options: %+v", blockDeviceOpts)
 	return blockDeviceOpts, nil
 }
+
+func dhcpDisabled(subnets []subnets.Subnet) bool {
```
kayrus, line 635: I don't see that this func is used.

afritzler:
One other reason for explicit support in the `OpenStackMachineClass`: we had in the past the situation where the `dhcp` interfaces where present in the subnet. However they became unresponsive. We already thought back where it would be good to have the `ConfigDrive` support as an alternative available, but didn't follow through.

zuzzas:
@afritzler 
I've removed all the auto-detection logic and switched to a simple config option. It was nice, but it relied on some assumptions (as pointed by https://github.com/gardener/machine-controller-manager/pull/457#discussion_r418215596) that can clearly break user's expectations.

pkg/driver/driver_openstack.go
zuzzas: Done.

pkg/driver/driver_openstack.go
zuzzas: I've removed all the auto-detection logic due to this. We can't make such assumptions.

pkg/driver/driver_openstack.go
zuzzas: Removed auto-detection.

pkg/driver/driver_openstack.go
```diff
@@ -25,10 +25,9 @@ import (
 	"runtime"
 	"strings"
 
-	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
-
 	"github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
 	"github.com/gardener/machine-controller-manager/pkg/metrics"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
```
zuzzas, line 30: This was done by `goimports`.

hardikdr:
@afritzler It Would be nice if you could take a look again.

hardikdr:
There seems to be a minor file-conflict, can you please take a look @zuzzas .
If there are no more objections, I'll take a final look and prefer to merge it soon.

zuzzas:
@hardikdr, it is done.

hardikdr:
@kayrus @afritzler I would be nice if you could take a look and approve if possible :)

afritzler:
Let me test that against our environment and get back to you by Monday. Pretty swamped atm.

prashanth26:
Hi @afritzler ,

Did you get a chance to test these changes?

hardikdr:
cc @dkistner @MartinWeindel

MartinWeindel:
@zuzzas Can use adjust the PR description to reflect the latest state? There is still described that automatic detection is used, but it's not in the code anymore.

afritzler:
Ok, I finally also came around to test this PR. Creating machines without and with config drive support
```
useConfigDrive: true
```
in the `MachineClass` is working on our OpenStack environment.

/lgtm

zuzzas:
>@zuzzas Can use adjust the PR description to reflect the latest state? There is still described that automatic detection is used, but it's not in the code anymore.

Done.

hardikdr:
Thanks, everyone. I will go ahead merging then.

PR #471, opened 2020-06-09
Generate CRDs for APIs

mvladev:

CRDs are broken on newer K8S clusters.

- CRDs are now generated with `make generate` with `controller-gen` from
`controller-runtime`.
- Removed unneeded `MachineTemplate`.
- Use `k8s.io/api/autoscaling/v1.Scale` instead of home-brew copy.
- Split all versioned APIs into separate `go` files for better code
navigation.

Fixes #

Can you have a look at the generated CRDs, especially at the `required` sections in the OpenAPI schema.

**Release note**:

```improvement operator
CRDs are now generated with full OpenAPI schema.
```


prashanth26:
Hi Martin,

Thank you for this PR as it is very much needed.

PR #475, opened 2020-06-17
fix vm creation when using image id

### LINKED ISSUE #474: [Azure] VMs with images from Shared Image Gallery cannot be created.
**What happened**:
I tried to create a shoot with an image from a shared image gallery, but the VM has never been created. 

In the logs of MCM I see 
```
I0616 10:18:14.073996       1 machine.go:375] Creating machine "shoot--foo--bar-pool-z1-654bf6f885-hxj2h", please wait!
I0616 10:18:14.976829       1 machine.go:403] Adopted an existing VM azure:///westeurope/shoot--foo--bar-pool-z1-654bf6f885-hxj2h for machine object shoot--foo--bar-pool-z1-654bf6f8
85-hxj2h.
I0616 10:18:14.976863       1 machine.go:433] Created/Adopted machine: "shoot--foo--bar-pool-z1-654bf6f885-hxj2h", MachineID: azure:///westeurope/shoot--foo--bar-pool-z1-654bf6f885-
hxj2h
```
However I found no such VM in the azure resource group.

So, I have downgraded MCM to v0.28.0 and it has successfully created the VM.
I have not investigated further, but I suspect that here https://github.com/gardener/machine-controller-manager/blob/fd427655d235631bf41bbb9b97c8b8f54d7af3c4/pkg/driver/driver_azure.go#L519-L526
a regression has been introduced with #463, because URN fields are nil, but there is no check about this.

The VM to be successfully created.

Create a shoot cluster with image from a Shared Image Gallery and MCM version v0.30.0.

/kind bug
/platform azure

**Environment**:

MSSedusch:
fixes regression when using VM image resource id / an image from a shared image gallery

Fixes #474 

**Release note**:

```improvement operator
Azure: Fixes regression when using a VM image resource id instead of an URN (for example an image from a shared image gallery)
```

### DIFF
--- a/pkg/driver/driver_azure.go
+++ b/pkg/driver/driver_azure.go
@@ -106,6 +106,8 @@ func (d *AzureDriver) getVMParameters(vmName string, image *compute.VirtualMachi
 
 	var plan *compute.Plan
 	if image != nil && image.Plan != nil {
+		// If image.Plan exists, create a plan object and attach it to the VM
+		klog.V(2).Infof("Creating a plan object and attaching it to the VM - %q", vmName)
 		plan = &compute.Plan{
 			Name:      image.VirtualMachineImageProperties.Plan.Name,
 			Product:   image.VirtualMachineImageProperties.Plan.Product,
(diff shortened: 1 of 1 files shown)


dkistner:
Will review soon.

PR #478, opened 2020-06-24
Enable controller configurations via MachineDeployment

hardikdr:
**What this PR does / why we need it**:  This enables the controller-level configurations via a field in machine-deployment. 

Partially related to https://github.com/gardener/gardener/issues/1843

**Release note**:

```improvement operator
Enable finer machine configurations per MachineDeployment/MachineSet/Machines basis. Exposed configurations: MachineDrainTimeout, MachinHealthTimeout, MachineCreationTimeout, MaxEvictRetries, NodeConditions. 
```


pkg/controller/deployment_util.go
```diff
@@ -390,6 +390,56 @@ func SetNewMachineSetNodeTemplate(deployment *v1alpha1.MachineDeployment, newIS
...
+			newIS.Annotations[RevisionHistoryAnnotation] = oldRevision
+		} else {
+			oldRevisions = append(oldRevisions, oldRevision)
+			newIS.Annotations[RevisionHistoryAnnotation] = strings.Join(oldRevisions, ",")
+		}
+	}
+	// If the new machine set is about to be created, we need to add replica annotations to it.
+	if !exists && SetReplicasAnnotations(newIS, (deployment.Spec.Replicas), (deployment.Spec.Replicas)+MaxSurge(*deployment)) {
+		configChanged = true
+	}
```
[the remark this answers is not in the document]
hardikdr: I have simply adopted it from the other 2 parallel methods used in the same check: `SetNewMachineSetAnnotations`, `SetNewMachineSetNodeTemplates`, any changes to these 2 methods could otherwise break the flow.

pkg/apis/machine/types.go
```diff
@@ -92,6 +95,27 @@ type MachineTemplateSpec struct {
...
+	MachineHealthTimeout *metav1.Duration
+
+	// MachineCreationTimeout is the timeout after which machinie creation is declared failed.
+	MachineCreationTimeout *metav1.Duration
+
+	// MaxEvictRetries is the number of retries that will be attempted while draining the node.
+	MaxEvictRetries *int32
+
+	// MachineSafetyOrphanVMsPeriod is the time-period after which safety-controller attempts to find the orphan VMs.s
+	MachineSafetyOrphanVMsPeriod *metav1.Duration
```
prashanth26, line 113: This field wouldn't make sense here I think. Because this is purely related to the machine class and not a machine-specific field.

pkg/apis/machine/types.go
```diff
@@ -92,6 +95,27 @@ type MachineTemplateSpec struct {
 	Spec MachineSpec
 }
 
+// MachineControllerConfig describes the configurations useful for the machine-controller.
+type MachineControllerConfig struct {
```
prashanth26, line 99: Is it better that we call this `MachineConfiguration`? Not sure just was thinking of a better name

pkg/controller/deployment_util.go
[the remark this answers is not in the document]
prashanth26: I have tested this flow in the OOT migration. I think a lot of stuff is redundant here as this logic keeps running multiple times. This is some bare minimum that is required i think - https://github.com/prashanth26/machine-controller-manager/commit/5be393e3ea43f59d3b894c97f4389b53fb4472d5#diff-3e28f59dda362e1974d0f3df011e28f1R393. However, if you aren't sure. We can keep this for now, however, we will need to revist it in the future.

pkg/controller/machine.go
```diff
@@ -1036,6 +1042,38 @@ func shouldReconcileMachine(machine *v1alpha1.Machine, now time.Time) bool {
...
+			c.safetyOptions.MachineCreationTimeout = *machine.Spec.MachineControllerConfig.MachineCreationTimeout
+		}
+
+		if machine.Spec.MachineControllerConfig.MaxEvictRetries != nil {
+			c.safetyOptions.MaxEvictRetries = *machine.Spec.MachineControllerConfig.MaxEvictRetries
+		}
+
+		if machine.Spec.MachineControllerConfig.MachineSafetyOrphanVMsPeriod != nil {
+			c.safetyOptions.MachineSafetyOrphanVMsPeriod = *machine.Spec.MachineControllerConfig.MachineSafetyOrphanVMsPeriod
+		}
```
prashanth26, line 1067: Might need to remove this field.

pkg/controller/machine.go
```diff
@@ -161,6 +161,12 @@ func (c *controller) reconcileClusterMachine(machine *v1alpha1.Machine) error {
 		return err
 	}
 
+	// Populate the controller configurations if there are any.
```
prashanth26, line 164: This is setting it as a global flag. So if two machines are being reconciled simultaneously with different values. The values should get mixed up. This can be tested with a simple test. 

I think this values should be fetched during usage and not set here at a global level.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -140,6 +140,12 @@ func (c *controller) reconcileClusterMachine(machine *v1alpha1.Machine) (machine
 		return machineutils.DoNotRetryOp, err
 	}
 
+	// Populate the controller configurations if there are any.
+	err = c.setMachineConfigurations(machine)
```
prashanth26, line 144: Might need to move this as well. 

We should probably use a `getMachineDrainTimeOutMethod()` whose return is populated dynamically as and when required. E.g- in the drain.go

pkg/apis/machine/types.go
hardikdr: Yes, I am fine with either. Let me know if it should be changed to MachineConfiguration.

pkg/apis/machine/types.go
hardikdr: Sure, I shall remove it.

pkg/controller/deployment_util.go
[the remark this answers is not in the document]
hardikdr: Sure, I think there is some redundancy in couple of methods, which can be optimized with a bit of refactoring, but outside the scope of this PR.

pkg/controller/machine.go
hardikdr: Yup, I will remove it.

pkg/apis/machine/types.go
hardikdr: done.

pkg/apis/machine/types.go
prashanth26: Anything is fine with me as well. This should be fine as well.

prashanth26: I see you have already chanaged it to `MachineConfiguration`. Good then.

prashanth26:
There are some errors with the build. You can check the CI pipeline results for the same.

pkg/controller/machine.go
hardikdr: done

pkg/controller/machine.go
hardikdr: Sure, I'll implement the getMachineDrainTimeOut() sort of methods for all configuration, thanks for the catch.

pkg/controller/machine.go
```diff
@@ -1036,8 +1036,91 @@ func shouldReconcileMachine(machine *v1alpha1.Machine, now time.Time) bool {
 	return true
 }
 
+// // setMachineConfigurations checks if there are any machine-controller configurations on the machine-object, and sets if there are any.
```
prashanth26, line 1039: Removed unused/commented block.

pkg/controller/machine.go
```diff
@@ -1036,8 +1036,91 @@ func shouldReconcileMachine(machine *v1alpha1.Machine, now time.Time) bool {
...
+// getEffectiveMaxEvictRetries returns the maxEvictRetries set on the machine-object, otherwise returns the evict retries set using the global-flag.
+func (c *controller) getEffectiveMaxEvictRetries(machine *v1alpha1.Machine) *int32 {
+	var maxEvictRetries *int32
+	if machine.Spec.MachineConfiguration.MaxEvictRetries != nil {
+		maxEvictRetries = machine.Spec.MachineConfiguration.MaxEvictRetries
+	} else {
+		maxEvictRetries = &c.safetyOptions.MaxEvictRetries
+	}
+	return maxEvictRetries
+}
```
prashanth26, line 1093:
```suggestion
func (c *controller) getEffectiveMaxEvictRetries(machine *v1alpha1.Machine) int32 {
	var maxEvictRetries int32
	if machine.Spec.MachineConfiguration.MaxEvictRetries != nil {
		maxEvictRetries = *machine.Spec.MachineConfiguration.MaxEvictRetries
	} else {
		maxEvictRetries = c.safetyOptions.MaxEvictRetries
	}
	return maxEvictRetries
}
```
Change return type to int32 so that it can be directly used without dereferencing?  Only suggesting, feel free to ignore if you feel otherwise.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1058,3 +1058,58 @@ func (c *controller) deleteNodeObject(machine *v1alpha1.Machine) (machineutils.R
...
+// getEffectiveMaxEvictRetries returns the maxEvictRetries set on the machine-object, otherwise returns the evict retries set using the global-flag.
+func (c *controller) getEffectiveMaxEvictRetries(machine *v1alpha1.Machine) *int32 {
+	var maxEvictRetries *int32
+	if machine.Spec.MachineConfiguration.MaxEvictRetries != nil {
+		maxEvictRetries = machine.Spec.MachineConfiguration.MaxEvictRetries
+	} else {
+		maxEvictRetries = &c.safetyOptions.MaxEvictRetries
+	}
+	return maxEvictRetries
+}
```
prashanth26, line 1082:
```suggestion
func (c *controller) getEffectiveMaxEvictRetries(machine *v1alpha1.Machine) int32 {
	var maxEvictRetries int32
	if machine.Spec.MachineConfiguration.MaxEvictRetries != nil {
		maxEvictRetries = *machine.Spec.MachineConfiguration.MaxEvictRetries
	} else {
		maxEvictRetries = c.safetyOptions.MaxEvictRetries
	}
	return maxEvictRetries
}
```
Same here.

pkg/util/provider/machinecontroller/machine.go
hardikdr: done

pkg/controller/machine.go
hardikdr: I'd generally be fine, to keep it consistent with similar methods.

prashanth26: Cool.

pkg/controller/machine.go
hardikdr: Ah, thanks. Yet to take the final look. Btw, can did you make the suggestion/changes, I don't see it, here. I'll anyways remove the commented code soon.

prashanth26: No. That's all it was.

pkg/controller/deployment_util.go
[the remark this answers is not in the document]
ggaurav10: one side effect would be that if all - `SetNewMachineSetAnnotations()`, `SetNewMachineSetNodeTemplate()` and now `SetNewMachineSetConfig()` end up changing something, then we would have triplicates of `revisionHistoryAnnotation` for each revision in the machine deployment, because they all `append` it.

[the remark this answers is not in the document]
hardikdr: Yes, we need to solve this more holistically, by carving the common code/logic out. We should probably open an issue to track this optimization separately.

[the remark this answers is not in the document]
ggaurav10: sure. should be fine

Issue #449, opened 2020-04-21
Provide an option to retain failed machine


aylei:
**What would you like to be added**:
An `--machine-delete-policy` option that support retaining failed machine. Two strategies are proposed:
- `Delete`: current behavior
- `Orphan`: remove failed machine from the machine set (orphan it)

**Why is this needed**:
I run stateful applications on gardener and using local persistent volume, machine deletion is a critical operation because it also delete all data in the local PV. And I've witnessed that all nodes were get replaces when the load balancer of shoot apiserver was unhealthy and lost all local data. 
As a more conservative strategy, the machineset controller could orphan the failed machine from the machineset without actually deleting it.

hardikdr:
Overall sounds like a good idea to me. 
The only part, we might want to consider is, if there are constant failures, such machines could stack up, and lead to unnecessary costs if not seen by the user. We have a `freeze` mechanism in place, where if number of machine-objects goes beyond a certain number MCM stops scaling-up. 
- You might want to complement the above-suggested feature with freeze, at some point we stop creating the new machines. 

Also, out of curiosity, if you are not willing to replace the unhealthy machine, you can simply set a large health-timeout flag, and MCM should not touch it. There could be an improvement to set `-1` to signify infinite timeout. 
- Or is it that, you do want to spawn new machines in the place of unhealthy ones, but also willing to keep the old ones?

hardikdr:
>And I've witnessed that all nodes were get replaces when the load balancer of shoot apiserver was unhealthy and lost all local data.

I think we have not faced it but thought about this case. Basically, kubelets can't reach the APIServer, but all the rest of the control-plane components[kcm, mcm] can reach it. Hence, node-objects go stale, KCM starts evicting and, for MCM machines are unhealthy hence will be replaced. 

@amshuman-kr @ggaurav10  you worked on this case and also implemented the mitigation by disabling the control-plane if the load-balancer is unhealthy, right? 
Can you please elaborate a little around, how can this be avoided?

dguendisch:
Out of curiosity: what is important on that local volume data? I ask as pods managing state typically do this on PVCs. I mean your pod can be evicted and respawn on another node anytime, also with no machine failures, the pods also "loose" the local data in such cases.

aylei:
> * You might want to complement the above-suggested feature with freeze, at some point we stop creating the new machines.

Good call! I've considered something similar and it is great to know we already have this.

[...]

PR #482, opened 2020-06-29
An option to throttle deletion of Failed Machines

zuzzas:

This PR adds an ability to throttle Failed Machine deletion. Why? It's time for a story!

Imagine a Kubernetes cluster, custom-built for a customer. Virtual machines are created via MCM, masters are static. A tragically under-tested infrastructure platform release that severs the routing between Master and Worker Nodes via improperly configured routing tables in a CNI provider.

After losing connection, Machines transfer to the Unknown state, and, after a timeout, into the Failed state, and then MCM deletes them all. All of them at the same time. The recovery time of 80 Nodes in a vSphere environment is abysmal. Disks are slow to detach and attach, the hypervisors become overloaded due to a lot of new machines springing to life.

Moral of the story: if Nodes become detached from Master, the workload would not immediately falter, it's not healthy to remove all Machines from the cluster if they become unavailable. They can still work well, since properly written microservices cache Service Discovery and other information upon startup.

If you agree with the overall approach, I'd like to direct your attention to the new table tests. Please, verify the exhaustiveness of the provided test cases.

**Release note**:

```improvement user
A new option "--failed-machine-deletion-ratio" that configures throttling of Failed Machines deletion process until new Machines transition into Running state.
```

### DIFF
--- a/pkg/controller/machineset.go
+++ b/pkg/controller/machineset.go
@@ -340,13 +340,15 @@ func (c *controller) manageReplicas(allMachines []*v1alpha1.Machine, machineSet 
 	machineDeletionWindow := c.calculateStaleMachineDeletionWindow(activeMachines, runningMachines, staleMachines, machineSet)
 
 	if len(staleMachines) >= 1 {
-		klog.V(2).Infof("Deleting stale machines")
+		klog.V(2).Infof("Deleting %d stale machines out of total stale machines %d", machineDeletionWindow, len(staleMachines))
 	}
-	deletedStaleMachines, _ := c.terminateStaleMachines(staleMachines, machineDeletionWindow, machineSet)
-	staleMachinesLeft := len(staleMachines) - deletedStaleMachines
+	leftOverStaleMachines, err := c.terminateStaleMachines(staleMachines, machineDeletionWindow, machineSet)
+	if err != nil {
+		klog.Errorf("Failed deleting stale machines, operation will be retried in next iteration")
+	}
 
-	diff := len(activeMachines) - int(machineSet.Spec.Replicas)
-
+	assumedNumOfActiveMachines := len(activeMachines) + len(leftOverStaleMachines)
+	diff := assumedNumOfActiveMachines - int(machineSet.Spec.Replicas)
 	klog.V(4).Infof("Difference between current active replicas and desired replicas - %d", diff)
 
 	if diff < 0 {
@@ -357,11 +359,6 @@ func (c *controller) manageReplicas(allMachines []*v1alpha1.Machine, machineSet 
 		}
 
 		diff *= -1
-
-		// clamp machine creation to the count of recently deleted stale machines, so that we don't overshoot
-		if staleMachinesLeft > 0 && diff > deletedStaleMachines {
-			diff = deletedStaleMachines
-		}
 
 		if diff > BurstReplicas {
 			diff = BurstReplicas
@@ -422,6 +419,7 @@ func (c *controller) manageReplicas(allMachines []*v1alpha1.Machine, machineSet 
 		}
 		klog.V(2).Infof("Too many replicas for %v %s/%s, need %d, deleting %d", machineSet.Kind, machineSet.Namespace, machineSet.Name, (machineSet.Spec.Replicas), diff)
 
+		activeMachines = append(leftOverStaleMachines, activeMachines...)
 		machinesToDelete := getMachinesToDelete(activeMachines, diff)
 
 		// Snapshot the UIDs (ns/name) of the machines we're expecting to see
@@ -702,7 +700,8 @@ func (c *controller) terminateMachines(inactiveMachines []*v1alpha1.Machine, mac
 	return nil
 }
 
-func (c *controller) terminateStaleMachines(inactiveMachines []*v1alpha1.Machine, deletionWindow int, machineSet *v1alpha1.MachineSet) (int, error) {
+// terminateStaleMachines terminates the stale machines and returns the lefover machines.
+func (c *controller) terminateStaleMachines(inactiveMachines []*v1alpha1.Machine, deletionWindow int, machineSet *v1alpha1.MachineSet) ([]*v1alpha1.Machine, error) {
 	var (
 		deletedMachines int
 		wg              sync.WaitGroup
@@ -710,22 +709,25 @@ func (c *controller) terminateStaleMachines(inactiveMachines []*v1alpha1.Machine
 	)
 	defer close(errCh)
 
+	leftOverMachines := []*v1alpha1.Machine{}
+
 	wg.Add(deletionWindow)
 	for deletedMachines = 0; deletedMachines < deletionWindow && deletedMachines < len(inactiveMachines); deletedMachines++ {
 		go c.prepareMachineForDeletion(inactiveMachines[deletedMachines], machineSet, &wg, errCh)
 	}
+	leftOverMachines = inactiveMachines[deletionWindow:]
 	wg.Wait()
 
 	select {
 	case err := <-errCh:
 		// all errors have been reported before and they're likely to be the same, so we'll only return the first one we hit.
 		if err != nil {
-			return deletedMachines, err
+			return nil, err
 		}
 	default:
 	}
 
-	return deletedMachines, nil
+	return leftOverMachines, nil
 }
 
 // calculateStaleMachineDeletionWindow computes an amount of Failed Machines that can be deleted based on FailedMachineDeletionRatio
@@ -755,7 +757,7 @@ func (c *controller) calculateStaleMachineDeletionWindow(activeMachines, running
 	if desiredMachineSetReplica == 0 {
 		deletionWindow = numOfInactiveMachines
 	} else {
-		deletionWindow = int(math.Round(float64(desiredMachineSetReplica) * c.safetyOptions.FailedMachineDeletionRatio))
+		deletionWindow = int(math.Round(float64(numOfInactiveMachines) * c.safetyOptions.FailedMachineDeletionRatio))
 		if deletionWindow == 0 {
 			deletionWindow = 1
 		}
--- a/pkg/controller/machineset.go
+++ b/pkg/controller/machineset.go
@@ -712,6 +712,10 @@ func (c *controller) terminateStaleMachines(inactiveMachines []*v1alpha1.Machine
 	)
 	defer close(errCh)
 
+	if deletionWindow > len(inactiveMachines) {
+		deletionWindow = len(inactiveMachines)
+	}
+
 	wg.Add(deletionWindow)
 	for i := 0; i < deletionWindow; i++ {
 		go c.prepareMachineForDeletion(inactiveMachines[i], machineSet, &wg, errCh)
--- a/pkg/controller/machineset.go
+++ b/pkg/controller/machineset.go
@@ -346,9 +346,7 @@ func (c *controller) manageReplicas(allMachines []*v1alpha1.Machine, machineSet 
 	staleMachinesLeft := len(staleMachines) - deletedStaleMachines
 
 	diff := len(activeMachines) - int(machineSet.Spec.Replicas)
-	if staleMachinesLeft > 0 {
-		// Count the leftover stale machines against .spec.replicas to prevent sudden surge in machines which can happen if large number of stale machines are left over
-		diff += staleMachinesLeft
+
 	klog.V(4).Infof("Difference between current active replicas and desired replicas - %d", diff)
 
 	if diff < 0 {
@@ -359,6 +357,11 @@ func (c *controller) manageReplicas(allMachines []*v1alpha1.Machine, machineSet 
 		}
 
 		diff *= -1
+
+		// clamp machine creation to the count of recently deleted stale machines, so that we don't overshoot
+		if staleMachinesLeft > 0 && diff > deletedStaleMachines {
+			diff = deletedStaleMachines
+		}
 
 		if diff > BurstReplicas {
 			diff = BurstReplicas
--- a/pkg/controller/machineset.go
+++ b/pkg/controller/machineset.go
@@ -712,10 +712,6 @@ func (c *controller) terminateStaleMachines(inactiveMachines []*v1alpha1.Machine
 	)
 	defer close(errCh)
 
-	if deletionWindow > len(inactiveMachines) {
-		deletionWindow = len(inactiveMachines)
-	}
-
 	wg.Add(deletionWindow)
 	for i := 0; i < deletionWindow; i++ {
 		go c.prepareMachineForDeletion(inactiveMachines[i], machineSet, &wg, errCh)
@@ -773,6 +769,10 @@ func (c *controller) calculateStaleMachineDeletionWindow(activeMachines, running
 		} else {
 			deletionWindow = 0
 		}
+	}
+
+	if deletionWindow > numOfInactiveMachines {
+		deletionWindow = numOfInactiveMachines
 	}
 
 	return deletionWindow
--- a/pkg/controller/machineset.go
+++ b/pkg/controller/machineset.go
@@ -342,9 +342,21 @@ func (c *controller) manageReplicas(allMachines []*v1alpha1.Machine, machineSet 
 	if len(staleMachines) >= 1 {
 		klog.V(2).Infof("Deleting stale machines")
 	}
-	c.terminateStaleMachines(staleMachines, machineDeletionWindow, machineSet)
+	deletedStaleMachines, _ := c.terminateStaleMachines(staleMachines, machineDeletionWindow, machineSet)
+	staleMachinesLeft := len(staleMachines) - deletedStaleMachines
 
 	diff := len(activeMachines) - int(machineSet.Spec.Replicas)
+	if staleMachinesLeft > 0 {
+		// Count the leftover stale machines against .spec.replicas to prevent sudden surge in machines which can happen if large number of stale machines are left over
+		diff -= staleMachinesLeft
+		if diff < 0 {
+			// Typically, diff >= len(staleMachines)
+			// but this can happen when there are lot of failed machines and scale down happens. eg:
+			// allMachines list is 12. Failed machines = 10. deletionWindow = 3. spec.replicas is changed to 6. So active machines = 12-10 = 2.
+			// Here diff before throttling = -(2-6) = 4. diff after throttling = 4 - (10 - 3) = -3
+			return nil
+		}
+	}
 	klog.V(4).Infof("Difference between current active replicas and desired replicas - %d", diff)
 
 	if diff < 0 {
@@ -358,11 +370,6 @@ func (c *controller) manageReplicas(allMachines []*v1alpha1.Machine, machineSet 
 
 		if diff > BurstReplicas {
 			diff = BurstReplicas
-		}
-
-		// throttle machine creation until all stale machines are deleted
-		if len(staleMachines) > 0 && diff > machineDeletionWindow {
-			diff = machineDeletionWindow
 		}
 
 		if diff == 0 {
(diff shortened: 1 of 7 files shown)


pkg/controller/machineset.go
```diff
@@ -677,6 +678,64 @@ func (c *controller) terminateMachines(inactiveMachines []*v1alpha1.Machine, mac
...
+		return nil
+	}
+
+	for _, activeMachine := range activeMachines {
+		if IsMachineRunning(activeMachine) {
+			numOfRunningMachines++
+		}
+	}
+
+	if numOfRunningMachines == 0 || numOfMachineSetReplicas == 0 || c.safetyOptions.FailedMachineDeletionRatio == 1 || c.safetyOptions.FailedMachineDeletionRatio == 0 {
```
hardikdr, line 703: I was hoping, with `FailedMachineDeletionRatio == 0`, we could simply stop replacing the failed machines.

hardikdr:
I had a first cut review, overall looks good, and thanks for the table based tests. 

I'll try to test this PR mid next week on our environment.

pkg/controller/machineset.go
```diff
@@ -677,6 +678,64 @@ func (c *controller) terminateMachines(inactiveMachines []*v1alpha1.Machine, mac
...
+		if pendingMachines < numOfMachinesToDelete {
+			numOfMachinesToDelete -= pendingMachines
+		} else {
+			klog.Infof("Skipping terminateStaleMachines iteration on MachineSet %q/%q, previous batch is not Running yet", machineSet.Namespace, machineSet.Name)
+			numOfMachinesToDelete = 0
+		}
+	}
+
+	wg.Add(numOfMachinesToDelete)
+	for i := 0; i < numOfMachinesToDelete; i++ {
```
ggaurav10, line 722: a validation `if FailedMachineDeletionRatio <= 1` might be useful to make sure `numOfMachinesToDelete < len(inactiveMachines)`

pkg/controller/machineset.go
```diff
@@ -677,6 +678,64 @@ func (c *controller) terminateMachines(inactiveMachines []*v1alpha1.Machine, mac
...
+	for _, activeMachine := range activeMachines {
+		if IsMachineRunning(activeMachine) {
+			numOfRunningMachines++
+		}
+	}
+
+	if numOfRunningMachines == 0 || numOfMachineSetReplicas == 0 || c.safetyOptions.FailedMachineDeletionRatio == 1 || c.safetyOptions.FailedMachineDeletionRatio == 0 {
+		numOfMachinesToDelete = numOfInactiveMachines
+	} else {
+		numOfMachinesToDelete = int(math.Round(float64(numOfInactiveMachines) * c.safetyOptions.FailedMachineDeletionRatio))
```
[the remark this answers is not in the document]
zuzzas: Done in the previously not used `Validate()` method.

pkg/controller/machineset.go
```diff
@@ -336,7 +337,7 @@ func (c *controller) manageReplicas(allMachines []*v1alpha1.Machine, machineSet
 	if len(staleMachines) >= 1 {
 		klog.V(2).Infof("Deleting stale machines")
 	}
-	c.terminateMachines(staleMachines, machineSet)
+	c.terminateStaleMachines(activeMachines, staleMachines, machineSet)
 
 	diff := len(activeMachines) - int(machineSet.Spec.Replicas)
```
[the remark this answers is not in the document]
zuzzas: That does sound non-obvious and hard to debug. I've changed this ratio to relate to the `.spec.replicas` field of a MachineSet. Would that work?

pkg/controller/machineset.go
zuzzas: Done in the previously not used Validate() method.

pkg/controller/machineset.go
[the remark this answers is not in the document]
hardikdr: Hm, that indeed is non-obvious, though a nice point @ggaurav10, fairly critical. 

I think the catch here is that the `diff` is going to be nullified soon in next step as we are overwriting the `diff` by the `machineDeletionWindow`. 
- That should be fine as our trigger for the creation/deletion of machines should be based on the `numActiveMachine != .Spec.Replicas`.

Though, I am not sure if I have understood the point corrrectly. 
@ggaurav10 can you please confirm/elaborate if the issue you mentioned above still persists.

pkg/controller/machineset.go
```diff
@@ -677,6 +694,75 @@ func (c *controller) terminateMachines(inactiveMachines []*v1alpha1.Machine, mac
...
+func (c *controller) terminateStaleMachines(inactiveMachines []*v1alpha1.Machine, deletionWindow int, machineSet *v1alpha1.MachineSet) error {
+	var (
+		wg    sync.WaitGroup
+		errCh = make(chan error, deletionWindow)
+	)
+	defer close(errCh)
+
+	wg.Add(deletionWindow)
+	for i := 0; i < deletionWindow; i++ {
+		go c.prepareMachineForDeletion(inactiveMachines[i], machineSet, &wg, errCh)
```
hardikdr, line 706: @prashanth26 Could you please check, from your previous fix, if there is a scope of improvement. 
I fear, if only few of the threads fail here, rest are mid way, and later in new reconcilliation new set of stale-machines are chosen for deletion, we may break the contract for ratio.

pkg/controller/machineset.go
```diff
@@ -677,6 +694,75 @@ func (c *controller) terminateMachines(inactiveMachines []*v1alpha1.Machine, mac
...
+		deletionWindow = numOfInactiveMachines
+	} else {
+		deletionWindow = int(math.Round(float64(numOfMachineSetReplicas) * c.safetyOptions.FailedMachineDeletionRatio))
+		if deletionWindow == 0 {
+			deletionWindow = 1
+		}
+
+		pendingMachines := numOfActiveMachines - numOfRunningMachines
+
+		if deletionWindow > pendingMachines {
```
hardikdr, line 756: @zuzzas Can you please explain the reasoning behind comparing with `pendingMachines`, I somehow failed to understand it.

pkg/controller/machineset.go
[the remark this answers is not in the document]
ggaurav10: it doesn't appear to solve the issue of sudden surge in machines in infra.
Earlier when we were at this step - **all** the failed machines (say `x`) were deleted.

After this change some fraction of the failed machine will **not** be deleted (say `x/2`). So the question is - should we account for these undeleted failed machines when creating new machines? If we create machines based on active machines only, we would have `.spec.replicas + x/2` machines after create operation in the infra

pkg/controller/machineset.go
```diff
@@ -677,6 +694,75 @@ func (c *controller) terminateMachines(inactiveMachines []*v1alpha1.Machine, mac
...
+
+	if numOfMachineSetReplicas == 0 {
+		deletionWindow = numOfInactiveMachines
+	} else {
+		deletionWindow = int(math.Round(float64(numOfMachineSetReplicas) * c.safetyOptions.FailedMachineDeletionRatio))
+		if deletionWindow == 0 {
+			deletionWindow = 1
+		}
+
+		pendingMachines := numOfActiveMachines - numOfRunningMachines
```
prashanth26, line 754: numOfActiveMachines (Pending, Available, Running, Unknown) - numOfRunningMachines (Running) = Machines which are either of the following statuses - `Available, Running, Unknown`? So it's not just pending machines, right?

pkg/controller/machineset.go
```diff
@@ -677,6 +694,75 @@ func (c *controller) terminateMachines(inactiveMachines []*v1alpha1.Machine, mac
...
+	}
+
+	if c.safetyOptions.FailedMachineDeletionRatio == 1 {
+		return numOfInactiveMachines
+	}
+
+	if numOfMachineSetReplicas == 0 {
+		deletionWindow = numOfInactiveMachines
+	} else {
+		deletionWindow = int(math.Round(float64(numOfMachineSetReplicas) * c.safetyOptions.FailedMachineDeletionRatio))
```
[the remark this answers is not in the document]
prashanth26: Yes, even i would've expected the following

```suggestion
		deletionWindow = int(math.Round(float64(numOfInactiveMachines) * c.safetyOptions.FailedMachineDeletionRatio))
```

pkg/controller/machineset.go
prashanth26: I am with @hardikdr on this. I am failing to understand the logic here. You could perhaps, help us why you are not taking the calculations with just `inActiveMachines (Failed)` machines? Is there a case you are trying to cover that we have missed. Explanation with an example would be helpful.

pkg/controller/machineset.go
```diff
@@ -677,6 +694,75 @@ func (c *controller) terminateMachines(inactiveMachines []*v1alpha1.Machine, mac
...
+		if err != nil {
+			return err
+		}
+	default:
+	}
+
+	return nil
+}
+
+// calculateStaleMachineDeletionWindow computes
```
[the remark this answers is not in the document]
prashanth26: Would help if you could add some comments on what exactly this method is trying to achieve.

pkg/controller/machineset.go
zuzzas: The provided ration will now be taken into account if it equates to `0`.

pkg/controller/machineset.go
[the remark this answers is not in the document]
zuzzas: Are we okay with progressively decaying `deletionWindow`? Since `inactiveMachines` list will progressively shrink, we'll delete fewer and fewer Failed Machines with each iteration.

pkg/controller/machineset.go
zuzzas: `pendingMachines` is a horrible variable name, since it throws everyone off due to collision with the `Pending` Machines state. I've renamed it into a longer, but a proper name `numOfActiveNotRunningMachines`.

`Pending`, `Unknown`, `Terminating` are transient Machine states. Machines transition through them into more concrete `Running` or `Failed` states. I believe that we should consider all Machines in these transient states when we calculate `deletionWindow`. This way we won't delete Failed Machines until Machines in the aforementioned transient states stabilize into `Running` or `Failed` state.

Btw, `Available` state is not used anywhere in the codebase (might be a candidate for deletion?).

pkg/controller/machineset.go
zuzzas: I've expanded on the issue in [another thread](https://github.com/gardener/machine-controller-manager/pull/482#discussion_r468195769). Let's discuss the naming and the algorithm there.

pkg/controller/machineset.go
```diff
@@ -677,6 +694,75 @@ func (c *controller) terminateMachines(inactiveMachines []*v1alpha1.Machine, mac
...
+func (c *controller) calculateStaleMachineDeletionWindow(activeMachines, runningMachines, inactiveMachines []*v1alpha1.Machine, machineSet *v1alpha1.MachineSet) int {
+	var (
+		deletionWindow          int
+		numOfActiveMachines     = len(activeMachines)
+		numOfRunningMachines    = len(runningMachines)
+		numOfInactiveMachines   = len(inactiveMachines)
+		numOfMachineSetReplicas = int(machineSet.Spec.Replicas)
```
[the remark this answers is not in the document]
zuzzas: Makes sense. Done!

pkg/controller/machineset.go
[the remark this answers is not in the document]
zuzzas: Done! Let me know if a comment should be expanded.

pkg/controller/machineset.go
```diff
@@ -677,6 +678,64 @@ func (c *controller) terminateMachines(inactiveMachines []*v1alpha1.Machine, mac
...
+		}
+
+		pendingMachines := numOfActiveMachines - numOfRunningMachines
+
+		if pendingMachines < numOfMachinesToDelete {
+			numOfMachinesToDelete -= pendingMachines
+		} else {
+			klog.Infof("Skipping terminateStaleMachines iteration on MachineSet %q/%q, previous batch is not Running yet", machineSet.Namespace, machineSet.Name)
+			numOfMachinesToDelete = 0
+		}
```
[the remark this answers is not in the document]
zuzzas: A horrible variable name that leads to confusion. Let's move all the discussion of the naming and the algorithm into [this thread](https://github.com/gardener/machine-controller-manager/pull/482#discussion_r468195769), in which I've elaborated on these issues.

pkg/controller/machineset.go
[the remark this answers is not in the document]
zuzzas: @ggaurav10 
I've taken this into account when I've been changing the way ratio works. Doesn't [this](https://github.com/gardener/machine-controller-manager/pull/482/files#diff-16f7b87f40829e0c81cb6c486f3997ccR363-R370) solve the issue?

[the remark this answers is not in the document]
ggaurav10: ah yes. that throttling should solve the issue. 👍

pkg/controller/machineset.go
```diff
@@ -683,6 +700,75 @@ func (c *controller) terminateMachines(inactiveMachines []*v1alpha1.Machine, mac
...
+	}
+
+	if c.safetyOptions.FailedMachineDeletionRatio == 1 {
+		return numOfInactiveMachines
+	}
+
+	if desiredMachineSetReplica == 0 {
+		deletionWindow = numOfInactiveMachines
+	} else {
+		deletionWindow = int(math.Round(float64(desiredMachineSetReplica) * c.safetyOptions.FailedMachineDeletionRatio))
```
ggaurav10, line 755: could this be more than `numOfInactiveMachines`? i think the code will panic in `terminateStaleMachines()` if that happens. Can you please verify. thanks.

zuzzas: Nice catch. I've [clamped](https://github.com/gardener/machine-controller-manager/pull/482/files#diff-16f7b87f40829e0c81cb6c486f3997ccR710-R712) it down to the deletion window.

ggaurav10: should we also adjust the same when throttling machine creation in `manageReplicas()`?

zuzzas: After some deliberation, I fully agree. There is no reason to burst Machine creation if there is no need. Pushed the change.

pkg/controller/machineset.go
```diff
@@ -349,9 +355,20 @@ func (c *controller) manageReplicas(allMachines []*v1alpha1.Machine, machineSet
...
 		diff *= -1
+
 		if diff > BurstReplicas {
 			diff = BurstReplicas
 		}
+
+		// throttle machine creation until all stale machines are deleted
+		if len(staleMachines) > 0 && diff > machineDeletionWindow {
+			diff = machineDeletionWindow
+		}
```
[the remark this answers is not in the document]
zuzzas: Thanks. This and other suggestions should be encoded into unit tests. I'll get right to it.

pkg/controller/machineset.go
prashanth26: I think failing here should be fine. Considering that we are using waitgroups anyways for all the threads to complete right? WDYT?

hardikdr:
@ggaurav10 @prashanth26 Can you please resolve the comments, if you think they are taken care of, else it would great if you could follow-up :)

pkg/controller/machineset.go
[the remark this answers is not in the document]
prashanth26: Okay sure @zuzzas . Makes sense to me now.

pkg/controller/machineset.go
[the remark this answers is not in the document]
hardikdr: Agreed, we should recompute the `diff` , with `diff = diff - (len(staleMachines) - len(machineDeletionWindow))` to avoid throttled machine-creations. 
This is clearly a bit mind-boggling, thanks for @ggaurav10 for catching and @zuzzas for implementing.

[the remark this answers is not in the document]
zuzzas: It is done.
https://github.com/gardener/machine-controller-manager/pull/482/files#diff-16f7b87f40829e0c81cb6c486f3997ccR345-R359

zuzzas (commented):
I'd like to direct the attention of reviewers to a set of test cases. If I have forgotten a case, please, remind me of it. I've refactored this part into a table form.

@ggaurav10 @prashanth26 @hardikdr

pkg/controller/machineset.go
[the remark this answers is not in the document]
hardikdr: thanks. @ggaurav10 can you please take a quick look at this aspect.

hardikdr:
>I'd like to direct the attention of reviewers to a set of test cases. If I have forgotten a case, please, remind me of it. I've refactored this part into a table form.

Thanks for moving to table format, looks sufficient to me.

pkg/controller/machineset.go
```diff
@@ -323,22 +324,39 @@ func (c *controller) manageReplicas(allMachines []*v1alpha1.Machine, machineSet
...
-	c.terminateMachines(staleMachines, machineSet)
+	deletedStaleMachines, _ := c.terminateStaleMachines(staleMachines, machineDeletionWindow, machineSet)
+	staleMachinesLeft := len(staleMachines) - deletedStaleMachines
 
 	diff := len(activeMachines) - int(machineSet.Spec.Replicas)
+	if staleMachinesLeft > 0 {
+		// Count the leftover stale machines against .spec.replicas to prevent sudden surge in machines which can happen if large number of stale machines are left over
+		diff -= staleMachinesLeft
+		if diff < 0 {
+			// Typically, diff >= len(staleMachines)
```
hardikdr, line 353: This is a good catch, thanks.

pkg/controller/machineset.go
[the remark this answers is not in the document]
hardikdr: I see, now we are recomputing the diff at the start itself, and the new diff[which considers the left out stale machines]. At the first sight, it looks good, I'd like to test this aspect separately.

ggaurav10 (changes requested):

pkg/controller/machineset.go
```diff
@@ -323,22 +324,39 @@ func (c *controller) manageReplicas(allMachines []*v1alpha1.Machine, machineSet
...
 		klog.V(2).Infof("Deleting stale machines")
 	}
-	c.terminateMachines(staleMachines, machineSet)
+	deletedStaleMachines, _ := c.terminateStaleMachines(staleMachines, machineDeletionWindow, machineSet)
+	staleMachinesLeft := len(staleMachines) - deletedStaleMachines
 
 	diff := len(activeMachines) - int(machineSet.Spec.Replicas)
+	if staleMachinesLeft > 0 {
+		// Count the leftover stale machines against .spec.replicas to prevent sudden surge in machines which can happen if large number of stale machines are left over
+		diff -= staleMachinesLeft
```
hardikdr, line 351: If `diff` is already negative, the further substraction will increase the number. This step should either be below in the if block or some other alternative.
Simple example from Gaurav: 
```
active machines = 12, spec = 15, deleted 2, left over = 1,
we should now be creating only 2 machines
….and not 4
```
@zuzzas Can you please take care of this part?
credits @ggaurav10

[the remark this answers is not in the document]
zuzzas: But that won't work, because it'll panic in `getMachinesToDelete()` since diff will be bigger than activeMachines count.

It's my fault for moving this calculation outside of the `if` branch. [This](https://github.com/gardener/machine-controller-manager/pull/482/files#diff-16f7b87f40829e0c81cb6c486f3997ccR361-R364) condition should suffice.

zuzzas: Taken care of.

hardikdr:
@aylei Would you want to take a brief look as well, this seems to be related to what you proposed here: https://github.com/gardener/machine-controller-manager/issues/449 . - sorry for the late invite.

@zuzzas Did you finally get any experience from running these changes in the production?

pkg/controller/machineset.go
hardikdr: Resolving this, as code is changed now.

pkg/controller/machineset.go
```diff
@@ -349,6 +357,12 @@ func (c *controller) manageReplicas(allMachines []*v1alpha1.Machine, machineSet
 		}
 
 		diff *= -1
+
+		// clamp machine creation to the count of recently deleted stale machines, so that we don't overshoot
+		if staleMachinesLeft > 0 && diff > deletedStaleMachines {
+			diff = deletedStaleMachines
```
hardikdr, line 363: Hm, this seems to be breaking the scale-up situation.
Eg. active machines are 10, desired is 13, and stale machines are 4, and thresold: 50% .
Then I would want to create 2+3=5 machines, and not 2.

hardikdr:
I have opened this with the help from @ggaurav10  : https://github.com/flant/machine-controller-manager/pull/1 , I have tried to convey the core idea.

hardikdr:
Trying to put down the overall approach for the feature, from what we learned so far,

- A flag should be introduced to determine the expected portion of the `Failed` machines be removed.
   - In fact, even better if it's configurable via `MachineDeployment.Spec.FailedDeletionRatio`. Similar to what we have for `healthTimeout`, see: https://github.com/gardener/machine-controller-manager/pull/478 . This allows us to set it per MD.
- The Ratio should be calculated based on the current number of `inactiveMachines` and not `Spec.Replicas`. The calculation should be re-done with every reconciliation of machine-set as the number of `Failed` machines could change with time.
- There should be a lower threshold, after which we simply delete all of them, and not re-calculate based on ratio - eg last 3 machines should be simply deleted altogether.
- The next iteration of the deletion of the failed machines should wait till the previous bunch has joined the cluster. 
- Overall the logic should also work with ongoing scale-up or scale-down scenarios, eg machine-deployment is scaled-up/down while few machines are already being `Failed`.

Most of the above requirements have already been taken care of in the PR. 
@zuzzas Wdyt , does the description above works for you ?

PR #487, opened 2020-07-13
Bugfix/finalizers

### LINKED ISSUE #486: Secret without finalizers
**What happened**:
The secret has already been deleted for a shoot cluster in deletion.

MCM Logs below
```
I0624 05:40:39.914874       1 secret.go:132] Successfully added/removed finalizer on the secret "shoot--garden--vsphere-scp-cpu-worker-z1-a3efd"
I0624 05:50:39.966983       1 secret.go:132] Successfully added/removed finalizer on the secret "shoot--garden--vsphere-scp-worker2-z1-148c1"
I0624 05:50:40.162211       1 secret.go:132] Successfully added/removed finalizer on the secret "shoot--garden--vsphere-scp-cpu-worker-z1-a3efd"
I0624 06:00:40.415868       1 secret.go:132] Successfully added/removed finalizer on the secret "shoot--garden--vsphere-scp-worker2-z1-148c1"
I0624 06:10:40.571230       1 secret.go:132] Successfully added/removed finalizer on the secret "shoot--garden--vsphere-scp-worker2-z1-148c1"
I0624 06:20:40.871823       1 secret.go:132] Successfully added/removed finalizer on the secret "shoot--garden--vsphere-scp-worker2-z1-148c1"
```

When you trigger shoot deletion for a shoot with OOT sometimes.

This is caused by a race between the in-tree and out-of-tree controllers which put the same finalizer on the secret object. We need different finalizers to fix this issue.

**Environment**:

prashanth26:
Fixes bugfixes on OOT while trying to delete finalizers on secrets

Fixes #486 

**Release note**:

```improvement operator
Avoids race between secret finalizer creation/deletion between in-tree & OOT controllers
```

### DIFF
--- a/pkg/util/provider/machinecontroller/controller.go
+++ b/pkg/util/provider/machinecontroller/controller.go
@@ -51,8 +51,9 @@ const (
 	ClassAnnotation = "machine.sapcloud.io/class"
 	// MachineIDAnnotation is the annotation used to identify a machine ID
 	MachineIDAnnotation = "machine.sapcloud.io/id"
-	// DeleteFinalizerName is the finalizer used to identify the controller acting on an object
-	DeleteFinalizerName = "machine.sapcloud.io/machine-controller-manager"
+	// MCMFinalizerName is the finalizer used to tag dependecies before deletion
+	// of the object. This finalizer is carried over from then MCM
+	MCMFinalizerName = "machine.sapcloud.io/machine-controller-manager"
 )
 
 // NewController returns a new Node controller.
--- a/pkg/util/provider/machinecontroller/controller.go
+++ b/pkg/util/provider/machinecontroller/controller.go
@@ -52,8 +52,13 @@ const (
 	// MachineIDAnnotation is the annotation used to identify a machine ID
 	MachineIDAnnotation = "machine.sapcloud.io/id"
 	// MCMFinalizerName is the finalizer used to tag dependecies before deletion
-	// of the object. This finalizer is carried over from then MCM
+	// of the object. This finalizer is carried over from the MCM
 	MCMFinalizerName = "machine.sapcloud.io/machine-controller-manager"
+	// MCFinalizerName is the finalizer created for the external
+	// machine controller to differentiate it from the MCMFinalizerName
+	// This finalizer is added only on secret-objects to avoid race between in-tree and out-of-tree controllers.
+	// This is a stopgap solution to resolve: https://github.com/gardener/machine-controller-manager/issues/486. 
+	MCFinalizerName = "machine.sapcloud.io/machine-controller"
 )
 
 // NewController returns a new Node controller.
--- a/pkg/util/provider/machinecontroller/secret.go
+++ b/pkg/util/provider/machinecontroller/secret.go
@@ -76,7 +76,7 @@ func (c *controller) reconcileClusterSecret(secret *corev1.Secret) error {
 			return err
 		}
 	} else {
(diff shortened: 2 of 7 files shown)


pkg/util/provider/machinecontroller/secret.go
```diff
@@ -76,7 +76,7 @@ func (c *controller) reconcileClusterSecret(secret *corev1.Secret) error {
 			return err
 		}
 	} else {
-		if finalizers := sets.NewString(secret.Finalizers...); !finalizers.Has(DeleteFinalizerName) {
+		if finalizers := sets.NewString(secret.Finalizers...); !finalizers.Has(MCFinalizerName) {
```
hardikdr, line 79: If I understand the solution, only `Secrets` get a machine-controller finalizer, and the rest of the objects gets machine-controller-manager finalizer. This could be little confusing. But I also understand the race condition, that both intree and OOT controllers are interested in setting/unsetting finalizers on the secret objects. 

wdys?

Overall, I suggest we consider this solution to be a stopgap, and once we get rid of the in-tree controllers[it doesn't need to worry about the provider-specific machine-classes], we align the finalizers to be the same.

pkg/util/provider/machinecontroller/controller.go
```diff
@@ -51,8 +51,14 @@ const (
...
-	// DeleteFinalizerName is the finalizer used to identify the controller acting on an object
-	DeleteFinalizerName = "machine.sapcloud.io/machine-controller-manager"
+	// MCMFinalizerName is the finalizer used to tag dependecies before deletion
+	// of the object. This finalizer is carried over from then MCM
+	MCMFinalizerName = "machine.sapcloud.io/machine-controller-manager"
+	// MCFinalizerName is the new finalizer created for the external
+	// machine controller to seperate it from the older MCMFinalizerName
+	// This is useful add finalizers to secrets where in-tree controller might
+	// still be looking for mappings and acting on it.
+	MCFinalizerName = "machine.sapcloud.io/machine-controller"
```
hardikdr, line 61:
```suggestion
	// of the object. This finalizer is carried over from the MCM
	MCMFinalizerName = "machine.sapcloud.io/machine-controller-manager"
	// MCFinalizerName is the finalizer created for the external
	// machine controller to differentiate it from the MCMFinalizerName
	// This finalizer is added only on secret-objects to avoid race between in-tree and out-of-tree controllers.
	// This is a stopgap solution to resolve: https://github.com/gardener/machine-controller-manager/issues/486. 
	MCFinalizerName = "machine.sapcloud.io/machine-controller"
```

hardikdr: Please accept the suggestion only we agree on the overall review.

hardikdr:
Another solution could be to dedicate all the objects managed by external-machine-controller to have `machine-controller` finalized and rest gets MCMFinalizer. But of course only in the future, when we don't have in-tree machine-controllers.

pkg/util/provider/machinecontroller/secret.go
prashanth26: Yes agreed. Right now, it's a little tricky to migrate the existing finalizers on machine objects with the migration logic associated with it. But yes, in future let's stick to a consistent finalizer solution and this would be an intermediate solution until that.

prashanth26:
> Another solution could be to dedicate all the objects managed by external-machine-controller to have machine-controller finalized and rest gets MCMFinalizer. But of course only in the future, when we don't have in-tree machine-controllers.

The problem with is the migration logic is a bit tricky here. I would suggest we do it in another step.

PR #492, opened 2020-07-21
Add termination condition to deleting nodes

### LINKED ISSUE #488: MCM to mark terminating nodes

When MCM replaces nodes (unhealthy node termination, node rolling updates, ...), nodes that are being terminated should be marked in some way (e.g. annotation) to reflect the reason of their termination. 

**Why is this needed**:
End-users can setup monitoring for the progress of node rolling updates. If PDBs are used, end-users can automate removal of PDBs from pods that are running on nodes which are being terminated due to health issues.

guydaichs:
Helps users monitor and react to ongoing node terminations.

Fixes #488

Assumes that termination is final in MCM, and a terminating node cannot be restored. So, only situations where the termination condition is true are handled. 

Note that some RBAC changes are required to allow MCM updates of the node's status. 

**Release note**:
```noteworthy operator
RBAC policies have to be updated to allow updating of `node/status` resources. 
```
```improvement user
Node condition is added to the status of terminating nodes indicating the termination start time and reason (Unhealthy|ScaleDown)
```

### DIFF
--- a/pkg/controller/controller_utils.go
+++ b/pkg/controller/controller_utils.go
@@ -37,6 +37,7 @@ import (
 
 	"github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
 	machineapi "github.com/gardener/machine-controller-manager/pkg/client/clientset/versioned/typed/machine/v1alpha1"
+	conditionutils "github.com/gardener/machine-controller-manager/pkg/util/conditions"
 	hashutil "github.com/gardener/machine-controller-manager/pkg/util/hash"
 	taintutils "github.com/gardener/machine-controller-manager/pkg/util/taints"
 	v1 "k8s.io/api/core/v1"
@@ -88,8 +89,8 @@ const (
 	SlowStartInitialBatchSize = 1
 )
 
-// UpdateTaintBackoff is the backoff period used while updating taint
-var UpdateTaintBackoff = wait.Backoff{
+// Backoff is the backoff period used while updating nodes
+var Backoff = wait.Backoff{
 	Steps:    5,
 	Duration: 100 * time.Millisecond,
 	Jitter:   1.0,
@@ -901,7 +902,7 @@ func AddOrUpdateTaintOnNode(c clientset.Interface, nodeName string, taints ...*v
 		return nil
 	}
 	firstTry := true
-	return clientretry.RetryOnConflict(UpdateTaintBackoff, func() error {
+	return clientretry.RetryOnConflict(Backoff, func() error {
 		var err error
 		var oldNode *v1.Node
 		// First we try getting node from the API server cache, as it's cheaper. If it fails
@@ -958,7 +959,7 @@ func RemoveTaintOffNode(c clientset.Interface, nodeName string, node *v1.Node, t
 	}
 
 	firstTry := true
-	return clientretry.RetryOnConflict(UpdateTaintBackoff, func() error {
+	return clientretry.RetryOnConflict(Backoff, func() error {
 		var err error
 		var oldNode *v1.Node
 		// First we try getting node from the API server cache, as it's cheaper. If it fails
@@ -1027,6 +1028,55 @@ func UpdateNodeTaints(c clientset.Interface, nodeName string, oldNode *v1.Node, 
 	_, err := c.CoreV1().Nodes().Update(newNodeClone)
 	if err != nil {
 		return fmt.Errorf("failed to create update taints for node %q: %v", nodeName, err)
+	}
+
+	return nil
+}
+
+// GetNodeCondition get the nodes condition matching the specified type
+func GetNodeCondition(c clientset.Interface, nodeName string, conditionType v1.NodeConditionType) (*v1.NodeCondition, error) {
+	node, err := c.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return conditionutils.GetNodeCondition(node, conditionType), nil
+}
+
+// AddOrUpdateConditionsOnNode adds a condition to the node's status
+func AddOrUpdateConditionsOnNode(c clientset.Interface, nodeName string, condition v1.NodeCondition) error {
+	firstTry := true
+	return clientretry.RetryOnConflict(Backoff, func() error {
+		var err error
+		var oldNode *v1.Node
+		// First we try getting node from the API server cache, as it's cheaper. If it fails
+		// we get it from etcd to be sure to have fresh data.
+		if firstTry {
+			oldNode, err = c.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{ResourceVersion: "0"})
+			firstTry = false
+		} else {
+			oldNode, err = c.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{})
+		}
+
+		if err != nil {
+			return err
+		}
+
+		var newNode *v1.Node
+		oldNodeCopy := oldNode
+		newNode = conditionutils.AddOrUpdateCondition(oldNodeCopy, condition)
+		return UpdateNodeConditions(c, nodeName, oldNode, newNode)
+	})
+}
+
+// UpdateNodeConditions is for updating the node conditions from oldNode to the newNode
+// using the nodes Update() method
+func UpdateNodeConditions(c clientset.Interface, nodeName string, oldNode *v1.Node, newNode *v1.Node) error {
+	newNodeClone := oldNode.DeepCopy()
+	newNodeClone.Status.Conditions = newNode.Status.Conditions
+
+	_, err := c.CoreV1().Nodes().UpdateStatus(newNodeClone)
+	if err != nil {
+		return fmt.Errorf("failed to create update conditions for node %q: %v", nodeName, err)
 	}
 
 	return nil
--- a/pkg/controller/machine.go
+++ b/pkg/controller/machine.go
@@ -565,9 +565,51 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
 
 	if finalizers := sets.NewString(machine.Finalizers...); finalizers.Has(DeleteFinalizerName) {
 		klog.V(2).Infof("Deleting Machine %q", machine.Name)
+		var (
+			forceDeletePods         = false
+			forceDeleteMachine      = false
+			timeOutOccurred         = false
+			maxEvictRetries         = int32(math.Min(float64(*c.getEffectiveMaxEvictRetries(machine)), c.getEffectiveDrainTimeout(machine).Seconds()/PodEvictionRetryInterval.Seconds()))
+			pvDetachTimeOut         = c.safetyOptions.PvDetachTimeout.Duration
+			timeOutDuration         = c.getEffectiveDrainTimeout(machine).Duration
+			forceDeleteLabelPresent = machine.Labels["force-deletion"] == "True"
+		)
+
+		// Timeout value obtained by subtracting last operation with expected time out period
+		timeOut := metav1.Now().Add(-timeOutDuration).Sub(machine.Status.CurrentStatus.LastUpdateTime.Time)
+		timeOutOccurred = timeOut > 0
+
+		if forceDeleteLabelPresent || timeOutOccurred {
+			// To perform forceful machine drain/delete either one of the below conditions must be satified
+			// 1. force-deletion: "True" label must be present
+			// 2. Deletion operation is more than drain-timeout minutes old
+			forceDeleteMachine = true
+			forceDeletePods = true
+			timeOutDuration = 1 * time.Minute
+			maxEvictRetries = 1
+
+			klog.V(2).Infof(
+				"Force deletion has been triggerred for machine %q due to ForceDeletionLabel:%t, Timeout:%t",
+				machine.Name,
+				forceDeleteLabelPresent,
+				timeOutOccurred,
+			)
+		}
 
 		// If machine was created on the cloud provider
 		machineID, _ := driver.GetExisting()
+
+		// update node with the machine's state prior to termination
+		if nodeName != "" && machineID != "" {
+			if err = c.UpdateNodeTerminationCondition(machine); err != nil {
+				if forceDeleteMachine {
+					klog.Warningf("failed to update node conditions: %v. However, since it's a force deletion shall continue deletion of VM.", err)
+				} else {
+					klog.Error(err)
+					return err
+				}
+			}
+		}
 
 		if machine.Status.CurrentStatus.Phase != v1alpha1.MachineTerminating {
 			lastOperation := v1alpha1.LastOperation{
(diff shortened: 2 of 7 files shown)


pkg/controller/machine_util_test.go
```diff
@@ -1850,4 +1850,207 @@ var _ = Describe("machine_util", func() {
 		)
 
 	})
+
+	FDescribe("#AnnotateTerminatingMachineNode", func() {
```
prashanth26, line 1854:
```suggestion
	Describe("#AnnotateTerminatingMachineNode", func() {
```

guydaichs: sorry, missed that. fixed now.

hardikdr:
Thanks @guydaichs for the great PR, and sorry for the delay in review :)

pkg/controller/machine.go
```diff
@@ -569,6 +569,9 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
 		// If machine was created on the cloud provider
 		machineID, _ := driver.GetExisting()
 
+		// update node with the machine's state prior to termination
+		c.AnnotateTerminatingMachineNode(machine)
```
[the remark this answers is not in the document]
guydaichs: > I guess, should be fine to ignore the error here, but then it becomes the best-effort feature, and not guruanteed. I would prefer to understand the consequences on both side.

I was concerned about adding something to the critical path of node deprovisioning. Since you approve, I will return an error here, to ensure that this is retried and the annotations is added (if the node exists). 

> Also, I'd be curious to know your plans of making use of this feature in general.

We run some single-replica deployments in our cluster. The associated pods must not be disrupted during cluster node upgrades, at least until some timeout is reached. The MCM drain performs voluntary evictions, so we use PDBs with `maxUnavailable: 0` to enforce this behavior. 

However, the same draining behavior would apply when replacing unhealthy nodes. We want to identify this situation, and automate the removal of PDBs from unhealthy nodes that are being deprovisioned, to ensure that they are removed promptly.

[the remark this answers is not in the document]
hardikdr: Sure, sounds good.

hardikdr:
Thanks @guydaichs for the changes, I'd be testing this today quickly and would like to go ahead merging.
@prashanth26 @ggaurav10  Can you please take a final look if possible.

pkg/controller/machine_util.go
```diff
@@ -496,3 +500,21 @@ func SyncMachineTaints(
...
 	return toBeUpdated
 }
+
+func (c *controller) AnnotateTerminatingMachineNode(machine *v1alpha1.Machine) error {
+	if machine.Status.CurrentStatus.Phase == "" || machine.Status.CurrentStatus.Phase == v1alpha1.MachineTerminating {
+		return nil
+	}
+
+	nodeName := machine.Status.Node
+	if machine.Status.CurrentStatus.Phase == "" {
```
hardikdr, line 510: Is there any specific reason to have a same check twice ?

pkg/controller/machine.go
```diff
@@ -565,10 +565,50 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
...
+				timeOutOccurred,
+			)
+		}
 
 		// If machine was created on the cloud provider
 		machineID, _ := driver.GetExisting()
 
+		// update node with the machine's state prior to termination
+		if err = c.AnnotateTerminatingMachineNode(machine); err != nil {
+			if forceDeleteMachine {
```
hardikdr, line 604: Good catch, thanks.

hardikdr: I think we should add `nodeName != ""` check here to avoid making an extra update call.

hardikdr:
I could test it quickly and works pretty well.

guydaichs:
Thanks @hardikdr, @ggaurav10  for the comments, i'll try to fix these issues soon.

pkg/controller/controller_utils.go
```diff
@@ -1032,6 +1033,52 @@ func UpdateNodeTaints(c clientset.Interface, nodeName string, oldNode *v1.Node,
...
+}
+
+// UpdateNodeAnnotations is for updating the node annotations from oldNode to the newNode
+// using the nodes Update() method
+func UpdateNodeAnnotations(c clientset.Interface, nodeName string, oldNode *v1.Node, newNode *v1.Node) error {
+	newNodeClone := oldNode.DeepCopy()
+	newNodeClone.ObjectMeta.Annotations = newNode.ObjectMeta.Annotations
+
+	_, err := c.CoreV1().Nodes().Update(newNodeClone)
+	if err != nil {
```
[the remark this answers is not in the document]
guydaichs: Hi @ggaurav10  - I was keeping with the conventions of the other utils here: `NotFound` errors are retried, but the caller suppresses the error to avoid re-queuing of the machine. This is less limiting for future consumers that may want to handle the `NotFound` error differently when using the annotation util. Is that acceptable to you?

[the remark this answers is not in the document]
ggaurav10: sure, sounds good. Just wanted to make sure if a custom error like the one sent in this function would also result in the desired behaviour in case of `NotFound` error.

guydaichs:
Hi @amshuman-kr 
> isn't it better to expose this information via node.status.conditions (say, with type: Termination and reason: Unhealthy|ScaleDown|RollingUpdate

Providing a proper reason for this condition is a bit tricky. I couldn't find a clear termination reason maintained by MCM on the machine object's status, so I only used the machine's state prior to termination (Healthy/Failed/Unknown). 

I suppose that it is possible to extend the `MachineSet` and `Machine` status to include a `LastOperation` reason. However, I suspect that this would require a significant change in the `MachineSet` and `MachineDeployment` controllers, in order to classify every termination operations and propagate the reason. 

I agree that a condition would provide more value for consumers. Do you see a way for MCM to provide the reason without making significant changes? for example, still rely on the `Phase` to speculate the termination reason: If the machine is failed, MCM will assume that the termination reason is `unhealthy`, otherwise:  `deleted`. In future, we can enhance this condition to provide more accurate termination reasons. WDYT?

amshuman-kr:
> Do you see a way for MCM to provide the reason without making significant changes? for example, still rely on the Phase to speculate the termination reason: If the machine is failed, MCM will assume that the termination reason is unhealthy, otherwise: deleted. In future, we can enhance this condition to provide more accurate termination reasons. WDYT?

@guydaichs Thanks for pointing this out. I hesitated before typing `|RollingUpdate` for this reason 😅 
I agree. We could start with using `reason: Unhealthy` for unhealthy or failed machines and a general `reason: ScaleDown` for all other scenarios leading to machine deletion and only introduce more reasons later. We could even justify using `ScaleDown` because in all those cases, technically, the `MachineSet` _is_ being scaled down.

@prashanth26 Is there any existing mechanism to set the node conditions in MCM? I see that [`syncMachineNodeTemplates `](https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/machine_util.go#L285-L346) only syncs annotations, labels and taints between `machines` and `nodes`. And [`updateMachineConditions`](https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/machine.go#L786-L879) seem to copy node conditions to machine conditions.

prashanth26:
> I agree that a condition would provide more value for consumers. Do you see a way for MCM to provide the reason without making significant changes? for example, still rely on the Phase to speculate the termination reason: If the machine is failed, MCM will assume that the termination reason is unhealthy, otherwise: deleted. In the future, we can enhance this condition to provide more accurate termination reasons. WDYT?

I agree. We could start with something simple is deriving it from the `Phase` field. In the future we could add more logic to it. 

> @prashanth26 Is there any existing mechanism to set the node conditions in MCM? I see that [`syncMachineNodeTemplates `](https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/machine_util.go#L285-L346) only syncs annotations, labels and taints between `machines` and `nodes`. And [`updateMachineConditions`](https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/machine.go#L786-L879) seem to copy node conditions to machine conditions.

@amshuman-kr - At this point, we only modify annotations, labels, and taints on nodes and nothing more. We don't update any node conditions yet.

hardikdr:
Having it as a `NodeCondition` would only be about appending a new node-condition in the node-object, instead of annotating. Currently, MCM doesn't modify/add any conditions, it only copies it from machine-object.
- Also, for now it should be fine to speculate the `Unhealthyness` due to the `Failed` phase, and in all other cases simply `ScaleDown`, as discussed above. 

Also, I am sure finding the exact reason behind the node-termination will demand larger changes and could be avoided as not the need of the hour, in this PR.

Though I am fine with either of the approaches, annotation, or condition, happy to leave the decision to @guydaichs .

pkg/util/conditions/conditions.go
```diff
@@ -0,0 +1,46 @@
+package annotations
```
hardikdr, line 1: Guess, this should be re-named to `conditions` :)

guydaichs:
As discussed, I re-implemented this PR using node conditions. Please note that some RBAC changes are also required. If any additional deployment descriptors should be updated, please let me know.

pkg/util/conditions/conditions.go
guydaichs: My bad. Fixed this and some other annotation references.

pkg/controller/controller_utils.go
```diff
@@ -1032,6 +1033,53 @@ func UpdateNodeTaints(c clientset.Interface, nodeName string, oldNode *v1.Node,
...
+}
+
+// AddOrUpdateConditionsOnNode adds a condition to the node's status
+func AddOrUpdateConditionsOnNode(c clientset.Interface, nodeName string, condition v1.NodeCondition) error {
+	firstTry := true
+	return clientretry.RetryOnConflict(Backoff, func() error {
+		var err error
+		var oldNode *v1.Node
+		if firstTry {
+			oldNode, err = c.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{ResourceVersion: "0"})
```
amshuman-kr, line 1052: @guydaichs Could you please explain the rationale behind the `firstTry` logic?

pkg/controller/machine.go
```diff
@@ -565,10 +565,52 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
...
+				forceDeleteLabelPresent,
+				timeOutOccurred,
+			)
+		}
 
 		// If machine was created on the cloud provider
 		machineID, _ := driver.GetExisting()
 
+		// update node with the machine's state prior to termination
+		if nodeName != "" {
```
amshuman-kr, line 603: Also, what is the rationale behind removing the `machineID != "" && ` part of the check?

pkg/controller/machine_util.go
```diff
@@ -496,3 +504,61 @@ func SyncMachineTaints(
...
+
+	if cond != nil && machine.Status.CurrentStatus.Phase == v1alpha1.MachineTerminating {
+		// do not consider machine terminating phase if node already terminating
+		terminationCondition.Reason = cond.Reason
+		terminationCondition.Message = cond.Message
+	} else {
+		setTerminationReasonByPhase(machine.Status.CurrentStatus.Phase, &terminationCondition)
+	}
+
+	if machine.Status.CurrentStatus.Phase == v1alpha1.MachineFailed {
```
amshuman-kr, line 539: Isn't this if-else block purious after the call to `setTerminationReasonByPhase`?

pkg/controller/controller_utils.go
guydaichs: That's the pattern for fetching nodes in this util, apparently related to a caching optimization. I added a comment explaining this. If you feel that this is redundant, let me know, and i'll remove it.

amshuman-kr: Thanks for the clarification and the comment. I checked the other places with the same pattern. This is fine.

prashanth26:
Hi Colleagues,

Something we have probably forgotten to consider is that this logic is currently maintained in two code paths also here - https://github.com/gardener/machine-controller-manager/blob/master/pkg/util/provider/machinecontroller/machine_util.go#L817-L951. Ideally the change has to be in both code paths until we have completely finished the OOT migration. 

CC: @amshuman-kr

pkg/controller/machine_util.go
```diff
@@ -496,3 +504,51 @@ func SyncMachineTaints(
...
 }
+
+func (c *controller) UpdateNodeTerminationCondition(machine *v1alpha1.Machine) error {
+	if machine.Status.CurrentStatus.Phase == "" {
+		return nil
+	}
+
+	nodeName := machine.Status.Node
+
+	terminationCondition := v1.NodeCondition{
```
hardikdr, line 515: Just something to mention, now we have a node-condition which only pops up during the machine-deletion. 

I believe, node-conditions are generally maintained throughout the life of the machine/resource, and updated with the current state.
Would be a little weird to have it while the machine is running as well, as it's not really terminated. I cant see, anything that needs to be done with this PR.
cc @amshuman-kr

pkg/controller/machine.go
hardikdr: Can you please resolve it, if the intended change is done already.

amshuman-kr: @hardikdr I do not get that option. I do not seem to have the access rights. But the changes are done, so actually resolved. I have approved too.

pkg/controller/machine_util.go
amshuman-kr: @hardikdr Good point. I am not sure about the behaviour of custom node conditions maintained by NPD, if they are maintained throughout node life-cycle or maintained when the condition occurs. Might be good check.

Regarding maintaining the termination condition throughout the node life-cycle, if we call the type `Termination` then it doesn't sound right to maintain it throughout. But if we name if something like `SelectedForTermination`, then it might still be ok to maintain it throughout and not just during machine deletion.

I am not proposing that we already change this PR to maintain the condition throughout the node life-cycle. Merely to rename the condition type to work both when we add it momentarily during machine deletion and when we want to change in the future to maintain it always.

WDYT?

hardikdr (approved):
Overall, looks good, thanks for the PR @guydaichs .

As per the discussion with Prashanth, we realized, it's better to have the related OOT changes be done in a subsequent PR. 
/lgtm

prashanth26 (approved):

pkg/controller/machine.go
```diff
@@ -565,10 +565,52 @@ func (c *controller) machineDelete(machine *v1alpha1.Machine, driver driver.Driv
...
+				machine.Name,
+				forceDeleteLabelPresent,
+				timeOutOccurred,
+			)
+		}
 
 		// If machine was created on the cloud provider
 		machineID, _ := driver.GetExisting()
 
+		// update node with the machine's state prior to termination
```
prashanth26, line 602: Hi @guydaichs ,

Looks like this change in moving the machine phase to termination is affecting the drain logic making it trigger a force deletion (as force deletion is dependent on the last update time whicch is to be set on machine deletion) many of times. Refer - https://github.com/gardener/machine-controller-manager/issues/563. I will have to move the setting machine Terminating phase to before updating the node conditions. - https://github.com/gardener/machine-controller-manager/pull/492/files#diff-d52f3782153512c736878e861c6cafd75dd23f214af3f017eea14de256da458bL572.

PR #493, opened 2020-07-30
Fixes regressions caused by PR #471

prashanth26:
https://github.com/gardener/machine-controller-manager/pull/471

Fixes regressions caused by [PR #471](https://github.com/gardener/machine-controller-manager/pull/471)

With k8s version 1.18 `metadata.description` isn't supported. And also there was a bug while trying to create a machineSet with a null value for `newstatus.LastOperation.LastUpdateTime` which can no longer be null due to validation. 

```
E0730 17:15:56.773686   16403 machineset.go:535] Update machineSet test-machine-deployment-7dbb854886 failed with: MachineSet.machine.sapcloud.io "test-machine-deployment-7dbb854886" is invalid: status.lastOperation.lastUpdateTime: Invalid value: "null": status.lastOperation.lastUpdateTime in body must be of type string: "null"
```

**Release note**:

```improvement developer
NONE
```

### DIFF
--- a/kubernetes/crds/machine.sapcloud.io_machinedeployments.yaml
+++ b/kubernetes/crds/machine.sapcloud.io_machinedeployments.yaml
@@ -60,7 +60,6 @@ spec:
             submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
           type: string
         metadata:
-          description: Standard object metadata.
           type: object
         spec:
           description: Specification of the desired behavior of the MachineDeployment.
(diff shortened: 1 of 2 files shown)


kubernetes/crds/machine.sapcloud.io_machinedeployments.yaml
```diff
@@ -60,7 +60,6 @@ spec:
             submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
           type: string
         metadata:
-          description: Standard object metadata.
```
hardikdr, line 63: >With k8s version 1.18 metadata.description isn't supported
Can you please give a bit more context on what does it mean? 
Also, do we also need to remove from other crd types, or is it already removed?

prashanth26: I wasn't able to check in detail. However, when I tried to deploy this onto k8 1.18 it failed with a similar message - https://github.com/kubernetes/kubernetes/issues/80493. 

However, once I removed it worked well. It probably is an unknown validation field or a deprecated one. Also, i don't think it is meant for meaning anything. 

I checked other CRDs and they don't have this field set so should be fine there.

Issue #472, opened 2020-06-10
Disable scale-down during cluster-rollouts.


hardikdr:
**What would you like to be added**: 

As discussed here https://github.com/gardener/autoscaler/issues/48#issuecomment-640600405 , we should add the annotation `cluster-autoscaler.kubernetes.io/scale-down-disabled` on the new machine-sets during roll-outs. 
- This will allow the system to keep CA running even during rolling-updates.

prashanth26:
Points to remember during the implementation refer - https://github.com/gardener/autoscaler/issues/48#issuecomment-640600405

PR #496, opened 2020-08-14
Inhibit scale-down by autoscaler during roll-outs.

hardikdr:
**What this PR does / why we need it**: With this PR, during roll out of the machine-deployment, all the machines under it are annotated with `cluster-autoscaler.kubernetes.io/scale-down-disabled: "True"`. 

This will help us keeping the cluster-autoscaler running even during the cluster-rollouts. CA will not scale-down a specific machine-deployment which is being rolled, but will be able to scale-down the rest of them. Also, CA will be able to scale-up all the machine-deployments during roll-outs.

There are mainly 2 parts to the PR:
1. Adding the annotation during roll-out.
   - To determine if the roll-out if on-going we check if there are old-machine-sets that are not scaled-to-zero completely. We check the `up-to-date`, and `ready` replicas as well here.
 
2. Removing the annotation after the roll-out.
   - Keeping the controller nature in mind, we need a way to differentiate between 2 possibilities:
       - Cluster rollout just got over, MachineDeployment is complete[all machine-replicas ready] and we should remove the annotation. 
       - There have been no roll-outs recently, MachineDeployment is complete and we shouldn't remove the annotation. Looking at older machine-set doesn't help as I see them retained after roll-outs.
    - If we don't consider the possibilities above, we may end up removing the same ca-annotation which might have been added by the user via `NodeTemplate`.
    - I have mitigated the situation, by introducing the `by-mcm` annotation, which is added on the node along with the original annotation. This place-holder-annotation helps in deciding if the original annotation is added by mcm or by the user. This helps in avoiding potential race-condition with `nodeTemplate`  logic.
    - Open for further suggestions or alternatives. 

Unfortunately, we cant use the `nodeTemplate` feature for adding/removing the annotation:
1. If we set the annotation on the MachineSet's `nodeTemplate`, machine-deployment overwrites it back.
   - If we ignore the specific annotation in code, user-provided CA-annotation might also be overwritten. 
2. If we set the annotation on the MachineDeployment's `nodeTemplate`, worker-controller will eventually overwrite it to the original state.

Fixes https://github.com/gardener/autoscaler/issues/48
FIxes https://github.com/gardener/machine-controller-manager/issues/472

We need to consider the possibility that users might also be using the same annotation for their use cases. We should make sure, autoscaler annotations from users are maintained. 

**Release note**:

```improvement operator
Machine-deployments being rolled-out are annotated with `cluster-autoscaler.kubernetes.io/scale-down-disabled: "True"` for the duration of the rolling-update.
```

### DIFF
--- a/cmd/machine-controller-manager/app/options/options.go
+++ b/cmd/machine-controller-manager/app/options/options.go
@@ -62,6 +62,7 @@ func NewMCMServer() *MCMServer {
 			KubeAPIBurst:            30,
 			LeaderElection:          leaderelectionconfig.DefaultLeaderElectionConfiguration(),
 			ControllerStartInterval: metav1.Duration{Duration: 0 * time.Second},
+			AutoscalerScaleDownAnnotationDuringRollout: true,
 			SafetyOptions: machineconfig.SafetyOptions{
 				SafetyUp:                                 2,
 				SafetyDown:                               1,
@@ -114,6 +115,8 @@ func (s *MCMServer) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&s.NodeConditions, "node-conditions", s.NodeConditions, "List of comma-separated/case-sensitive node-conditions which when set to True will change machine to a failed state after MachineHealthTimeout duration. It may further be replaced with a new machine if the machine is backed by a machine-set object.")
 	fs.StringVar(&s.BootstrapTokenAuthExtraGroups, "bootstrap-token-auth-extra-groups", s.BootstrapTokenAuthExtraGroups, "Comma-separated list of groups to set bootstrap token's \"auth-extra-groups\" field to")
 	fs.BoolVar(&s.DeleteMigratedMachineClass, "delete-migrated-machine-class", false, "Deletes any (provider specific) machine class that has the machine.sapcloud.io/migrated annotation")
+
+	fs.BoolVar(&s.AutoscalerScaleDownAnnotationDuringRollout, "autoscaler-scaldown-annotation-during-rollout", true, "Add cluster autoscaler scale-down disabled annotation during roll-out.")
 
 	leaderelectionconfig.BindFlags(&s.LeaderElection, fs)
 	// TODO: DefaultFeatureGate is global and it adds all k8s flags
--- a/pkg/controller/controller_utils.go
+++ b/pkg/controller/controller_utils.go
@@ -33,10 +33,12 @@ import (
 	"sync/atomic"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/validation"
 
 	"github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
 	machineapi "github.com/gardener/machine-controller-manager/pkg/client/clientset/versioned/typed/machine/v1alpha1"
+	annotationsutils "github.com/gardener/machine-controller-manager/pkg/util/annotations"
 	conditionutils "github.com/gardener/machine-controller-manager/pkg/util/conditions"
 	hashutil "github.com/gardener/machine-controller-manager/pkg/util/hash"
 	taintutils "github.com/gardener/machine-controller-manager/pkg/util/taints"
@@ -91,6 +93,13 @@ const (
 
 // Backoff is the backoff period used while updating nodes
 var Backoff = wait.Backoff{
+	Steps:    5,
+	Duration: 100 * time.Millisecond,
+	Jitter:   1.0,
+}
+
+// UpdateAnnotationBackoff is the backoff period used while updating the annotation
+var UpdateAnnotationBackoff = wait.Backoff{
 	Steps:    5,
 	Duration: 100 * time.Millisecond,
 	Jitter:   1.0,
@@ -1110,4 +1119,123 @@ func ComputeHash(template *v1alpha1.MachineTemplateSpec, collisionCount *int32) 
 	}
 
 	return machineTemplateSpecHasher.Sum32()
+}
+
+// AddOrUpdateAnnotationOnNode add annotations to the node. If annotation was added into node, it'll issue API calls
+// to update nodes; otherwise, no API calls. Return error if any.
+func AddOrUpdateAnnotationOnNode(c clientset.Interface, nodeName string, annotations map[string]string) error {
+	if annotations == nil {
+		return nil
+	}
+	firstTry := true
+	return clientretry.RetryOnConflict(UpdateAnnotationBackoff, func() error {
+		var err error
+		var oldNode *v1.Node
+		// First we try getting node from the API server cache, as it's cheaper. If it fails
+		// we get it from etcd to be sure to have fresh data.
+		if firstTry {
+			oldNode, err = c.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{ResourceVersion: "0"})
+			firstTry = false
+		} else {
+			oldNode, err = c.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{})
+		}
+		if errors.IsNotFound(err) {
+			klog.Warningf("Node %s not found while updating annotation. Err: %v", nodeName, err)
+			return nil
+		}
+		if err != nil {
+			return err
+		}
+
+		var newNode *v1.Node
+		updated := false
+
+		newNode, updated, err = annotationsutils.AddOrUpdateAnnotation(oldNode, annotations)
+
+		if !updated {
+			return nil
+		}
+		return UpdateNodeAnnotations(c, nodeName, oldNode, newNode)
+	})
+}
+
+// UpdateNodeAnnotations is for updating the node annotations from oldNode to the newNode
+// using the nodes Update() method
+func UpdateNodeAnnotations(c clientset.Interface, nodeName string, oldNode *v1.Node, newNode *v1.Node) error {
+	newNodeClone := oldNode.DeepCopy()
+	newNodeClone.Annotations = newNode.Annotations
+
+	_, err := c.CoreV1().Nodes().Update(newNodeClone)
+	if err != nil {
+		return fmt.Errorf("failed to create or update annotations for node %q: %v", nodeName, err)
+	}
+
+	return err
+}
+
+// RemoveAnnotationsOffNode is for cleaning up annotations temporarily added to node,
+// won't fail if target annotation doesn't exist or has been removed.
+// If passed a node it'll check if there's anything to be done, if annotation is not present it won't issue
+// any API calls.
+func RemoveAnnotationsOffNode(c clientset.Interface, nodeName string, annotations map[string]string) error {
+
+	// Short circuit if annotation doesnt exist for limiting API calls.
+	if annotations == nil || nodeName == "" {
+		return nil
+	}
+
+	firstTry := true
+	return clientretry.RetryOnConflict(UpdateAnnotationBackoff, func() error {
+		var err error
+		var oldNode *v1.Node
+		// First we try getting node from the API server cache, as it's cheaper. If it fails
+		// we get it from etcd to be sure to have fresh data.
+		if firstTry {
+			oldNode, err = c.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{ResourceVersion: "0"})
+			firstTry = false
+		} else {
+			oldNode, err = c.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{})
+		}
+		if errors.IsNotFound(err) {
+			klog.Warningf("Node %s not found while removing annotation. Err: %v", nodeName, err)
+			return nil
+		}
+
+		if err != nil {
+			return err
+		}
+
+		var newNode *v1.Node
+		oldNodeCopy := oldNode
+		updated := false
+
+		// Remove the annotations from the node.
+		newNode, updated, err = annotationsutils.RemoveAnnotation(oldNodeCopy, annotations)
+
+		if !updated {
+			return nil
+		}
+		return UpdateNodeAnnotations(c, nodeName, oldNode, newNode)
+	})
+}
+
+// GetAnnotationsFromNode returns all the annotations of the provided node.
+func GetAnnotationsFromNode(c clientset.Interface, nodeName string) (map[string]string, error) {
+
+	// Short circuit if annotation doesnt exist for limiting API calls.
+	if nodeName == "" {
+		return nil, nil
+	}
+
+	node, err := c.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{})
+	if errors.IsNotFound(err) {
+		klog.Warningf("Node %s not found while fetching annotation. Err: %v", nodeName, err)
+		return nil, nil
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	return node.Annotations, nil
 }
--- a/pkg/controller/deployment_recreate.go
+++ b/pkg/controller/deployment_recreate.go
@@ -25,10 +25,18 @@ package controller
 import (
 	"github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog"
 )
 
 // rolloutRecreate implements the logic for recreating a machine set.
 func (dc *controller) rolloutRecreate(d *v1alpha1.MachineDeployment, isList []*v1alpha1.MachineSet, machineMap map[types.UID]*v1alpha1.MachineList) error {
+
+	clusterAutoscalerScaleDownAnnotations := make(map[string]string)
+	clusterAutoscalerScaleDownAnnotations[ClusterAutoscalerScaleDownDisabledAnnotationKey] = ClusterAutoscalerScaleDownDisabledAnnotationValue
+
+	// We do this to avoid accidentally deleting the user provided annotations.
+	clusterAutoscalerScaleDownAnnotations[ClusterAutoscalerScaleDownDisabledAnnotationByMCMKey] = ClusterAutoscalerScaleDownDisabledAnnotationByMCMValue
+
 	// Don't create a new RS if not already existed, so that we avoid scaling up before scaling down.
 	newIS, oldISs, err := dc.getAllMachineSetsAndSyncRevision(d, isList, machineMap, false)
 	if err != nil {
@@ -36,6 +44,20 @@ func (dc *controller) rolloutRecreate(d *v1alpha1.MachineDeployment, isList []*v
 	}
 	allISs := append(oldISs, newIS)
 	activeOldISs := FilterActiveMachineSets(oldISs)
+
+	if dc.autoscalerScaleDownAnnotationDuringRollout {
+		// Add the annotation on the all machinesets if there are any old-machinesets and not scaled-to-zero.
+		// This also helps in annotating the node under new-machineset, incase the reconciliation is failing in next
+		// status-rollout steps.
+		if len(oldISs) > 0 && !dc.machineSetsScaledToZero(oldISs) {
+			// Annotate all the nodes under this machine-deployment, as roll-out is on-going.
+			err := dc.annotateNodesBackingMachineSets(allISs, clusterAutoscalerScaleDownAnnotations)
+			if err != nil {
+				klog.Errorf("Failed to add %s on all nodes. Error: %s", clusterAutoscalerScaleDownAnnotations, err)
+				return err
+			}
+		}
+	}
 
 	// scale down old machine sets.
 	scaledDown, err := dc.scaleDownOldMachineSetsForRecreate(activeOldISs, d)
@@ -67,6 +89,14 @@ func (dc *controller) rolloutRecreate(d *v1alpha1.MachineDeployment, isList []*v
 	}
 
 	if MachineDeploymentComplete(d, &d.Status) {
+		if dc.autoscalerScaleDownAnnotationDuringRollout {
+			// Check if any of the machine under this MachineDeployment contains the by-mcm annotation, and
+			// remove the original autoscaler-annotion only after.
+			err := dc.removeAutoscalerAnnotationsIfRequired(allISs, clusterAutoscalerScaleDownAnnotations)
+			if err != nil {
+				return err
+			}
+		}
 		if err := dc.cleanupMachineDeployment(oldISs, d); err != nil {
 			return err
 		}
--- a/pkg/controller/deployment_rolling.go
+++ b/pkg/controller/deployment_rolling.go
@@ -44,6 +44,13 @@ var (
 
 // rolloutRolling implements the logic for rolling a new machine set.
 func (dc *controller) rolloutRolling(d *v1alpha1.MachineDeployment, isList []*v1alpha1.MachineSet, machineMap map[types.UID]*v1alpha1.MachineList) error {
+
+	clusterAutoscalerScaleDownAnnotations := make(map[string]string)
+	clusterAutoscalerScaleDownAnnotations[ClusterAutoscalerScaleDownDisabledAnnotationKey] = ClusterAutoscalerScaleDownDisabledAnnotationValue
+
+	// We do this to avoid accidentally deleting the user provided annotations.
+	clusterAutoscalerScaleDownAnnotations[ClusterAutoscalerScaleDownDisabledAnnotationByMCMKey] = ClusterAutoscalerScaleDownDisabledAnnotationByMCMValue
+
 	newIS, oldISs, err := dc.getAllMachineSetsAndSyncRevision(d, isList, machineMap, true)
 	if err != nil {
 		return err
@@ -57,6 +64,21 @@ func (dc *controller) rolloutRolling(d *v1alpha1.MachineDeployment, isList []*v1
 			Effect: "PreferNoSchedule",
 		},
 	)
+
+	if dc.autoscalerScaleDownAnnotationDuringRollout {
+		// Add the annotation on the all machinesets if there are any old-machinesets and not scaled-to-zero.
+		// This also helps in annotating the node under new-machineset, incase the reconciliation is failing in next
+		// status-rollout steps.
+		if len(oldISs) > 0 && !dc.machineSetsScaledToZero(oldISs) {
+			// Annotate all the nodes under this machine-deployment, as roll-out is on-going.
+			err := dc.annotateNodesBackingMachineSets(allISs, clusterAutoscalerScaleDownAnnotations)
+			if err != nil {
+				klog.Errorf("Failed to add %s on all nodes. Error: %s", clusterAutoscalerScaleDownAnnotations, err)
+				return err
+			}
+		}
+	}
+
 	if err != nil {
 		klog.Warningf("Failed to add %s on all nodes. Error: %s", PreferNoScheduleKey, err)
 	}
@@ -82,6 +104,14 @@ func (dc *controller) rolloutRolling(d *v1alpha1.MachineDeployment, isList []*v1
 	}
 
 	if MachineDeploymentComplete(d, &d.Status) {
+		if dc.autoscalerScaleDownAnnotationDuringRollout {
+			// Check if any of the machine under this MachineDeployment contains the by-mcm annotation, and
+			// remove the original autoscaler annotation only after.
+			err := dc.removeAutoscalerAnnotationsIfRequired(allISs, clusterAutoscalerScaleDownAnnotations)
+			if err != nil {
+				return err
+			}
+		}
 		if err := dc.cleanupMachineDeployment(oldISs, d); err != nil {
 			return err
 		}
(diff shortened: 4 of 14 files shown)


pkg/controller/deployment_rolling.go
```diff
@@ -341,3 +365,161 @@ func (dc *controller) taintNodesBackingMachineSets(MachineSets []*v1alpha1.Machi
...
+	for _, machineSet := range MachineSets {
+		if machineSet.Spec.Replicas != 0 && machineSet.Status.AvailableReplicas != 0 && machineSet.Status.FullyLabeledReplicas != 0 && machineSet.Status.ReadyReplicas != 0 && machineSet.Status.Replicas != 0 {
+			return false
+		}
+	}
+	return true
+}
+
+// annotateNodesBackingMachineSets annotates all nodes backing the machineSets
+func (dc *controller) removeAnnotationFromNodesBackingMachineSets(MachineSets []*v1alpha1.MachineSet, annotations map[string]string) error {
```
hardikdr, line 424: This is a helper function, not actively used yet. Happy to remove, if we don't see it being useful.

pkg/apis/machine/v1alpha1/shared_types.go
```diff
@@ -72,3 +72,15 @@ type MachineSummary struct {
...
+	// ClusterAutoscalerScaleDownDisabledAnnotationKey annotation to disable the scale-down of the nodes.
+	ClusterAutoscalerScaleDownDisabledAnnotationKey = "cluster-autoscaler.kubernetes.io/scale-down-disabled"
+	// ClusterAutoscalerScaleDownDisabledAnnotationValue annotation to disable the scale-down of the nodes.
+	ClusterAutoscalerScaleDownDisabledAnnotationValue = "True"
+
+	// ClusterAutoscalerScaleDownDisabledAnnotationByMCMKey annotation to disable the scale-down of the nodes.
+	ClusterAutoscalerScaleDownDisabledAnnotationByMCMKey = "cluster-autoscaler.kubernetes.io/scale-down-disabled-by-mcm"
+	// ClusterAutoscalerScaleDownDisabledAnnotationByMCMValue annotation to disable the scale-down of the nodes.
+	ClusterAutoscalerScaleDownDisabledAnnotationByMCMValue = "True"
+)
```
prashanth26, line 86: I would expect that these consts to be moved out of this package as it's less related to the APIs. 
Probably `pkg/util/*` or `pkg/controller` ?

hardikdr: Hm, I kept it here, thinking this keyword will actually be an API contract for external consumption. eg Cluster autoscaler. This has to be maintained across implementations, though open for the thoughts.

vlerenc:
> I have mitigated the situation, by introducing the by-mcm annotation, which is added on the node along with the original annotation. This place-holder-annotation helps in deciding if the original annotation is added by mcm or by the user. This helps in avoiding potential race-condition with nodeTemplate logic.
> Open for further suggestions or alternatives.

Not really, nothing more elegant. That would have been also my approach/was also what I was thinking when I heard/read the above.

amshuman-kr (commented):
@hardikdr Thanks for the PR. Just a couple of questions/comments. Otherwise looks good, apart from tests.

pkg/controller/deployment_rolling.go
```diff
@@ -44,6 +44,13 @@ var (
 
 // rolloutRolling implements the logic for rolling a new machine set.
 func (dc *controller) rolloutRolling(d *v1alpha1.MachineDeployment, isList []*v1alpha1.MachineSet, machineMap map[types.UID]*v1alpha1.MachineList) error {
+
+	clusterAutoscalerScaleDownAnnotations := make(map[string]string)
+	clusterAutoscalerScaleDownAnnotations[v1alpha1.ClusterAutoscalerScaleDownDisabledAnnotationKey] = v1alpha1.ClusterAutoscalerScaleDownDisabledAnnotationValue
+
+	// We do this avoid accidentally deleting the user provided annotations.
+	clusterAutoscalerScaleDownAnnotations[v1alpha1.ClusterAutoscalerScaleDownDisabledAnnotationByMCMKey] = v1alpha1.ClusterAutoscalerScaleDownDisabledAnnotationByMCMValue
```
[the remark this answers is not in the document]
hardikdr: Hi Amshu, I believe, if we add the `by-mcm` annotation on the MachineDeployment's NodeTemplate, the worker-extension will overwrite it back to original.

pkg/controller/deployment_rolling.go
```diff
@@ -82,11 +89,28 @@ func (dc *controller) rolloutRolling(d *v1alpha1.MachineDeployment, isList []*v1
...
+		if err != nil {
+			return err
+		}
 		if err := dc.cleanupMachineDeployment(oldISs, d); err != nil {
 			return err
 		}
 	}
 
+	// Check if all the oldMachineSets are set to zero, if not there is likely be a roll-out ongoing.
+	if len(oldISs) > 0 && !MachineDeploymentComplete(d, &d.Status) && !dc.machineSetsScaledToZero(oldISs) {
```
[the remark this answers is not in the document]
hardikdr: This logic adds the annotations on all the node-objects immediately the roll-out starts, seems that's the best MCM could in this case.

hardikdr:
Thanks for the reviews.
I'll rebase and add unit tests soon.

pkg/controller/deployment_rolling.go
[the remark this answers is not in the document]
amshuman-kr: Thanks, understood. Then a modified follow up question. How about using the `nodeTemplate` mechanism in the `machinesets`? I.e. adding the annotations in to the `nodeTemplate` in the `machinesets`?

The reason for asking this is that, if we annotate the nodes directly, then only existing nodes get annotated. If in the meantime any of the `machinesets` get scaled up (probably, manually), the new nodes won't get the annotation unless a `machinedeploy` reconciliation is executed. Admittedly, that is a contrived scenario and not at all a common or important one. But thought it might be nice if we can avoid that edge case.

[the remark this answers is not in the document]
hardikdr: >Then a modified follow up question. How about using the nodeTemplate mechanism in the machinesets? I.e. adding the annotations in to the nodeTemplate in the machinesets?

In this case, it'd be overwritten again. If there are any annotations on the `nodeTemplate` of the MachineSet but not on the `nodeTemplate` of the MachineDeployment, the deployment-controller shall overwrite it back to original. 

At the moment, every join of the node re-enqueues the Machine -> MachiineSet -> MachineDeployment, hence new nodes should immediately be annotated they join.  

Also, to mention, cluster-autoscaler prefers slow-scaledown, it has certain timeouts like `scale-down-delay-after-add` before which it decides to trigger scale-down of a given node. This should also complement the logic.

pkg/controller/deployment_rolling.go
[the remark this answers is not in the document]
amshuman-kr: > annotations on all the node-objects immediately the roll-out starts, seems that's the best MCM could in this case.

This is true. But that is not what the code in this PR does. The code annotates the nodes before the deployment status is synced. Machine roll-out starts already earlier with the [tainting of the old nodes](https://github.com/gardener/machine-controller-manager/pull/496/files#diff-d60d538aae4eb1c57f7cd31e87804382R60-R66) and the scaling of [new](https://github.com/gardener/machine-controller-manager/pull/496/files#diff-d60d538aae4eb1c57f7cd31e87804382R72) and [old](https://github.com/gardener/machine-controller-manager/pull/496/files#diff-d60d538aae4eb1c57f7cd31e87804382L75) `machinesets`. Or am I missing something?

amshuman-kr (commented):
One more question. The PR covers only the `RollingUpdate` strategy. Wouldn't the feature be complete if we cover the `Recreate` strategy too?

pkg/controller/deployment_rolling.go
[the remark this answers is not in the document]
amshuman-kr: > In this case, it'd be overwritten again. If there are any annotations on the `nodeTemplate` of the MachineSet but not on the `nodeTemplate` of the MachineDeployment, the deployment-controller shall overwrite it back to original.

Yes. But this argument is not as strong as the argument against updating the `nodeTemplate` on the `MachineDeployment` because modifying the worker-extension is not desirable. We could modify the existing code code path in the deployment controller to make it not overwrite the annotations rather than introduce a new place and mechanism for of annotating the nodes in the code. Similar logic already exists for not overwriting manually maintained annotations on the node. WDYT?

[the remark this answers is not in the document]
hardikdr: > However, we could modify the existing code code path in the deployment controller to make it not overwrite the annotations rather than introduce a new place and mechanism for annotating the nodes in the code.

There is already a mechanism to add the `PreferNoSchedule` taints directly on the node-objects once roll-out starts, this logic takes the same path. 

> Similar logic already exists for not overwriting manually maintained annotations on the node. WDYT?

MCM doesn't specifically ignore the manually added annotations on the node, it only maintains the ones which are part of the `nodeTemplate`, so there is actually no mechanism of ignoring the annotations in `nodeTemplate`.

Also, if MCM starts ignoring only `cluster-autoscaler.kubernetes.io/scale-down-disabled: "True"` in the machine-set's `nodeTemplate`, then the same annotation put by the end-user on the machine-deployment's `nodeTemplate` will also be ignored.  Then we need to find a way to ignore the tuple of annotations in `nodeTemplate`, which looks a little complex to me, and output would probably be not worth changing the entire approach.

pkg/controller/deployment_rolling.go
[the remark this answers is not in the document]
hardikdr: Ah, sorry, I mistook the taint by the ones put by the autoscaler, but I guess you meant the `PreferNoSchedule` taint from MCM. 
Can you please elaborate a bit around, how `PreferNoSchedule` taint and `ScaleDown` annotation can impact each other's function?

hardikdr:
>The PR covers only the RollingUpdate strategy. Wouldn't the feature be complete if we cover the Recreate strategy too?

Recreate strategy simply deletes all the existing machines and creates the new ones altogether. That is already super disruptive, not sure how autoscaler could harm it any further.

pkg/controller/deployment_rolling.go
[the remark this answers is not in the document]
amshuman-kr: >  if MCM starts ignoring only cluster-autoscaler.kubernetes.io/scale-down-disabled: "True" in the machine-set's nodeTemplate, then the same annotation put by the end-user on the machine-deployment's nodeTemplate will also be ignored.

I was not suggesting to not use the `by-mcm` annotation. That approach would work equally well if the nodes are annotated directly or via the `nodeTemplate`.

> There is already a mechanism to add the `PreferNoSchedule` taints directly on the node-objects once roll-out starts, this logic takes the same path.

I would suggest to change that also to use the `nodeTemplate` path to streamline the interaction with nodes. But not in this PR, of course.

But anyway, my point about `nodeTemplate` was a nice-to-have to avoid too many paths for modifying the node objects and that it makes life difficult to simulate all the scenarios to be tested.

pkg/controller/deployment_rolling.go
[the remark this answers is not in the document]
amshuman-kr: > Recreate strategy simply deletes all the existing machines and creates the new ones altogether. That is already super disruptive, not sure how autoscaler could harm it any further.

What if CA scales down/up while the new machines are still joining? Isn't it better to cover all bases?

amshuman-kr:
> Recreate strategy simply deletes all the existing machines and creates the new ones altogether. That is already super disruptive, not sure how autoscaler could harm it any further.

What if CA scales down/up while the new machines are still joining? Isn't it better to cover all bases?

pkg/controller/deployment_rolling.go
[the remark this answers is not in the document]
amshuman-kr: >Ah, sorry, I mistook the taint by the ones put by the autoscaler, but I guess you meant the PreferNoSchedule taint from MCM.

Yes, I was talking about the `PreferNoSchedule` taints. But not only that. I was also talking about `reconcileNewMachineSet` and `reconcileOldMachineSets` which also happen before adding the annotation in the current PR. Not only that, the code path of `reconcileNewMachineSet` and `reconcileOldMachineSets` might even [return](https://github.com/gardener/machine-controller-manager/blob/3def9452ec9b1796e5f2171616abc31a0ce7f84e/pkg/controller/deployment_rolling.go#L67) [from](https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/deployment_rolling.go#L71) reconciliation [without](https://github.com/gardener/machine-controller-manager/blob/3def9452ec9b1796e5f2171616abc31a0ce7f84e/pkg/controller/deployment_rolling.go#L77) [adding](https://github.com/gardener/machine-controller-manager/blob/3def9452ec9b1796e5f2171616abc31a0ce7f84e/pkg/controller/deployment_rolling.go#L81) the annotation. Isn't it better to annotate the nodes before all this?

hardikdr:
>What if CA scales down/up while the new machines are still joining? Isn't it better to cover all bases?

I spent a bit more time and experimented. Basically, during `Recreate` the old machine-set is completely brought down and made sure no old machines are running. Once there are no old machines running, the new machine set is brought up to the full. A couple of points to ponder upon:

1.  It is hard to differentiate between the following cases: 
     - Scale-up of machineset due to `usual-scale-up of machine-deployment`
     - Scale-up of machineset due to `recreate-update` flow.
- It's certainly possible to make it work, just that it'll demand more changes in the code as it seems.

2.  CA scales down a node only if the node is registered and under-utilized for a certain period of time. So if few nodes have joined the cluster and not taking any `Pending` workload for a certain timeout period during `recreate-update`, I believe CA will anyways remove it immediately after roll-out. 

3. Scale-up is anyways allowed in both strategies. 

Overall, `Recreate` doesn't seem to be a production-friendly feature. If an application can survive literally no-machines in the cluster/worker-group, we should probably not invest too much to avoid minor scale down in between.
- Just to add, this is not being exposed/used by the gardener, also not planned afaik.

In summary, I felt it definitely sounds sensible to cover all the strategies for the feature from bird's eye view, but there are more details. But if we see the value, I am more than happy to somehow make it work. :)

pkg/controller/deployment_rolling.go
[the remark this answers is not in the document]
hardikdr: Hm, that's a good point. Though it's not possible to simply move the [logic](https://github.com/gardener/machine-controller-manager/blob/44d70a27e902b52d98d44da7717bc4ad471d099d/pkg/controller/deployment_rolling.go#L104-L112) up before tainting, else the `MachineDeploymentComplete` will return `true`, and MCM may not end up annotating at all.

But I do see the point that we don't want to see the unannotated nodes only because the reconciliation failed in-between for unknown reasons. I'll add the logic to annotate only the `OldMachineSets` near the taint-mechanism.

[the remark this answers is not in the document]
hardikdr: Can you please check if this change: https://github.com/gardener/machine-controller-manager/blob/b99e8a8371e2bb1f7de25461b26970e4b3e99ed4/pkg/controller/deployment_rolling.go#L71-L78  , serves the purpose?
- I'd expect, at the end when the MachineDeployment is complete [here](https://github.com/gardener/machine-controller-manager/blob/b99e8a8371e2bb1f7de25461b26970e4b3e99ed4/pkg/controller/deployment_rolling.go#L104), the annotations will be removed altogether.

pkg/apis/machine/v1alpha1/shared_types.go
prashanth26: Okay. I would prefer to keep it outside the `pkg/apis/machines` as it has very little to do with actual machine CRD contracts. Also, this code in this pkg should be duplicated then to even - https://github.com/gardener/machine-controller-manager/blob/44d70a27e902b52d98d44da7717bc4ad471d099d/pkg/apis/machine/v1alpha1/shared_types.go. I honestly don't think we should be adding such consts to these files which are used for auto generation and are duplicated. 

However, i do get your point about the API contracts part. How about perhaps moving this to something like `pkg/apis/clusterautoscaler`?

pkg/controller/deployment_rolling.go
prashanth26: I think we can remove it if it's not used. We can always look back at the commits and reintroduce it if required.

pkg/apis/machine/v1alpha1/shared_types.go
hardikdr: Good point, I moved it to the utils file of deployment with this [change](https://github.com/gardener/machine-controller-manager/pull/496/commits/3a557bbf6fc38ef5ade63b869364dff462d18558).

prashanth26: Thank you.

pkg/controller/deployment_rolling.go
hardikdr: Removed.

pkg/controller/deployment_rolling.go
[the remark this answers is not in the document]
amshuman-kr: Thanks! That looks better. But now the check on `err` from `taintNodesBackingMachineSets` could potentially overwritten before checking.

Basically, the following has to be immediately after the taint call.

https://github.com/gardener/machine-controller-manager/blob/b99e8a8371e2bb1f7de25461b26970e4b3e99ed4/pkg/controller/deployment_rolling.go#L80-L82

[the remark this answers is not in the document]
amshuman-kr: BTW, with this change I am not sure if the later annotate call is required at all.

https://github.com/gardener/machine-controller-manager/blob/b99e8a8371e2bb1f7de25461b26970e4b3e99ed4/pkg/controller/deployment_rolling.go#L117-L125

amshuman-kr:
> It is hard to differentiate between the following cases:
> * Scale-up of machineset due to `usual-scale-up of machine-deployment`
> * Scale-up of machineset due to `recreate-update` flow.

I still see slight difference between regular scale up and recreate update. For example, recreate update is still multiple steps.

1. Scale down old `MachineSet` and return for a reconciliation later.
https://github.com/gardener/machine-controller-manager/blob/e16e726ce2e9ab23734ea8c8dadd8b98396dabe5/pkg/controller/deployment_recreate.go#L38-L48

2. Check if old `MachineSet` has really scaled down and if true then return for a reconciliation.
https://github.com/gardener/machine-controller-manager/blob/e16e726ce2e9ab23734ea8c8dadd8b98396dabe5/pkg/controller/deployment_recreate.go#L50-L53

3. Scale up new `MachineSet` (create it if necessary).
https://github.com/gardener/machine-controller-manager/blob/e16e726ce2e9ab23734ea8c8dadd8b98396dabe5/pkg/controller/deployment_recreate.go#L55-L67

Only the third step seems to be comparable to the regular scale up. But even here, to my naive reading, it looks like there will be two reconciliations involved because the `MachineDeploymentComplete` check is done _before_ the `MachineDeployment` status has been updated.
https://github.com/gardener/machine-controller-manager/blob/e16e726ce2e9ab23734ea8c8dadd8b98396dabe5/pkg/controller/deployment_recreate.go#L69-L73

https://github.com/gardener/machine-controller-manager/blob/e16e726ce2e9ab23734ea8c8dadd8b98396dabe5/pkg/controller/deployment_recreate.go#L75-L76 

So, I see still value in adding/removing annotation at the beginning of this process and removing at the the end of all the multiple reconciliations involved. WDYT?

pkg/controller/deployment_rolling.go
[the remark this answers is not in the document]
hardikdr: Yes, that's true. I have also realized the same and experimenting a bit to make sure we don't lose a corner case. I'll most likely remove the annotation-logic put later.

hardikdr:
>So, I see still value in adding/removing annotation at the beginning of this process and removing at the the end of all the multiple reconciliations involved. WDYT?

I gave it a try [here](https://github.com/gardener/machine-controller-manager/pull/496/commits/0de2ff46db1998d329300467f602ad169bbc6b4c). I am not very confident yet, can you please take a look?

amshuman-kr:
> > So, I see still value in adding/removing annotation at the beginning of this process and removing at the the end of all the multiple reconciliations involved. WDYT?
> 
> I gave it a try [here](https://github.com/gardener/machine-controller-manager/pull/496/commits/0de2ff46db1998d329300467f602ad169bbc6b4c). I am not very confident yet, can you please take a look?

Thanks for the change! Looks good to me (haven't test though).

pkg/controller/deployment_rolling.go
```diff
@@ -341,3 +367,112 @@ func (dc *controller) taintNodesBackingMachineSets(MachineSets []*v1alpha1.Machi
...
+		}
+
+		for _, machine := range filteredMachines {
+			if machine.Status.Node != "" {
+
+				nodeAnnotations, err := GetAnnotationsFromNode(
+					dc.targetCoreClient,
+					machine.Status.Node,
+				)
+				if err != nil {
```
ggaurav10, line 454: should we ignore "not found" error?

pkg/util/annotations/annotations.go
```diff
@@ -0,0 +1,85 @@
...
+
+	for annotationKey, annotationValue := range annotations {
+		if nodeAnnotationValue, exists := nodeAnnotations[annotationKey]; exists {
+			if nodeAnnotationValue == annotationValue {
+				// Annotation is already available on the node.
+				continue
+			}
+			// If the given annotation exists in nodeAnnotation but value is not matching.
+			nodeAnnotations[annotationKey] = annotationValue
+			updated = true
```
ggaurav10, line 44:
```suggestion

```
This is anyway done in next 2 lines.

ggaurav10 (commented):
some minor suggestions. I believe we also need to add the logic to add/remove CA annotations in `rollback()`, isn't it?
Also, wondering if it would scale for large machine deployments. Any thoughts? Should we put this behind a global flag so adding/removing CA annotations can be disabled when CA is not even required to be deployed?
Just thinking out loud.

pkg/controller/controller_utils.go
```diff
@@ -1111,3 +1119,108 @@ func ComputeHash(template *v1alpha1.MachineTemplateSpec, collisionCount *int32)
...
+func UpdateNodeAnnotations(c clientset.Interface, nodeName string, oldNode *v1.Node, newNode *v1.Node) error {
+	newNodeClone := oldNode.DeepCopy()
+	newNodeClone.Annotations = newNode.Annotations
+
+	_, err := c.CoreV1().Nodes().Update(newNodeClone)
+	if err != nil {
+		return fmt.Errorf("failed to create or update annotations for node %q: %v", nodeName, err)
+	}
+
+	return nil
```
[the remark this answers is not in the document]
hardikdr: If `err != nil`, I am anyways returning error just above. But the suggestion looks cleaner, thanks.

pkg/controller/controller_utils.go
```diff
@@ -1111,3 +1119,108 @@ func ComputeHash(template *v1alpha1.MachineTemplateSpec, collisionCount *int32)
...
+}
+
+// RemoveAnnotationsOffNode is for cleaning up annotations temporarily added to node,
+// won't fail if target annotation doesn't exist or has been removed.
+// If passed a node it'll check if there's anything to be done, if annotation is not present it won't issue
+// any API calls.
+func RemoveAnnotationsOffNode(c clientset.Interface, nodeName string, annotations map[string]string) error {
+
+	// Short circuit if annotation doesnt exist for limiting API calls.
+	if annotations == nil && nodeName == "" {
```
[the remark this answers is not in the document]
hardikdr: Very good catch, thanks :)  .

pkg/controller/deployment_rolling.go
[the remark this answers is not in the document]
hardikdr: Done, removed.

pkg/controller/deployment_rolling.go
hardikdr: thanks, it's a good point. I have basically taken the methods from a similar mechanism of taints already implemented- check [here](https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/controller_utils.go#L900-L937).

But I see a value in ignoring the node-not-found error. I believe, otherwise, the controller will constantly retry to find the node, a bit of load on apiserver. 
@prashanth26  Did you see any reason for not handling the `NotFound` errors for taints as well back then. Else, I am happy to patch other methods related to taints[+conditions] here.

hardikdr: Done for the annotation [here](https://github.com/gardener/machine-controller-manager/pull/496/commits/1c1543420a6b3b89c78079aabedcb49d6bd862e4) . 

- A side thought would then be, that the machine will be ignored for further reconciliation if the node-object is going to join in sometime. I hope the event for node's joining should trigger the reconciliation later.

pkg/controller/deployment_rolling.go
```diff
@@ -82,6 +102,12 @@ func (dc *controller) rolloutRolling(d *v1alpha1.MachineDeployment, isList []*v1
 	}
 
 	if MachineDeploymentComplete(d, &d.Status) {
+		// Check if any of the machine under this MachineDeployment contains the by-mcm annotation, and
+		// remove the original autoscaler-annotion only after.
```
prashanth26, line 106:
```suggestion
		// remove the original autoscaler annotation only after.
```

pkg/controller/deployment_rolling.go
prashanth26: This method was already taking care of it, right? - https://github.com/gardener/machine-controller-manager/pull/496/files#diff-44bba2c68a6ea2b4ca0dbb7f4465828eR1231-R1234. No?

pkg/controller/deployment_recreate.go
```diff
@@ -25,10 +25,18 @@ package controller
...
 )
 
 // rolloutRecreate implements the logic for recreating a machine set.
 func (dc *controller) rolloutRecreate(d *v1alpha1.MachineDeployment, isList []*v1alpha1.MachineSet, machineMap map[types.UID]*v1alpha1.MachineList) error {
+
+	clusterAutoscalerScaleDownAnnotations := make(map[string]string)
+	clusterAutoscalerScaleDownAnnotations[ClusterAutoscalerScaleDownDisabledAnnotationKey] = ClusterAutoscalerScaleDownDisabledAnnotationValue
+
+	// We do this to avoid accidentally deleting the user provided annotations.
+	clusterAutoscalerScaleDownAnnotations[ClusterAutoscalerScaleDownDisabledAnnotationByMCMKey] = ClusterAutoscalerScaleDownDisabledAnnotationByMCMValue
```
prashanth26, line 38: We add this annotation in both cases, right? i.e When no such annotation is added by the user and when it is added by the user. How are we keeping track of when the annotation was added by user prior to adding of this annotation during rollout. Maybe I missed it in the changes?

amshuman-kr: It is not handled while adding the annotation but handled during [removal](https://github.com/gardener/machine-controller-manager/pull/496/files#diff-d60d538aae4eb1c57f7cd31e87804382R461).

pkg/controller/deployment_rolling.go
hardikdr: I added it just now :D 
There are some methods for taints implemented earlier, which doesn't handle the not-found case - https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/controller_utils.go#L900-L937 .
- I meant if we should update these methods already or keep it for later.

pkg/controller/deployment_recreate.go
hardikdr: Yes, basically we need a way for controller to identify if the existing ca-annotation was added by the user or mcm. 
- If the `by-mcm` annotation exists, MCM assumes, it is added by the MCM and removes it.
- In the next iteration, the user-initiated annotations will be put again, maybe via `NodeTemplate`. And not MCM won't remove it, as `by-mcm` annotation is missing.
- During the roll-out, we add the ca-annotations on all nodes, irrespective of they exist there already or not.

prashanth26: Okay yes, forgot about the `NodeTemplate` readdition.

hardikdr:
>I believe we also need to add the logic to add/remove CA annotations in rollback(), isn't it?

The existing solution should work with Rollback as well. I did a couple of experiments and looked fine.
Essentially, the following events cant be considered.
- Roll-out is initiated.
- All the nodes are annotated with the existing logic, in both old and new machine-sets.
- Roll-back is triggered.
   - At this point, the older-machineset is scaled up while the newer one is scaled-down.
   - MachineDeployment spec is modified automatically to reflect older-machine-set's spec.
   - Eventually, the same rolling-update is triggered, machines in the old-machineset get the annotations during roll-out.
   - Once rollback is over, `MachineDeploymentComplete` succeeds in the next reconciliation, and all the ca-annotations are removed.

This is from my quick look and experiments. 
Can you please also double-verify if possible? @ggaurav10

hardikdr:
>Should we put this behind a global flag so adding/removing CA annotations can be disabled when CA is not even required to be deployed?

I personally do see value in having a global-flag to switch it off, as there are users possibly consuming mcm without CA.
I'd like to hear the opinion of @prashanth26 @amshuman-kr. If we agree, I can add a switch, defaulting to `true`.

prashanth26:
> I personally do see value in having a global-flag to switch it off, as there are users possibly consuming mcm without CA.
I'd like to hear the opinion of @prashanth26 @amshuman-kr. If we agree, I can add a switch, defaulting to true

I think it makes sense as well.

amshuman-kr:
> I personally do see value in having a global-flag to switch it off, as there are users possibly consuming mcm without CA.
> I'd like to hear the opinion of @prashanth26 @amshuman-kr. If we agree, I can add a switch, defaulting to `true`.

Yes. I think a global flag is a good idea.

hardikdr:
I've added a switch, open for better naming suggestions.

cmd/machine-controller-manager/app/options/options.go
```diff
@@ -114,6 +115,8 @@ func (s *MCMServer) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&s.NodeConditions, "node-conditions", s.NodeConditions, "List of comma-separated/case-sensitive node-conditions which when set to True will change machine to a failed state after MachineHealthTimeout duration. It may further be replaced with a new machine if the machine is backed by a machine-set object.")
 	fs.StringVar(&s.BootstrapTokenAuthExtraGroups, "bootstrap-token-auth-extra-groups", s.BootstrapTokenAuthExtraGroups, "Comma-separated list of groups to set bootstrap token's \"auth-extra-groups\" field to")
 
+	fs.BoolVar(&s.AutoscalerScaleDownAnnotationDuringRollout, "autoscaler-scaldown-annotation-during-rollout", true, "Add cluster autoscaler scale-down disabled annotation during roll-out.")
```
prashanth26, line 118:
```suggestion
	fs.BoolVar(&s.AutoscalerScaleDownAnnotationDuringRollout, "autoscaler-scaledown-annotation-during-rollout", true, "Add cluster autoscaler scale-down disabled annotation during rollout/recreation of machine deployments")
```

prashanth26 (commented):
Minor typo. 
/lgtm otherwise

Also ditto comment on squashing commits.

prashanth26 (approved):

hardikdr:
@amshuman-kr @ggaurav10 Can you please take a final look and approve if possible?

ggaurav10 (approved):
lgtm 👍

amshuman-kr (approved):

Issue #444, opened 2020-04-14
Add an intermediate Machine state for non-atomic Cloud Provider APIs


zuzzas:
**What would you like to be added**:

An intermediate Machine state for non-atomic Cloud Provider APIs. After VM creation is complete, apply a second wave of changes to the machine. Consider Machine Ready only after all API calls are complete.

**Why is this needed**:

There are some API methods in some of the Cloud Provider's APIs that require performing a second (after the creation of a VM) API call to change some parameters. A perfect (and dear to my heart) example is [setting](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ModifyNetworkInterfaceAttribute.html) the `source-dest-check` option on AWS instances.

If MCM is restarted (or somehow lost internal state), the second non-atomic call to the Cloud Provider's API would never be made.

prashanth26:
Also, we should update the status on machine object right before VM creation request is made ot the provider.

prashanth26:
Hi @zuzzas ,

I thought a little deeper about the issue which you have raised. I see two separate requirements here, correct me if I understood it wrongly. 

1. Having multiple cloud provider calls for a single machine creation so that operations are continued/retried on failures (MCM restarts). This is something that should be achievable using the OOT code currently where `CreateMachine()` call could be smart enough to decode the previous state using the [lastKnownState](https://github.com/gardener/machine-controller-manager/blob/master/pkg/util/provider/driver/driver.go#L67) - This field can be used by the providers to save partial state during VM creation/deletions (also at this. And also the state of the partial state of the VMs on providers can always be retrieved if VMs are tagged correctly with the initial VM operation call. 

Although the above-mentioned logic is achievable with the OOT IMO, however one small caveat I observe is that the OOT currently would delete such partially created machines bcoz of the machineSet controller that deletes failed machines. However, the fix I see for this would be to keep the machine in a `Pending` state when machine creation is partially created here - https://github.com/gardener/machine-controller-manager/blob/master/pkg/util/provider/machinecontroller/machine_util.go#L418. We can probably also switch between `Pending` / `Failed` / `AnyOtherPhase` based on the type of error returned by the provider `CreateMachine()` [code here]( https://github.com/gardener/machine-controller-manager/blob/master/pkg/util/provider/machinecontroller/machine_util.go#L404). I am currently working on the AWS OOT provider and will try have a stable release in the next week. You can probably test it to verify your use-case to add your partially created machine logic. 

[...]

PR #504, opened 2020-08-26
Make src/dst checks configurable on awsmachineclass

### LINKED ISSUE #499: Make src/dst checks configurable on awsmachineclass
To support calico's "CrossSubnet" mode on gardener clusters for provider type `aws` the src and dst checks (https://docs.aws.amazon.com/vpc/latest/userguide/VPC_NAT_Instance.html#EIP_Disable_SrcDestCheck) have to be disabled on machine network interface level. Therefore I would like to expose a field in the `awsmachineclass` (https://github.com/gardener/machine-controller-manager/blob/master/pkg/apis/machine/v1alpha1/aws_machineclass_types.go#L179) which contains a boolean value which describes if the src/dst checks on the interface level of the machines are enabled or disabled.
I would suggest something like 
```
// Describes a network interface.
// Please also see https://docs.aws.amazon.com/goto/WebAPI/ec2-2016-11-15/MachineAWSNetworkInterfaceSpecification
type AWSNetworkInterfaceSpec struct { 
// If set to false, source and destination checks are disabled, default is true
SrcAndDstChecksEnabled: bool `json:"srcAndDstChecksEnabled,omitempty"`
}
```
WDYT?
/cc: @zanetworker

DockToFuture:
This PR makes the aws src/dst checks configurable on a machine network interface level.
To disable the the src/dst checks the `srcAndDstChecksEnabled` field has to be set to false in the `awsmachineclass`.
The default value is set to true.
```
spec:
  networkInterfaces:
  -   srcAndDstChecksEnabled: false
```
Fixes #499

```improvement operator
Src/dst checks are now configurable on awsmachineclass via `spec.networkInterfaces.srcAndDstChecksEnabled` field. The checks are enabled by default.
```

### DIFF
--- a/pkg/driver/driver_aws.go
+++ b/pkg/driver/driver_aws.go
@@ -61,9 +61,7 @@ func (d *AWSDriver) Create() (string, string, error) {
 	if err != nil {
 		return "Error", "Error", err
 	}
-
 	UserDataEnc := base64.StdEncoding.EncodeToString([]byte(d.UserData))
-
 	var imageIds []*string
 	imageID := aws.String(d.AWSMachineClass.Spec.AMI)
 	imageIds = append(imageIds, imageID)
@@ -96,7 +94,7 @@ func (d *AWSDriver) Create() (string, string, error) {
 	if err != nil {
 		return "Error", "Error", err
 	}
-
+	srcDstChecks := true
 	var networkInterfaceSpecs []*ec2.InstanceNetworkInterfaceSpecification
 	for i, netIf := range d.AWSMachineClass.Spec.NetworkInterfaces {
 		spec := &ec2.InstanceNetworkInterfaceSpecification{
@@ -113,6 +111,10 @@ func (d *AWSDriver) Create() (string, string, error) {
 		}
 
 		networkInterfaceSpecs = append(networkInterfaceSpecs, spec)
+
+		if netIf.SrcAndDstChecksEnabled != nil && !*netIf.SrcAndDstChecksEnabled {
+			srcDstChecks = false
+		}
 	}
 
 	// Specify the details of the machine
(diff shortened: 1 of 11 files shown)


pkg/driver/driver_aws.go
```diff
@@ -150,7 +152,14 @@ func (d *AWSDriver) Create() (string, string, error) {
 		return "Error", "Error", err
 	}
 	metrics.APIRequestCount.With(prometheus.Labels{"provider": "aws", "service": "ecs"}).Inc()
-
+	// disable srcDstChecks on ec2 Instance if set
+	if !srcDstChecks {
+		klog.V(2).Infof("Disabling src/dst checks for machine with instanceID: %s", *runResult.Instances[0].InstanceId)
+		_, err = svc.ModifyInstanceAttribute(&ec2.ModifyInstanceAttributeInput{InstanceId: runResult.Instances[0].InstanceId, SourceDestCheck: &ec2.AttributeBooleanValue{Value: aws.Bool(false)}})
```
hardikdr, line 158: This would be the `Best Effort` and not `Guaranteed` feature, would that be feasible for the request for now?
- Basically, a VM might not have the src/dst check disabled if MCM crashes immediately after making a CreateVM() call, doesnt get a chance for `ModifyInstanceAttr` call. 
It basically loses the internal state and adopts the existing VM in the next reconciliation It's rare but it can happen.

To resolve it, we need to fix #444 . 
Happy to go ahead with further review/merge if we think this is a feasible solution with a known shortcoming.
cc @zanetworker

DockToFuture: It's a feasible solution for now but in the long term we require the fix from #444 as you mentioned. I'm eager to see that change happen.

zanetworker: @hardikdr This is an important call and 'best effort' would not be enough IMO. New VMs without src/destination checks will fail to communicate cross zones which is a downtime that is not desired.

DockToFuture: For testing the feature it would be enough but if you want to go live with it, it wouldn't.

prashanth26: Yes, to achieve something like this we will need in-place updates or some sort of intermediate updates. This solution wouldn't work is what even i think.

prashanth26: Also, something worth mentioning would be that. 

We could probably take this PR for now. So what would happen is say machine creation fails in intermediate step (VM created but setting this flag fails). Then the machine is assumed to be created (as the VM creation succeeds, and next recconcilation assumes that VM is found) and waits for it to join the cluster in 20mins. And I am guessing that the machine won't join once we enable calico's "CrossSubnet" mode. And then machine would re-create. We could use this approach, however I wouldn't prefer this. We could use this only as a stop-gap solution until we support in-place updates.

However, my preferred approach would be to wait for the in place updates.

zanetworker: @prashanth26 The machine will probably join (access to the internet is not much affected by this change, its NATed from the node to the internet). The affected clusters are multi-zoned clusters and intra-node routes (node-to-node networking).So I would not rely on Calico failing the join procedure.

hardikdr:
Update: As discussed with @zanetworker and @DockToFuture, we'll first complete #444 and then merge this PR.
cc @prashanth26

PR #516, opened 2020-09-23
AWS: Allow deletion of VMs even on list image or modify instance failure

### LINKED ISSUE #514: Allow deletion of VM when AMI is not found in AWS.
**What would you like to be added**: At the moment, we have certain checks to default the `DeleteOnTermination` attribute on disks before actually deleting the VM. 
Codepath: https://github.com/gardener/machine-controller-manager/blob/master/pkg/driver/driver_aws.go#L283-L332 .

This can block the deletion of the VM if the underlying AMI is removed/expired. 
We should remove this codepath if not required anymore, as we allow to set `DeleteOnTermination` via MachineClass.

**Why is this needed**: MCM should allow the VM deletion if AMI is not found.

prashanth26:
Allow deletion of VMs even on list image or modify instance failure

Fixes #514 

**Release note**:

```improvement operator
AWS: Allow deletion of VMs even on list image or modify instance failure
```

### DIFF
--- a/pkg/driver/driver_aws.go
+++ b/pkg/driver/driver_aws.go
@@ -286,44 +286,30 @@ func (d *AWSDriver) Delete(machineID string) error {
 	metrics.APIRequestCount.With(prometheus.Labels{"provider": "aws", "service": "ecs"}).Inc()
 
 	if len(describeImageOutput.Images) < 1 {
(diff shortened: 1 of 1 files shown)


hardikdr (approved):
/lgtm

PR #520, opened 2020-09-23
Allow deletion of VM even on image/instance get failures.

prashanth26:
Allow deletion of VM even on image/instance get failures.

Fixes #520 

**Release note**:

```noteworthy user
Allow deletion of VM even on image/instance get failures.
```


hardikdr (approved):
/lgtm

PR #523, opened 2020-10-01
Introduce delay in re-enqueuing machines during creation--deletion failures.

hardikdr:
Co-authored-by: Prashanth <prashanth@sap.com>

**What this PR does / why we need it**:  Currently we immediately re-enqueue the machine-object on creation/deletion failure. This PR adds a constant delay before retrying the operation. Eventually, we must build exponential retry on top or parallel to this solution. 
- PR introduces a new phase `CrashLoopBackOff`. This should be used when machine creation fails, but machine-set doesn't really need to replace the machine-object[as that won't anyways help]. 

Fixes #

**Release note**:

```noteworthy operator
Introduce a backoff in re-enqueuing machines on creation/deletion failures. Avoids throttling APIServer & provider calls.
```
```improvement developer
Adds a new phase `CrashLoopBackOff` that is set due to machine creation failures. 
```

### DIFF
--- a/pkg/apis/machine/types.go
+++ b/pkg/apis/machine/types.go
@@ -222,6 +222,9 @@ const (
 
 	// MachineFailed means operation failed leading to machine status failure
 	MachineFailed MachinePhase = "Failed"
+
+	// MachineCrashLoopBackOff means creation or deletion of the machine is failing.
+	MachineCrashLoopBackOff MachinePhase = "CrashLoopBackOff"
 )
 
 // MachinePhase is a label for the condition of a machines at the current time.
--- a/pkg/apis/machine/v1alpha1/machine_types.go
+++ b/pkg/apis/machine/v1alpha1/machine_types.go
@@ -146,9 +146,12 @@ const (
 
 	// MachineFailed means operation failed leading to machine status failure
 	MachineFailed MachinePhase = "Failed"
+
+	// MachineCrashLoopBackOff means creation or deletion of the machine is failing.
+	MachineCrashLoopBackOff MachinePhase = "CrashLoopBackOff"
 )
 
-// MachinePhase is a label for the condition of a machines at the current time.
+// MachineState is a current state of the machine.
 type MachineState string
 
 // These are the valid statuses of machines.
--- a/pkg/controller/machine.go
+++ b/pkg/controller/machine.go
@@ -27,12 +27,14 @@ import (
 
 	"k8s.io/klog"
 
+	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
 
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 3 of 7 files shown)


pkg/controller/machine.go
```diff
@@ -433,7 +491,7 @@ func (c *controller) machineCreate(machine *v1alpha1.Machine, driver driver.Driv
 			LastUpdateTime: metav1.Now(),
 		}
 		currentStatus := v1alpha1.CurrentStatus{
-			Phase:          v1alpha1.MachineFailed,
+			Phase:          v1alpha1.MachineCrashLoopBackOff,
```
prashanth26, line 494: This is while deletion right? Should this be `MachineCrashLoopBackOff` or `MachineTerminating`? as otherwise the deletion starts switching the conditions between the two?

pkg/controller/machine.go
```diff
@@ -775,14 +834,18 @@ func (c *controller) updateMachineStatus(
...
 	if err != nil {
 		return machine, err
 	}
+	clone := machine.DeepCopy()
 
-	clone = clone.DeepCopy()
 	clone.Status.LastOperation = lastOperation
 	clone.Status.CurrentStatus = currentStatus
+	if c.isMachineStatusEqual(clone.Status, machine.Status) {
+		klog.V(3).Infof("Not updating the status of the machine object %q , as it is already same", clone.Name)
```
prashanth26, line 846:
```suggestion
		klog.V(4).Infof("Not updating the status of the machine object %q , as it is already same", clone.Name)
```

pkg/apis/machine/types.go
```diff
@@ -222,6 +222,9 @@ const (
 
 	// MachineFailed means operation failed leading to machine status failure
 	MachineFailed MachinePhase = "Failed"
+
+	// MachineCrashLoopBackOff means creation or deletion of the machine is failing.
```
prashanth26, line 226:
```suggestion
	// MachineCrashLoopBackOff refers to failure on machine create operation and will retry op in the next machine resync period.
```

pkg/apis/machine/v1alpha1/machine_types.go
```diff
@@ -146,9 +146,12 @@ const (
 
 	// MachineFailed means operation failed leading to machine status failure
 	MachineFailed MachinePhase = "Failed"
+
+	// MachineCrashLoopBackOff means creation or deletion of the machine is failing.
```
prashanth26, line 150:
```suggestion
	// MachineCrashLoopBackOff refers to failure on machine create operation and will retry op in the next machine resync period.
```

pkg/controller/machine.go
```diff
@@ -793,6 +856,16 @@ func (c *controller) updateMachineStatus(
 	return clone, nil
 }
 
+// isMachineStatusEqual checks if the status of 2 machines is similar or not.
+func (c *controller) isMachineStatusEqual(s1, s2 v1alpha1.MachineStatus) bool {
+	s1Copy, s2Copy := s1.DeepCopy(), s2.DeepCopy()
+	s1Copy.LastOperation.Description, s2Copy.LastOperation.Description = "", ""
+	s1Copy.LastOperation.LastUpdateTime, s2Copy.LastOperation.LastUpdateTime = metav1.Time{}, metav1.Time{}
+	s1Copy.CurrentStatus.LastUpdateTime, s2Copy.CurrentStatus.LastUpdateTime = metav1.Time{}, metav1.Time{}
```
prashanth26, line 864: Did we want to change this logic to compare the time difference between the timestamps here right?

pkg/controller/machine.go
hardikdr: This is a machineCreate method, right.

pkg/controller/machine.go
hardikdr: Hm, if the status is older than ~30mins, then update it anyways, I'll do that.

pkg/controller/machine.go
prashanth26: Yes, sorry I got confused.

hardikdr:
Update: We decided to merge these changes on master-branch first, and hotfix it here only if we see a need later.
cc @prashanth26

prashanth26 (approved):
/lgtm tested locally.

PR #525, opened 2020-10-03
Cherry-pick on master: Introduce delay in re-enqueuing machines during creation-deletion failures.

hardikdr:
Co-authored-by: Prashanth <prashanth@sap.com>

**What this PR does / why we need it**:  Currently we immediately re-enqueue the machine-object on creation/deletion failure. This PR adds a constant delay before retrying the operation. Eventually, we must build exponential retry on top or parallel to this solution. 
- PR introduces a new phase `CrashLoopBackOff`. This should be used when machine creation fails, but machine-set doesn't really need to replace the machine-object[as that won't anyways help]. 

Fixes #

**Release note**:

```noteworthy operator
Introduce a delay in re-enqueuing machines during creation/deletion failures. 
```
```improvement developer
Adds a new phase `CrashLoopBackOff` that is set due to machine creation failures. 
```

### DIFF
--- a/pkg/controller/machine.go
+++ b/pkg/controller/machine.go
@@ -27,12 +27,14 @@ import (
 
 	"k8s.io/klog"
 
+	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
 
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -48,6 +50,9 @@ const (
 	// priority the more likely it is to be deleted first
 	// Default priority for a machine is set to 3
 	MachinePriority = "machinepriority.machine.sapcloud.io"
+
+	// MachineEnqueueRetryPeriod period after which a machine object is re-enqueued upon creation or deletion failures.
+	MachineEnqueueRetryPeriod = 2 * time.Minute
 )
 
 /*
@@ -158,7 +163,8 @@ func (c *controller) reconcileClusterMachine(machine *v1alpha1.Machine) error {
 	// Validate MachineClass
 	MachineClass, secretRef, err := c.validateMachineClass(&machine.Spec.Class)
 	if err != nil || secretRef == nil {
-		return err
+		c.enqueueMachineAfter(machine, MachineEnqueueRetryPeriod)
+		return nil
 	}
 
 	driver := driver.NewDriver(machine.Spec.ProviderID, secretRef, machine.Spec.Class.Kind, MachineClass, machine.Name)
@@ -200,7 +206,8 @@ func (c *controller) reconcileClusterMachine(machine *v1alpha1.Machine) error {
 	if machine.DeletionTimestamp != nil {
 		// Processing of delete event
 		if err := c.machineDelete(machine, driver); err != nil {
-			return err
+			c.enqueueMachineAfter(machine, MachineEnqueueRetryPeriod)
+			return nil
 		}
 	} else if machine.Status.CurrentStatus.TimeoutActive {
 		// Processing machine
@@ -213,7 +220,8 @@ func (c *controller) reconcileClusterMachine(machine *v1alpha1.Machine) error {
 			return nil
 		} else if actualProviderID == "" {
 			if err := c.machineCreate(machine, driver); err != nil {
-				return err
+				c.enqueueMachineAfter(machine, MachineEnqueueRetryPeriod)
+				return nil
 			}
 		} else if actualProviderID != machine.Spec.ProviderID {
 			if err := c.machineUpdate(machine, actualProviderID); err != nil {
@@ -231,6 +239,12 @@ func (c *controller) reconcileClusterMachine(machine *v1alpha1.Machine) error {
 */
 func (c *controller) addNodeToMachine(obj interface{}) {
 
+	node := obj.(*corev1.Node)
+	if node == nil {
+		klog.Errorf("Couldn't convert to node from object")
+		return
+	}
+
 	key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
 	if err != nil {
 		klog.Errorf("Couldn't get key for object %+v: %v", obj, err)
@@ -245,8 +259,10 @@ func (c *controller) addNodeToMachine(obj interface{}) {
 		return
 	}
 
-	klog.V(4).Infof("Add machine object backing node %q", machine.Name)
-	c.enqueueMachine(machine)
+	if machine.Status.CurrentStatus.Phase != v1alpha1.MachineCrashLoopBackOff && nodeConditionsHaveChanged(machine.Status.Conditions, node.Status.Conditions) {
+		klog.V(2).Infof("Enqueue machine object %q as backing node's conditions have changed", machine.Name)
+		c.enqueueMachine(machine)
+	}
 }
 
 func (c *controller) updateNodeToMachine(oldObj, newObj interface{}) {
@@ -254,7 +270,28 @@ func (c *controller) updateNodeToMachine(oldObj, newObj interface{}) {
 }
 
 func (c *controller) deleteNodeToMachine(obj interface{}) {
-	c.addNodeToMachine(obj)
+
+	node := obj.(*corev1.Node)
+	if node == nil {
+		klog.Errorf("Couldn't convert to node from object")
+		return
+	}
+
+	key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
+	if err != nil {
+		klog.Errorf("Couldn't get key for object %+v: %v", obj, err)
+		return
+	}
+
+	machine, err := c.getMachineFromNode(key)
+	if err != nil {
+		klog.Errorf("Couldn't fetch machine %s, Error: %s", key, err)
+		return
+	} else if machine == nil {
+		return
+	}
+
+	c.enqueueMachine(machine)
 }
 
 /*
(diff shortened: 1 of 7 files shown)


pkg/controller/machine.go
```diff
@@ -245,16 +258,41 @@ func (c *controller) addNodeToMachine(obj interface{}) {
 		return
 	}
 
-	klog.V(4).Infof("Add machine object backing node %q", machine.Name)
-	c.enqueueMachine(machine)
+	if machine.Status.CurrentStatus.Phase == v1alpha1.MachineCrashLoopBackOff {
+		return
+	} else if nodeConditionsHaveChanged(machine.Status.Conditions, node.Status.Conditions) {
```
prashanth26, line 263:
```suggestion
	if machine.Status.CurrentStatus.Phase != v1alpha1.MachineCrashLoopBackOff && nodeConditionsHaveChanged(machine.Status.Conditions, node.Status.Conditions) {
```
To keep the code more readable.

pkg/controller/machine.go
```diff
@@ -793,6 +856,25 @@ func (c *controller) updateMachineStatus(
 	return clone, nil
 }
 
+// isMachineStatusEqual checks if the status of 2 machines is similar or not.
+func (c *controller) isMachineStatusEqual(s1, s2 v1alpha1.MachineStatus) bool {
```
prashanth26, line 860:
```suggestion
func isMachineStatusEqual(s1, s2 v1alpha1.MachineStatus) bool {
```

pkg/controller/machine.go
```diff
@@ -775,14 +834,18 @@ func (c *controller) updateMachineStatus(
...
+	machine, err := c.controlMachineClient.Machines(machine.Namespace).Get(machine.Name, metav1.GetOptions{})
 	if err != nil {
 		return machine, err
 	}
+	clone := machine.DeepCopy()
 
-	clone = clone.DeepCopy()
 	clone.Status.LastOperation = lastOperation
 	clone.Status.CurrentStatus = currentStatus
+	if c.isMachineStatusEqual(clone.Status, machine.Status) {
```
prashanth26, line 845:
```suggestion
	if isMachineStatusEqual(clone.Status, machine.Status) {
```

pkg/controller/machine.go
```diff
@@ -793,6 +856,25 @@ func (c *controller) updateMachineStatus(
...
+
+	if (s1Copy.LastOperation.LastUpdateTime.Time.After(time.Now().Add(tolerateTimeDiff))) || (s2Copy.LastOperation.LastUpdateTime.Time.After(time.Now().Add(tolerateTimeDiff))) {
+		return false
+	}
+	s1Copy.LastOperation.LastUpdateTime, s2Copy.LastOperation.LastUpdateTime = metav1.Time{}, metav1.Time{}
+
+	if (s1Copy.CurrentStatus.LastUpdateTime.Time.After(time.Now().Add(tolerateTimeDiff))) || (s2Copy.CurrentStatus.LastUpdateTime.Time.After(time.Now().Add(tolerateTimeDiff))) {
+		return false
+	}
+	s1Copy.CurrentStatus.LastUpdateTime, s2Copy.CurrentStatus.LastUpdateTime = metav1.Time{}, metav1.Time{}
```
prashanth26, line 873: I don't understand this part. Why is this `if` block repeated twice? Also, is it possible to make the difference in timestamps using a single comparison instead of two?

hardikdr: One if block is for `LastOperation`, and the other is for `CurrentStatus`.

pkg/controller/machine_test.go
```diff
@@ -362,6 +387,108 @@ var _ = Describe("machine", func() {
...
+
+		DescribeTable("##table",
+			func(data *data) {
+				stop := make(chan struct{})
+				defer close(stop)
+
+				c, trackers := createController(stop, testNamespace, nil, nil, nil)
+				defer trackers.Stop()
+
+				equal := c.isMachineStatusEqual(data.action.s1, data.action.s2)
```
prashanth26, line 423:
```suggestion
				equal := isMachineStatusEqual(data.action.s1, data.action.s2)
```
Tests are failing currently.

hardikdr: Yes, the suggested changes seem to have broken both `check` and `test` step, can you please check.

pkg/controller/machine.go
```diff
@@ -48,6 +50,9 @@ const (
 	// priority the more likely it is to be deleted first
 	// Default priority for a machine is set to 3
 	MachinePriority = "machinepriority.machine.sapcloud.io"
+
+	// MachineEnqueueRetryPeriod period after which a machine object is re-enqueued upon creation or deletion failures.
+	MachineEnqueueRetryPeriod = 3 * time.Minute
```
prashanth26, line 55:
```suggestion
	MachineEnqueueRetryPeriod = 2 * time.Minute
```

pkg/controller/machine.go
```diff
@@ -775,14 +832,18 @@ func (c *controller) updateMachineStatus(
 	currentStatus v1alpha1.CurrentStatus,
 ) (*v1alpha1.Machine, error) {
 	// Get the latest version of the machine so that we can avoid conflicts
-	clone, err := c.controlMachineClient.Machines(machine.Namespace).Get(machine.Name, metav1.GetOptions{})
+	machine, err := c.controlMachineClient.Machines(machine.Namespace).Get(machine.Name, metav1.GetOptions{})
 	if err != nil {
 		return machine, err
 	}
+	clone := machine.DeepCopy()
```
prashanth26, line 839:
```suggestion
	latestMachine, err := c.controlMachineClient.Machines(machine.Namespace).Get(machine.Name, metav1.GetOptions{})
	if err != nil {
		return machine, err
	}
	clone := latestMachine.DeepCopy()
```
We need to have the old value of the `machine` here as there is a call below to retry with `machine` object and if this call fails, the value of machine becomes `nil` and would keep failing.

hardikdr: I'd guess, the error should be returned along with that, and should be taken care.
But a good point, to have diff name than argument as we had discussed, as a pointer may mess with the original flow.

prashanth26: Yes, i think even you are right eventually the call but would through. But i guess better safe :D

pkg/controller/machine.go
```diff
@@ -775,14 +832,18 @@ func (c *controller) updateMachineStatus(
 	currentStatus v1alpha1.CurrentStatus,
 ) (*v1alpha1.Machine, error) {
 	// Get the latest version of the machine so that we can avoid conflicts
-	clone, err := c.controlMachineClient.Machines(machine.Namespace).Get(machine.Name, metav1.GetOptions{})
+	latestMachine, err := c.controlMachineClient.Machines(machine.Namespace).Get(machine.Name, metav1.GetOptions{})
 	if err != nil {
 		return machine, err
```
prashanth26, line 837:
```suggestion
		return nil, err
```
Fixes the issues with the tests.

prashanth26 (approved):
/lgtm on squashing the changes. 

Tested the changes locally, looks good to me.

prashanth26 (approved):

pkg/controller/machine.go
```diff
@@ -245,16 +259,39 @@ func (c *controller) addNodeToMachine(obj interface{}) {
 		return
 	}
 
-	klog.V(4).Infof("Add machine object backing node %q", machine.Name)
-	c.enqueueMachine(machine)
+	if machine.Status.CurrentStatus.Phase != v1alpha1.MachineCrashLoopBackOff && nodeConditionsHaveChanged(machine.Status.Conditions, node.Status.Conditions) {
```
ggaurav10, line 262: `addNodeToMachine()` is also called when a node joins. If a new machine has been in crash loop and then becomes healthy, we will not enqueue the machine immediately when the node joins because of this check, right?

pkg/controller/machine.go
```diff
@@ -245,16 +259,39 @@ func (c *controller) addNodeToMachine(obj interface{}) {
...
 }
 
 func (c *controller) deleteNodeToMachine(obj interface{}) {
-	c.addNodeToMachine(obj)
+
+	node := obj.(*corev1.Node)
+	if node == nil {
+		klog.Errorf("Couldn't convert to node %+v from object", node.Name)
+		return
+	}
```
ggaurav10, line 278: same as above

pkg/controller/machine.go
```diff
@@ -793,6 +855,25 @@ func (c *controller) updateMachineStatus(
 	return clone, nil
 }
 
+// isMachineStatusEqual checks if the status of 2 machines is similar or not.
+func isMachineStatusEqual(s1, s2 v1alpha1.MachineStatus) bool {
+	tolerateTimeDiff := 30 * time.Minute
+	s1Copy, s2Copy := s1.DeepCopy(), s2.DeepCopy()
+	s1Copy.LastOperation.Description, s2Copy.LastOperation.Description = "", ""
+
+	if (s1Copy.LastOperation.LastUpdateTime.Time.After(time.Now().Add(tolerateTimeDiff))) || (s2Copy.LastOperation.LastUpdateTime.Time.After(time.Now().Add(tolerateTimeDiff))) {
```
ggaurav10, line 864: `time.Now().Add(tolerateTimeDiff)` is 30 minutes into the future. Should it be `Sub()` instead?

pkg/controller/machine.go
```diff
@@ -793,6 +855,25 @@ func (c *controller) updateMachineStatus(
...
+		return false
+	}
+	s1Copy.LastOperation.LastUpdateTime, s2Copy.LastOperation.LastUpdateTime = metav1.Time{}, metav1.Time{}
+
+	if (s1Copy.CurrentStatus.LastUpdateTime.Time.After(time.Now().Add(tolerateTimeDiff))) || (s2Copy.CurrentStatus.LastUpdateTime.Time.After(time.Now().Add(tolerateTimeDiff))) {
```
ggaurav10, line 869: same as above

ggaurav10 (changes requested):

pkg/controller/machine.go
hardikdr: Hm, I think it's safe to simply remove the check for `MachineCrashLoopBackOff` here, as it's unlikely to have node-object when creation itself fails.

pkg/controller/machine.go
hardikdr: +1

pkg/controller/machine.go
hardikdr: That's a good catch.

pkg/controller/machine.go
ggaurav10: it is also called from `updateNodeToMachine()`. Still it should be safe to remove the check

prashanth26: Yes, you need this check else the `updateNodeToMachine()` has to be modified. Also during node joining the nodeConditions do change as machineStatus conditions are initially nil.

pkg/controller/machine.go
prashanth26: This is why we need @ggaurav10 's eyes

pkg/controller/machine.go
hardikdr: @ggaurav10 @prashanth26 I made the changes, can you please take another look.

ggaurav10 (approved):

prashanth26 (approved):
/lgtm

PR #526, opened 2020-10-05
Bugfix: Continue of VM deletion on DescribeImage call failure

prashanth26:
VM deletion was failing is DescribeImage call was failing in case of a malformed AMI ID. This PR fixes it. 

Fixes #

**Release note**:

```improvement operator
AWS: Continue of VM deletion on DescribeImage call failure
```
/assign @hardikdr

### DIFF
--- a/pkg/driver/driver_aws.go
+++ b/pkg/driver/driver_aws.go
@@ -286,9 +286,10 @@ func (d *AWSDriver) Delete(machineID string) error {
 	describeImageOutput, err := svc.DescribeImages(&describeImagesRequest)
 	if err != nil {
 		metrics.APIFailedRequestCount.With(prometheus.Labels{"provider": "aws", "service": "ecs"}).Inc()
-		return err
+		klog.Warningf("DescribeImages call failed for %s. MachineID - %q", *imageID, machineID)
+	} else {
+		metrics.APIRequestCount.With(prometheus.Labels{"provider": "aws", "service": "ecs"}).Inc()
 	}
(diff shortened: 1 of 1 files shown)


pkg/driver/driver_aws.go
```diff
@@ -286,9 +286,10 @@ func (d *AWSDriver) Delete(machineID string) error {
 	describeImageOutput, err := svc.DescribeImages(&describeImagesRequest)
 	if err != nil {
 		metrics.APIFailedRequestCount.With(prometheus.Labels{"provider": "aws", "service": "ecs"}).Inc()
-		return err
+		klog.Warningf("DescribeImages call failed for %s. MachineID - %q", *imageID, machineID)
```
hardikdr, line 289: This may end up being a too generic solution. Hypothetically, if the account is rate-limited, the describe call may fail, but the next calls may get a chance, and deleting the machine? 
We could probably check the doku of AWS SDK, and understand what all errors are possible for an AMI, and by-pass specific errors like `MalformedAMI`.

prashanth26: Yes, make sense. But handling malformed AMI deletion is something we can ignore for now in that case? Bcoz I don't think machines would be created with the right AMI and then updated to the wrong one. And even if it is, updating to the right one would fix the issue. And hopefully, we can ignore this fix for now. As anyways AWS throws the right error. 

Just wondering about cases where the users don't have permissions to update Machines then we would have to handle each AWS error differently. Let's rather stick to the current solution and continue with what's working?

hardikdr: Sure, agree.

PR #531, opened 2020-10-14
Bugfix: Fixed issues with machineClass finalizers 

prashanth26:
Without proper finalizers, machine classes backing machines were lost on deletions. 

Fixes #

This partially reverts this change - https://github.com/gardener/machine-controller-manager/commit/0a568917461a5c6385974322784b36d62762e60a#diff-a327d9bce2e5949213510b45b329fa9d45d2dedfcce11e407e51e876549f265bL130-L137

**Release note**:

```improvement operator
Added a more comprehensive set of events to trigger machine class reconciliations.
```
```improvement operator
Finalizers are added by default for all machine class objects.
```

### DIFF
--- a/pkg/controller/alicloudmachineclass.go
+++ b/pkg/controller/alicloudmachineclass.go
@@ -135,6 +135,14 @@ func (c *controller) reconcileClusterAlicloudMachineClass(class *v1alpha1.Aliclo
 		return nil
 	}
 
+	// Add finalizer to avoid lossing machineClass object
+	if class.DeletionTimestamp == nil {
+		err = c.addAlicloudMachineClassFinalizers(class)
+		if err != nil {
+			return err
+		}
+	}
+
 	machines, err := c.findMachinesForClass(AlicloudMachineClassKind, class.Name)
 	if err != nil {
 		return err
@@ -144,13 +152,7 @@ func (c *controller) reconcileClusterAlicloudMachineClass(class *v1alpha1.Aliclo
 		// If deletion timestamp doesn't exist
 		_, annotationPresent := class.Annotations[machineutils.MigratedMachineClass]
 
-		if len(machines) > 0 {
-			// If 1 or more machine objects are referring the machineClass
-			err = c.addAlicloudMachineClassFinalizers(class)
-			if err != nil {
-				return err
-			}
-		} else if c.deleteMigratedMachineClass && annotationPresent {
+		if c.deleteMigratedMachineClass && annotationPresent && len(machines) == 0 {
 			// If controller has deleteMigratedMachineClass flag set
 			// and the migratedMachineClass annotation is set
 			err = c.controlMachineClient.AlicloudMachineClasses(class.Namespace).Delete(class.Name, &metav1.DeleteOptions{})
--- a/pkg/controller/alicloudmachineclass.go
+++ b/pkg/controller/alicloudmachineclass.go
@@ -68,8 +68,33 @@ func (c *controller) machineToAlicloudMachineClassAdd(obj interface{}) {
 }
 
 func (c *controller) machineToAlicloudMachineClassUpdate(oldObj, newObj interface{}) {
-	c.machineToAlicloudMachineClassAdd(oldObj)
-	c.machineToAlicloudMachineClassAdd(newObj)
+	oldMachine, ok := oldObj.(*v1alpha1.Machine)
+	if oldMachine == nil || !ok {
+		klog.Warningf("Couldn't get machine from object: %+v", oldObj)
+		return
+	}
+	newMachine, ok := newObj.(*v1alpha1.Machine)
+	if newMachine == nil || !ok {
+		klog.Warningf("Couldn't get machine from object: %+v", newObj)
+		return
+	}
+
+	if oldMachine.Spec.Class.Kind == newMachine.Spec.Class.Kind {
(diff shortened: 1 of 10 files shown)


hardikdr (approved):
Looks good, thanks for the PR. 
/lgtm

pkg/controller/alicloudmachineclass.go
```diff
@@ -67,6 +67,15 @@ func (c *controller) machineToAlicloudMachineClassDelete(obj interface{}) {
 	}
 }
 
+func (c *controller) machineToAlicloudMachineClassUpdate(oldObj, newObj interface{}) {
+	c.machineToAlicloudMachineClassAdd(oldObj)
+	c.machineToAlicloudMachineClassAdd(newObj)
```
ggaurav10, line 72: why would we want to reconcile both older and newer versions? 🤔

prashanth26: This function only receives the machine object, and we need to enqueue the machineClass object. And if the `kind` of the machineClass has changed (possible with migration) we might need to check both objects. 

Refer - https://github.com/gardener/machine-controller-manager/blob/ce211984c40181b626ee7e5417e3e2f2aa6dab97/pkg/controller/alicloudmachineclass.go#L66.

pkg/controller/alicloudmachineclass.go
```diff
@@ -135,6 +148,14 @@ func (c *controller) reconcileClusterAlicloudMachineClass(class *v1alpha1.Aliclo
 		return nil
 	}
 
+	// Add finalizer to avoid lossing machineClass object
```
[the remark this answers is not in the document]
prashanth26: Fixed it.

pkg/controller/alicloudmachineclass.go
ggaurav10: How about the following:

```
if new.Spec.Class.Kind ==  old.Spec.Class.Kind == Ali {
         c.machineToAlicloudMachineClassAdd(newObj)
} else {
         if new.Spec.Class.Kind == Ali {
              c.machineToAlicloudMachineClassAdd(newObj)
         } else old.Spec.Class.Kind == Ali {
              c.machineToAlicloudMachineClassAdd(oldObj)
         }
}
```
This would prevent unnecessary double operations in non-migration scenarios which is majority of the times. WDYT?

prashanth26: Good suggestion @ggaurav10 . How about this code keeping your logic in mind?

```suggestion
	oldMachine, ok := oldObj.(*v1alpha1.Machine)
	if oldMachine == nil || !ok {
		klog.Warningf("Couldn't get machine from object: %+v", oldObj)
		return
	}
	newMachine, ok := oldObj.(*v1alpha1.Machine)
	if newMachine == nil || !ok {
		klog.Warningf("Couldn't get machine from object: %+v", newObj)
		return
	}

	if oldMachine.Spec.Class.Kind == newMachine.Spec.Class.Kind {
		if newMachine.Spec.Class.Kind == AWSMachineClassKind {
			// Both old and new machine refer to the same machineClass object
			// And the correct kind so enqueuing only one of them.
			c.awsMachineClassQueue.Add(newMachine.Spec.Class.Name)
		}
	} else {
		// If both are pointing to different machineClasses
		// we might have to enqueue both.
		if oldMachine.Spec.Class.Kind == AWSMachineClassKind {
			c.awsMachineClassQueue.Add(oldMachine.Spec.Class.Name)
		}
		if newMachine.Spec.Class.Kind == AWSMachineClassKind {
			c.awsMachineClassQueue.Add(newMachine.Spec.Class.Name)
		}
	}
```

ggaurav10: lgtm 👍

prashanth26: And of course will replace the AWSMachineClassKind with class kind and queue with class queue

prashanth26: Have made the suggested change here - https://github.com/gardener/machine-controller-manager/pull/531/commits/91db3fff2956dc2e0a56c52dc42ce8998b1330c4.  PTAL. 

cc: @hardikdr

hardikdr:
```
newMachine, ok := oldObj.(*v1alpha1.Machine)
```
In the suggestion, above, shouldn't it be newObj ?

hardikdr: And you mean ClassKind above in the suggestion right, confirming though you already mentioned:
```
		if newMachine.Spec.Class.Kind == AWSMachineClassKind {
			c.awsMachineClassQueue.Add(newMachine.Spec.Class.Name)
		}
```

prashanth26: >In the suggestion, above, shouldn't it be newObj ?
Yes

prashanth26: Fixed it here - https://github.com/gardener/machine-controller-manager/pull/531/commits/e6ce7d31a660a9ba10973718c5b4c0d7aceb5686#diff-e37e335b3785e3c4f506cae9bde4ff92c2d1c968e78b9dba2a4bdf0cad3b228aR76.

ggaurav10 (approved):

hardikdr (approved):

PR #538, opened 2020-10-21
Added a new Creating phase

### LINKED ISSUE #444: Add an intermediate Machine state for non-atomic Cloud Provider APIs

zuzzas:

Adds a new "Creating" phase, which ensures correct VM creation when using non-atomic cloud APIs.

Band-aid for #444 

**Release note**:

```improvement operator
Added a new "Creating" Phase. Machines transition to this phase immediately upon entering the creation flow. If the Machine is already in the "Creating" phase, and it enters the creation flow again, that means that the Machine was not fully created on the cloud provider, and we move it to the Failed state. 
```

### DIFF
--- a/pkg/apis/machine/v1alpha1/machine_types.go
+++ b/pkg/apis/machine/v1alpha1/machine_types.go
@@ -149,6 +149,9 @@ const (
 
 	// MachineCrashLoopBackOff means creation or deletion of the machine is failing.
 	MachineCrashLoopBackOff MachinePhase = "CrashLoopBackOff"
+
+	// MachineCrashLoopBackOff means creation or deletion of the machine is failing. TODO
+	MachineCreating MachinePhase = "Creating"
 )
 
 // MachineState is a current state of the machine.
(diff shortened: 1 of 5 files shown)


prashanth26:
Hi @zuzzas ,

Thanks for the PR. I looked through your PR but wasn't able to relate how this PR helps you with achieving this goal of having multiple API calls for VM creations as you mentioned your requirement here - https://github.com/gardener/machine-controller-manager/issues/444#issue-599557527. Also would like to shift this discussion here on the issue.

zuzzas:
We shouldn't make changes to the in-tree controller. It is deprecated.
https://github.com/gardener/machine-controller-manager/issues/444#issuecomment-717033335

PR #545, opened 2020-10-27
Deploy VMs into existing OpenStack Neutron network

### LINKED ISSUE #535: OpenStack: add support for subnet specifying

Currently it is only possible to specify the VM network. However a network may contain a list of subnets.

**Why is this needed**:

An end user should be able to specify a subnet, where worker nodes must reside.

**Notes**:

This feature would be a little bit tricky, because nova API doesn't support a subnet in API spec. Thus a neutron port in a specified subnet must be created, then referenced in nova API spec. Also when a VM is deleted, a custom neutron port must be explicitly deleted, since a VM deletion doesn't delete a custom neutron port.

In addition, an end user should be able to create neutron ports in advance, and reuse them for VMs. This should allow to preserve private IPs list for a shoot cluster. In this case a custom neutron port must not be deleted during the VM deletion.

cc @vpnachev  @zuzzas

MrBatschner:
Adds the ability to specify an already existing OpenStack Neutron network in the subnetID of an OpenStackMachineClass. MCM will deploy new machines into the given subnet by pre-allocating Neutron ports and pass them to the Nova server object. Pre-allocated ports will be specifically removed upon server termination. This is required to specify an existing subnet in a shoot manifest. 
With this PR it is only possible to specify an already existing network, not a list of otherwise pre-allocated ports.

Partly fixes #535 

This change is a necessary precondition for [PR 169](https://github.com/gardener/gardener-extension-provider-openstack/pull/169) in gardener-extension-provider-openstack.

**Release note**:
```improvement user
Adds the ability to specify an already existing OpenStack Neutron network in the subnetID of an OpenStackMachineClass. MCM will deploy new machines into the given subnet by pre-allocating Neutron ports and pass them to the Nova server object. 
```

### DIFF
--- a/pkg/driver/driver_openstack.go
+++ b/pkg/driver/driver_openstack.go
@@ -27,7 +27,9 @@ import (
 
 	"github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
 	"github.com/gardener/machine-controller-manager/pkg/metrics"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/groups"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
 	"github.com/prometheus/client_golang/prometheus"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/klog"
@@ -112,6 +114,7 @@ func (d *OpenStackDriver) Create() (string, string, error) {
 	imageName := d.OpenStackMachineClass.Spec.ImageName
 	imageID := d.OpenStackMachineClass.Spec.ImageID
 	networkID := d.OpenStackMachineClass.Spec.NetworkID
+	subnetID := d.OpenStackMachineClass.Spec.SubnetID
 	specNetworks := d.OpenStackMachineClass.Spec.Networks
 	securityGroups := d.OpenStackMachineClass.Spec.SecurityGroups
 	availabilityZone := d.OpenStackMachineClass.Spec.AvailabilityZone
@@ -142,9 +145,49 @@ func (d *OpenStackDriver) Create() (string, string, error) {
 	var serverNetworks []servers.Network
 	var podNetworkIds = make(map[string]struct{})
 
+	// network port allocation in existing network if a networkID is specified
 	if len(networkID) > 0 {
-		serverNetworks = append(serverNetworks, servers.Network{UUID: networkID})
+		klog.V(3).Infof("existing network %q specified, need to pre-allocate ports ", networkID)
+
+		// create port in given subnet
+		if subnetID != nil && len(*subnetID) > 0 {
+			if _, err := subnets.Get(nwClient, *subnetID).Extract(); err != nil {
+				metrics.APIFailedRequestCount.With(prometheus.Labels{"provider": "openstack", "service": "neutron"}).Inc()
+				return "", "", fmt.Errorf("failed to get subnet information for subnetID %s: %s", *subnetID, err)
+			}
+
+			klog.V(3).Infof("creating port in subnet %s", *subnetID)
+
+			var securityGroupIDs []string
+			for _, securityGroup := range securityGroups {
+				securityGroupID, err := groups.IDFromName(nwClient, securityGroup)
+				if err != nil {
+					metrics.APIFailedRequestCount.With(prometheus.Labels{"provider": "openstack", "service": "neutron"}).Inc()
+					return "", "", fmt.Errorf("failed to get ID for security group %q: %s", securityGroup, err)
+				}
+				securityGroupIDs = append(securityGroupIDs, securityGroupID)
+			}
+
+			port, err := ports.Create(nwClient, &ports.CreateOpts{
+				Name:                d.MachineName,
+				NetworkID:           networkID,
+				FixedIPs:            []ports.IP{{SubnetID: *subnetID}},
+				AllowedAddressPairs: []ports.AddressPair{{IPAddress: podNetworkCidr}},
+				SecurityGroups:      &securityGroupIDs,
+			}).Extract()
+			if err != nil {
+				metrics.APIFailedRequestCount.With(prometheus.Labels{"provider": "openstack", "service": "neutron"}).Inc()
+				return "", "", fmt.Errorf("failed to create port in subnet with subnetID %s: %s", *subnetID, err)
+			}
+
+			klog.V(3).Infof("port with ID %s successfully created", port.ID)
+
+			serverNetworks = append(serverNetworks, servers.Network{UUID: networkID, Port: port.ID})
+		} else {
+			serverNetworks = append(serverNetworks, servers.Network{UUID: networkID})
+		}
 		podNetworkIds[networkID] = struct{}{}
+
 	} else {
 		for _, network := range specNetworks {
 			var resolvedNetworkID string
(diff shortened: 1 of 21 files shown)


dkistner (commented):
Had a brief look and seems to look good.
I need a little bit more time to play around with it.

dkistner (changes requested):
I guess we need to make sure that the port gets also deleted when the machine/server is already gone. Currently the port deletion logic is not executed when the machine is already deleted. We need to ensure that we check within each machine deletion flow for existing ports. See here: https://github.com/gardener/machine-controller-manager/blob/ec67a4354831540bf5553e2559ce98a7aa3aa772/pkg/driver/driver_openstack.go#L316-L324

MrBatschner:
> I guess we need to make sure that the port gets also deleted when the machine/server is already gone. Currently the port deletion logic is not executed when the machine is already deleted. We need to ensure that we check within each machine deletion flow for existing ports. 

True, we just pushed a commit that puts the port delete logic into a separate function which will always get called - even if the VM no longer exists.

**On another note:**
This whole PR means that MCM will (for new shoots created with our proposed changes to the openstack-provider-extension) always go through the _create port->create VM with port_ and the _delete VM -> delete its port_ logic. The old code-path _create VM and let OpenStack do the port stuff itself_ is still in the MCM but once the openstack-provider-extension starts providing the subnetID in the OpenStackMachineClass, this code will no longer be triggered.

I do not see a problem with this behaviour but what do you think about it?

dkistner (approved):
looks good to me.

/invite @kon-angelo 

@kon-angelo Could you please have a look as well?
@kayrus Could you also please have a brief look if you have time :)

pkg/driver/driver_openstack.go
```diff
@@ -143,8 +146,50 @@ func (d *OpenStackDriver) Create() (string, string, error) {
 	var podNetworkIds = make(map[string]struct{})
 
 	if len(networkID) > 0 {
-		serverNetworks = append(serverNetworks, servers.Network{UUID: networkID})
+		// network port dealing if subnetID is specified
+
+		klog.V(3).Infof("existing network %s specified, need to pre-allocate ports ", networkID)
```
kon-angelo, line 151: I think the comment here is misplaced. Shouldn't it be inside the `if` below.

pkg/driver/driver_openstack.go
```diff
@@ -285,21 +334,28 @@ func (d *OpenStackDriver) Delete(machineID string) error {
...
-		// waiting for the machine to be deleted to release consumed quota resources, 5 minutes should be enough
-		err = waitForStatus(client, machineID, nil, []string{"DELETED", "SOFT_DELETED"}, 300)
-		if err != nil {
-			return fmt.Errorf("error waiting for the %q server to be deleted: %s", machineID, err)
-		}
-		metrics.APIRequestCount.With(prometheus.Labels{"provider": "openstack", "service": "nova"}).Inc()
-		klog.V(3).Infof("Deleted machine with ID: %s", machineID)
-	} else {
+	err = servers.Delete(client, instanceID).ExtractErr()
+	if err != nil && isNotFoundError(err) == false {
```
kon-angelo, line 338: just a suggestion for consistency `if err != nil && !isNotFoundError(err)`

prashanth26 (approved):
/lgtm 
/needs second-opinion

pkg/driver/driver_openstack.go
```diff
@@ -143,8 +146,50 @@ func (d *OpenStackDriver) Create() (string, string, error) {
...
 
 	if len(networkID) > 0 {
-		serverNetworks = append(serverNetworks, servers.Network{UUID: networkID})
+		// network port dealing if subnetID is specified
+
+		klog.V(3).Infof("existing network %s specified, need to pre-allocate ports ", networkID)
+
+		if subnetID != nil && len(*subnetID) > 0 {
+			// create port in given subnet
+			_, err := subnets.Get(nwClient, *subnetID).Extract()
```
kon-angelo, line 155:
```go			
if _, err := subnets.Get(nwClient, *subnetID).Extract(); err != nil {
```

pkg/driver/driver_openstack.go
```diff
@@ -276,6 +320,11 @@ func (d *OpenStackDriver) Delete(machineID string) error {
 	} else if len(res) == 0 {
 		// No running instance exists with the given machine-ID
 		klog.V(2).Infof("No VM matching the machine-ID found on the provider %q", machineID)
+		err = d.deletePort()
+		if err != nil {
+			return err
+		}
+
```
kon-angelo, line 327: Another suggestion is to revert the conditional:

```go
if len(res) > 0 {
<do server delete>
}

deletePort()
```

Because it seems that you need to call deletePort in all cases anyway (whether or not the machine was found)

pkg/driver/driver_openstack.go
```diff
@@ -143,8 +146,50 @@ func (d *OpenStackDriver) Create() (string, string, error) {
...
+		// network port dealing if subnetID is specified
+
+		klog.V(3).Infof("existing network %s specified, need to pre-allocate ports ", networkID)
+
+		if subnetID != nil && len(*subnetID) > 0 {
+			// create port in given subnet
+			_, err := subnets.Get(nwClient, *subnetID).Extract()
+			if err != nil {
+				metrics.APIFailedRequestCount.With(prometheus.Labels{"provider": "openstack", "service": "neutron"}).Inc()
+				return "", "", fmt.Errorf("failed to get subnet information for subnetID %s: %s", *subnetID, err)
```
kon-angelo, line 158:
```go
				return "", "", fmt.Errorf("failed to get subnet information for subnetID %q: %s", *subnetID, err)
```
Personal preference but when it comes to names `%q` is a bit better. This just adds quotes around the string.

kon-angelo: Ditto for other places

MrBatschner: Used %q for names but left %s for uids.

pkg/driver/driver_openstack.go
MrBatschner: True, calling deletePort() twice is ugly. Changed the function to always call deletePort() as the last step and put the server delete login into the conditional as suggested.

pkg/driver/driver_openstack.go
MrBatschner: Ok, done. I believe this is preferred because it limits the visibility of the err variable, right?

pkg/driver/driver_openstack.go
MrBatschner: Corrected in both places.

pkg/driver/driver_openstack.go
MrBatschner: Changed the comment to make more sense and put it _above_ the `if`-statement.

kon-angelo (approved):
/lgtm

hardikdr:
cc @zuzzas

hardikdr:
Hm, earlier the integration test of freeze failed.
```
Check for freezing and unfreezing of machine-set due to rapid scaleUp and scaleDown	Failed: Freezing of machineSet failed. Exiting Test to avoid further conflicts.
```

And now the unit test related to the freeze failed:
```
Summarizing 1 Failure:
13:04:16

13:04:16
[Fail] #machine_safety ##reconcileClusterMachineSafetyOvershooting [It] Freeze a machineSet and its machineDeployment 
13:04:16
/tmp/build/a94a8fe5/pull-request-gardener_machine-controller-manager-pr_master/pkg/controller/machine_safety_test.go:247
```

Tests are flaky somehow.

hardikdr:
@MrBatschner Can you please squash the commits?

MrBatschner:
@hardikdr Done, squashed all 11 commits into one.

hardikdr (approved):
/lgtm
Let's wait for a day, in case @kayrus is willing to take a look.

PR #548, opened 2020-11-01
OOT: Fixes drain timeout issues on retires

### LINKED ISSUE #547: OOT:  Drain failure cases don't seem to respect drain timeouts
**What happened**:
In OOT, drain failure cases don't seem to respect drain timeouts. 

Try to terminate a VM with non-terminating pods and the VM deletion fails even after drain timeout. 

And this is because the force deletion timeout computation was messed up. 

**Environment**:
/platform gcp

prashanth26:
The OOT implementation reverted the logic that was required to track time between drain retries. This PR fixes it. Refer previous code here - https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/machine.go#L741.

Fixes #547 

The PR also contains minor nits. 

**Release note**:

```improvement operator
OOT: Fixes drain timeout issues on retires
```

### DIFF
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -145,20 +145,6 @@ func (c *controller) reconcileClusterMachine(machine *v1alpha1.Machine) (machine
 		return retry, err
 	}
 
-	/*
-		NOT NEEDED?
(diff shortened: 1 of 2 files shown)


hardikdr (approved):
Thanks for the PR.
/lgtm

PR #550, opened 2020-11-03
Adapt integration tests to handle possibly orphaned resources

### LINKED ISSUE #498: Improve CI to prevent any orphaned resources
Improve the CI script to prevent any orphaned resources

**Why is this needed**:
We have in the past observed orphaned volumes on the deletion of machines due to new API changes. Shall regressions must be checked by double-checking on infra resources (VMs. Disks, NICs) before CI tests terminate.

hardikdr:
**What this PR does / why we need it**: Adapts the integration test script to fail if there are any leftover orphan resources after the tests.

Fixes #498 

**Release note**:

```improvement operator
Adapted integration tests to handle possibly orphaned resources.
```

### DIFF
--- a/.ci/integration-tests
+++ b/.ci/integration-tests
@@ -115,6 +115,7 @@ function setup_environment() {
         
         gcloud auth activate-service-account --key-file servicekey.json
         gcloud config set project "$ProjectID"
+        gcloud config set compute/zone europe-west1-b
         gcloud config set disable_usage_reporting True
 
         printf "Successfully setup gcloud-cli\n"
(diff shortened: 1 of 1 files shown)


.ci/integration-tests
```diff
@@ -224,6 +261,37 @@ function hf_check_ms_freeze() {
     fi
 }
 
+function hf_orphan_resources() {
+    if [ $1 == "aws" ]; then
+        # Check if there are any VMs matching the tag exists.
```
prashanth26, line 266: Does it make sense to check for orphan NICs also here?

.ci/integration-tests
```diff
@@ -288,6 +356,19 @@ function tc_machine_deployment() {
     printf "\nCompleted TestCase\n"
 }
 
+function tc_check_orphan_resources() {
+    
+    printf "\nStarting TestCase: %s" "${FUNCNAME[0]}"
+     
+    # Check if there are any leftover resources.
+    if ! hf_orphan_resources "$provider"; then
+        printf "\n\t Failed: There are orphan resources in the cloud. Exiting the test."
```
prashanth26, line 365: Also, should we delete these orphan resources if/when we find them?

hardikdr: Sure, we can do that, I'll adapt it.

hardikdr: done - https://github.com/gardener/machine-controller-manager/pull/550/commits/5c9b6d9bfbe1a621e7c459774efa8f830f280adf

.ci/integration-tests
hardikdr: As discussed skipping the nic part.

prashanth26 (approved):

PR #552, opened 2020-11-04
OOT: Drain call is now invoked with the effective timeout

prashanth26:
Drain call was taking in the default drain timeout of 12hours. This change reverts the value to the effective value.

The drain call was made with the default timeout of 12hours. This PR reverts this change to use the effective timeout value to 2hours. This also makes the default drain timeout to 2hours.  

Fixes #

**Release note**:

```noteworthy user
Reduces default drain timeout from 12hours to 2hours.   
```
```improvement operator
OOT: Drain call was taking in the default drain timeout of 12hours. This change reverts the value to the effective value.
```

### DIFF
--- a/pkg/util/provider/machinecontroller/machine_util.go
+++ b/pkg/util/provider/machinecontroller/machine_util.go
@@ -911,7 +911,7 @@ func (c *controller) drainNode(deleteMachineRequest *driver.DeleteMachineRequest
 
 			drainOptions := drain.NewDrainOptions(
 				c.targetCoreClient,
-				timeOutDuration,
+				timeOut,
 				maxEvictRetries,
 				pvDetachTimeOut,
 				nodeName,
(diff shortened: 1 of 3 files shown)


pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -911,7 +911,7 @@ func (c *controller) drainNode(deleteMachineRequest *driver.DeleteMachineRequest
 
 			drainOptions := drain.NewDrainOptions(
 				c.targetCoreClient,
-				timeOutDuration,
+				timeOut,
```
hardikdr, line 914: Is it that MCM restarts can reset the drainTimeout? I hope there are no side-effects from it.
Does this also need to be fixed in-tree then? https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/machine.go#L708 ?

prashanth26: Oops. Nice catch. Wait if it's working as expected with in-tree why the difference in OOT. I guess the issue is something else then. Let me re-check.

prashanth26:
It seems like something else is causing the issue. Let me check it out. I will try to write a test to reproduce it. 

My bad above was the wrong RCA.

prashanth26:
Closing in favor of https://github.com/gardener/machine-controller-manager-provider-gcp/issues/9.

PR #554, opened 2020-11-04
Fixed regression (OOT) with maxEvictRetries and defaulted drain timeout to 2hours

prashanth26:
- The maxEvictRetries parameter was being computed wrongly with the OOT leading to longer (12hours) drain timeouts. Now this value is computed as it is with in-tree. Refer -[in-tree code]( https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/machine.go#L629). 
- The default drainTimeout value has been updated from 12hours to 2hours. 

Fixes https://github.com/gardener/machine-controller-manager-provider-gcp/issues/9

I have tested this PR and is in line with the in-tree implementation now. 

**Release note**:

```improvement user
The default drainTimeout value has been updated from 12hours to 2hours. 
```
```improvement user
OOT: Fixed regression with maxEvictRetries
```

### DIFF
--- a/pkg/controller/drain.go
+++ b/pkg/controller/drain.go
@@ -85,7 +85,7 @@ const (
 	EvictionSubresource = "pods/eviction"
 
 	// DefaultMachineDrainTimeout is the default value for MachineDrainTimeout
-	DefaultMachineDrainTimeout = 12 * time.Hour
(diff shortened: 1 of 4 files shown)


amshuman-kr (approved):
LGTM

hardikdr (approved):
/lgtm

PR #558, opened 2020-11-11
Use listers to get the machine-object during reconciliation

### LINKED ISSUE #536: Use cache-based listers to GET the machine-object while reconciling.
**What would you like to be added**: MCM is GETing the machine-object via a direct client in each reconciliation of the object, MCM should rather rely on the listers/caches. With nearly 1000 machines per seed, this can amount to a large number of direct calls, which is generally not needed. 

Specific code-path which could be fixed quickly/easily: https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/machine.go#L179-L188

**Why is this needed**: Lesser load on seed apiserver, in turn, lesser ingress/egress traffic as well.

hardikdr:
**What this PR does / why we need it**:  MCM uses the cache-based listers instead of direct client calls while machine-reconciliation. 

Fixes #536 

**Release note**:

```improvement operator
Use cache-based listers to GET the machine-object while reconciling.
```

### DIFF
--- a/pkg/controller/machine.go
+++ b/pkg/controller/machine.go
@@ -176,7 +176,7 @@ func (c *controller) reconcileClusterMachine(machine *v1alpha1.Machine) error {
 		return nil
 	}
 
-	machine, err = c.controlMachineClient.Machines(machine.Namespace).Get(machine.Name, metav1.GetOptions{})
+	machine, err = c.machineLister.Machines(machine.Namespace).Get(machine.Name)
 	if err != nil {
 		klog.Errorf("Could not fetch machine object %s", err)
 		if apierrors.IsNotFound(err) {


prashanth26:
There are more machines Get calls on the APIServer like the one here - https://github.com/gardener/machine-controller-manager/blob/b56ee8413f84367b25972f3a3afb8739572d6b62/pkg/controller/machine.go#L836-L839. However, that would also mean to update this logic here - https://github.com/gardener/machine-controller-manager/blob/b56ee8413f84367b25972f3a3afb8739572d6b62/pkg/controller/machine.go#L853.

hardikdr:
Yes, I checked other occurrences, but then it seems, they are intentionally kept that way with the comment above:
```
```

prashanth26:
This is not an ideal controller practice. We are supposed to only act on the cached objects (by making a DeepCopy) as seen in the sample controller here - https://github.com/kubernetes/sample-controller/blob/255b8477561f5140b345ffcab586e524a4fd3a4c/controller.go#L321-L333. So, ideally we shouldn't be making GET call via APIServer. I guess this logic as added wrongly.

hardikdr:
Sure, agree. I'll update other occurrences as well. 

I got confused only because there is a clear message/comment justifying why direct client-call is made there in past. Btw do you remember why did we do it that way in the past, and also add the comment, wondering if the reasoning is still valid and no side-effects? Or it was done but simply not needed?

prashanth26:
>I got confused only because there is a clear message/comment justifying why direct client-call is made there in past. Btw do you remember why did we do it that way in the past, and also add the comment, wondering if the reasoning is still valid and no side-effects? Or it was done but simply not needed?

Vaguely yes. We were retrying when updates fail and this was helping that. However, I think the right way would be to return the error from this function and also the actual reconcile loop. However, we have to keep in mind any future references to this machine object to be tracked properly.

pkg/controller/machine.go
```diff
@@ -832,8 +831,7 @@ func (c *controller) updateMachineStatus(
 	lastOperation v1alpha1.LastOperation,
 	currentStatus v1alpha1.CurrentStatus,
 ) (*v1alpha1.Machine, error) {
-	// Get the latest version of the machine so that we can avoid conflicts
-	latestMachine, err := c.controlMachineClient.Machines(machine.Namespace).Get(machine.Name, metav1.GetOptions{})
+	latestMachine, err := c.machineLister.Machines(machine.Namespace).Get(machine.Name)
```
prashanth26, line 834: I don't think this change wouldn't be enough. Because then the machine update failure would go into an infinite loop and block the MCM. Refer -  https://github.com/gardener/machine-controller-manager/pull/558#issuecomment-725229214.

hardikdr: Update: we discussed offline, this cache should update later, and will not fall into an infinite loop. 
Although making many changes in the in-tree machine-controller seems unnecessary as it's already fixed well in OOT, we decided to make the change at one place in reconciliation which is the only major contributor to the API-calls.

prashanth26 (approved):
/lgtm for now. However, completely removing caches would require more changes here (or) moving to OOT MCM.

PR #559, opened 2020-11-16
Allow migration to continue when machineClass with same name as Provi…

prashanth26:
…derMachineClass is found

I observed a rare occurrence (only 1 cluster until now, tried to look through the code couldn't understand how this could occur) of a machine class migration failing for a shoot where the new machine class was created with the expected data, however, the machine objects were referring to an old non-existing ProviderMachineClass. This PR fixes such race conditions. 

Fixes #

I have added unit tests to check for these cases as well. 

**Release note**:

```improvement operator
Allow migration to continue when ProviderMachineClass is missing but MachineClass with the same name as ProviderMachineClass is found. Updates Machine object references to the MachineClass.
```

### DIFF
--- a/pkg/util/provider/machinecontroller/migrate_machineclass.go
+++ b/pkg/util/provider/machinecontroller/migrate_machineclass.go
@@ -78,6 +78,15 @@ func (c *controller) getProviderSpecificMachineClass(classSpec *v1alpha1.ClassSp
 	}
 
 	if err != nil {
+		if apierrors.IsNotFound(err) {
+			_, innerErr := c.machineClassLister.MachineClasses(c.namespace).Get(classSpec.Name)
+			if innerErr == nil {
+				// MachineClass object with same name was found.
+				// Most likely an external controller has already performed the migration.
+				// And also {Provider}MachineClass isn't found, hence we can continue with the migration.
+				return nil, fmt.Errorf("ProviderMachineClass not found. Found a MachineClass with matching name")
+			}
+		}
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/util/provider/machinecontroller/migrate_machineclass.go
```diff
@@ -382,26 +391,36 @@ func (c *controller) TryMachineClassMigration(classSpec *v1alpha1.ClassSpec) (*v
 	var (
 		err                          error
 		providerSpecificMachineClass interface{}
+		updateMachineClassObjects    = true
 	)
 
 	// Get the provider specific (e.g. AWSMachineClass) from the classSpec
 	if providerSpecificMachineClass, err = c.getProviderSpecificMachineClass(classSpec); err != nil {
-		return nil, nil, machineutils.ShortRetry, err
+		if err.Error() == "ProviderMachineClass not found. Found a MachineClass with matching name" {
```
hardikdr, line 399: Looks a bit strange to compare the error string this way, wondering if there could be an elegant way, . If not, a string-constant could be used to compare.

hardikdr (approved):
Looks good, just a minor comment, ignorable.

pkg/util/provider/machinecontroller/migrate_machineclass.go
prashanth26: This is the way I found online on searching as well - https://stackoverflow.com/questions/39121172/how-to-compare-go-errors. But yes, will update it to string constants.

PR #564, opened 2020-11-20
Set Machine Phase to Terminating & make drain calculations based on deletionTimestamp

### LINKED ISSUE #563: Machines are being force drained
**What happened**:
Machines are being force drained (before drain timeout) instead of a regular drain. This leads to disruption of PDBs and serial PV detachments causing downtimes.

Machines to not be force drained before hitting drain timeout. 

Take a cluster with several nodes running smoothly over the default drain timeout period. And now delete the machine backing this node. Force drain is triggered (not always) but in many cases.

- This is most likely a regression due to https://github.com/gardener/machine-controller-manager/pull/492, where machine phase change to terminating state was moved in the code. This caused the drain to consider an old timeStamp leading to force drains. 
- Refer [machineSet deletion](https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/machineset.go#L641-L662). 

**Environment**:
MCM in-tree v0.34.3. OOT is unaffected by this.

/priority blocker
/kind regression

prashanth26:
This PR fixes the race condition that triggers force deletion of machines which are less than drain timeout minutes old. Fixes regressions from #492 . 

Fixes #563 

**Release note**:

```improvement operator
Set Machine Phase to Terminating before draining. 
```
```noteworthy operator
Machine force deletion computation is based on deletionTimestamp instead of LastUpdatedTimestamp. 
```
/invite @guydaichs

### DIFF
--- a/pkg/util/provider/machinecontroller/machine_util.go
+++ b/pkg/util/provider/machinecontroller/machine_util.go
@@ -40,6 +40,7 @@ import (
 	"github.com/gardener/machine-controller-manager/pkg/util/provider/machinecodes/status"
 	"github.com/gardener/machine-controller-manager/pkg/util/provider/machineutils"
 	utilstrings "github.com/gardener/machine-controller-manager/pkg/util/strings"
+	utiltime "github.com/gardener/machine-controller-manager/pkg/util/time"
 
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
@@ -912,9 +913,7 @@ func (c *controller) drainNode(deleteMachineRequest *driver.DeleteMachineRequest
 	if skipDrain {
 		state = v1alpha1.MachineStateProcessing
 	} else {
-		// Timeout value obtained by subtracting last operation with expected time out period
-		timeOut := metav1.Now().Add(-timeOutDuration).Sub(machine.Status.CurrentStatus.LastUpdateTime.Time)
-		timeOutOccurred = timeOut > 0
+		timeOutOccurred = utiltime.HasTimeOutOccurred(*machine.DeletionTimestamp, timeOutDuration)
 
 		if forceDeleteLabelPresent || timeOutOccurred {
 			// To perform forceful machine drain/delete either one of the below conditions must be satified
--- a/pkg/util/time/time_test.go
+++ b/pkg/util/time/time_test.go
@@ -63,7 +63,7 @@ var _ = Describe("time", func() {
 					timeOutOccurred: false,
 				},
 			}),
-			Entry("Time stamp is now and peroid is 5mins", &data{
+			Entry("Time stamp is now and period is 5mins", &data{
 				setup: setup{
 					timeStamp: metav1.Time{Time: time.Now()},
 					period:    5 * time.Minute,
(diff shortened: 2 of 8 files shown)


prashanth26:
Open for initial review. 
Please /hold until tests are done.

guydaichs (approved):
lgtm. thanks for catching this.

amshuman-kr (commented):
Why don't we use `deletionTimestamp` here? Using `CurrentStatus.LastUpdateTime` seems flaky.
```go
		timeOut := metav1.Now().Add(-timeOutDuration).Sub(machine.Status.CurrentStatus.LastUpdateTime.Time)
```

prashanth26:
Hi @amshuman-kr ,

I got the same thought lately when I was looking into the issue. But to maintain the original logic, I stuck to this. However, I also think deletionTimeStamp should be more consistent provided the drain logic isn't delayed for some other reason.

prashanth26:
Hi @amshuman-kr ,

I have made this change here - https://github.com/gardener/machine-controller-manager/pull/564/files#diff-d52f3782153512c736878e861c6cafd75dd23f214af3f017eea14de256da458bR658. However, one of the test cacse seems to be failing will fix that and test this change with a running ccluster. In the meanwhile have pushed the changes for you to have look.

prashanth26:
/unhold 
open for review

prashanth26:
/ok-to-test
Tests failed due to rate limiting at docker

prashanth26:
Just ran the MCM locally with and without the fix. Without the fix I am able to reproduce the force drain issue, but with the fix it no longer occurs.

amshuman-kr (approved):
LGTM

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -912,9 +913,7 @@ func (c *controller) drainNode(deleteMachineRequest *driver.DeleteMachineRequest
 	if skipDrain {
 		state = v1alpha1.MachineStateProcessing
 	} else {
-		// Timeout value obtained by subtracting last operation with expected time out period
-		timeOut := metav1.Now().Add(-timeOutDuration).Sub(machine.Status.CurrentStatus.LastUpdateTime.Time)
-		timeOutOccurred = timeOut > 0
+		timeOutOccurred = utiltime.HasTimeOutOccurred(*machine.DeletionTimestamp, timeOutDuration)
```
hardikdr, line 916: Nice, thanks for this.

pkg/util/time/time_test.go
```diff
@@ -0,0 +1,77 @@
...
+			Entry("Time stamp is one hour ago and period is 90mins", &data{
+				setup: setup{
+					timeStamp: metav1.Time{Time: time.Now().Add(-time.Hour)},
+					period:    90 * time.Minute,
+				},
+				expect: expect{
+					timeOutOccurred: false,
+				},
+			}),
+			Entry("Time stamp is now and peroid is 5mins", &data{
```
hardikdr, line 66:
```suggestion
			Entry("Time stamp is now and period is 5mins", &data{
```

hardikdr (approved):
/lgtm 
Thanks, just a minor typo, ignorable.

PR #566, opened 2020-11-23
Fix Azure machine deletion if resource group is gone

### LINKED ISSUE #562: machine-controller-manager cannot delete Machine when the resource group is already deleted
**What happened**:

```
$ k -n shoot--foo--bar get machines shoot--foo--bar-bar-z3-5bf676f44d-kt27p -o yaml
```

```yaml
status:
  currentStatus:
    lastUpdateTime: "2020-11-20T07:52:09Z"
    phase: Terminating
  lastOperation:
    description: 'Cloud provider message - compute.VirtualMachinesClient#List: Failure
      responding to request: StatusCode=404 -- Original Error: autorest/azure: Service
      returned an error. Status=404 Code="ResourceGroupNotFound" Message="Resource
      group ''shoot--foo--bar'' could not be found."'
    lastUpdateTime: "2020-11-20T07:52:09Z"
    state: Failed
    type: Delete
```

machine-controller-manager should be able to delete Machine when the resource group is already deleted. For example the azure cloud-controller-manager can handle this case and can delete Service of type LoadBalancer when the underlying resource-group is already deleted.

1. Create an Azure Machine
2. Delete the underlying resource group
3. Ensure that machine-controller-manager cannot delete the Machine CR

**Environment**:
- machine-controller-manager version: v0.34.3

dkistner:
Allow MCM to delete Azure machines even if the underlying resource groups are already deleted. E. g. user delete the resource group manually.

Fixes #562

**Release note**:

```improvement operator
MCM will delete Azure machines even if the underlying resource group is already deleted.
```

/invite @kon-angelo @ialidzhikov

### DIFF
--- a/pkg/driver/driver_azure.go
+++ b/pkg/driver/driver_azure.go
@@ -284,6 +284,14 @@ func (d *AzureDriver) Delete(machineID string) error {
 		resourceGroupName = d.AzureMachineClass.Spec.ResourceGroup
 	)
 
+	// Check if the underlying resource group still exists, if not skip the deletion as all resources are gone.
+	if _, err := clients.group.Get(ctx, resourceGroupName); err != nil {
+		if notFound(err) {
+			return nil
+		}
+		return err
+	}
+
 	var dataDiskNames []string
 	if d.AzureMachineClass.Spec.Properties.StorageProfile.DataDisks != nil && len(d.AzureMachineClass.Spec.Properties.StorageProfile.DataDisks) > 0 {
 		dataDiskNames = getAzureDataDiskNames(d.AzureMachineClass.Spec.Properties.StorageProfile.DataDisks, vmName, dataDiskSuffix)
@@ -386,6 +394,7 @@ type azureDriverClients struct {
 	vm          compute.VirtualMachinesClient
 	disk        compute.DisksClient
 	deployments resources.DeploymentsClient
+	group       resources.GroupsClient
 	images      compute.VirtualMachineImagesClient
 	marketplace marketplaceordering.MarketplaceAgreementsClient
 }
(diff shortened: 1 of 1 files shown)


kon-angelo (approved):
/lgtm

prashanth26:
cc: @AxiomSamarth 
TODO: incorporate in OOT branch.

prashanth26 (approved):
Thanks for this fix @dkistner .
/lgtm

ialidzhikov (changes requested):
Can you add an unit test about it to make sure that we don't regress it in the future?

dkistner:
> TODO: incorporate in OOT branch.

https://github.com/gardener/machine-controller-manager-provider-azure/issues/6 (currently private)

> Can you add an unit test about it to make sure that we don't regress it in the future?

Hmm. We don't have much unit tests yet for the Azure driver (primarily for the data disks feature) and implementing this would mean to mock a lot of Azure api calls. Don't get me wrong I wanted to have that tests, but I guess it would make more sense (from an effort point of view) to do it in the out-of-tree, where we anyways then need to increase the test coverage.

hardikdr (approved):

prashanth26:
>Hmm. We don't have much unit tests yet for the Azure driver (primarily for the data disks feature) and implementing this would mean to mock a lot of Azure api calls. Don't get me wrong I wanted to have that tests, but I guess it would make more sense (from an effort point of view) to do it in the out-of-tree, where we anyways then need to increase the test coverage.

I would second this. With the OOT @AxiomSamarth is already building a framework for mocking and writing tests for azure MCM OOT provider. Should be easier to get it done there.

prashanth26:
Can we merge this change for now? @ialidzhikov

ialidzhikov:
> Can we merge this change for now? @ialidzhikov

Sure, please go on. It was just a remark from outside that we miss unit tests for such small things ;).

prashanth26:
Currently, the core MCM logic has a good framework for unit tests. However, with driver's, we haven't invested here as we are moving all this logic into OOT.

ialidzhikov:
Sure, thanks. I got it.

/lgtm

kon-angelo:
Partially the problem lies with client libraries and what they expose to make unit tests. On AWS you get an interface that you can mock, but for others (azure, openstack) you mostly get structs and functions exposed and that means that there is quite a bit of manual labor involved to have proper unit tests. Not saying we shouldn't do it though.

PR #578, opened 2020-11-29
Support optional secret containing only credentials but no user-data; accept Gardener secret data keys

rfranzke:
/area open-source usability
/kind enhancement
/priority normal

This PR extends all in-tree and the out-of-tree machine classes with a new `.{spec.}credentialsSecretRef` field. It is an optional reference to a secret containing only the credentials, while today's `.{spec.}secretRef` still contains the user-data.
This is helpful for scenarios like Gardeners where the credentials for all machine classes are the same while only the user-data differs.

Additionally, all the secret keys used by Gardener are now also allowed as alternatives for the machine-controller-manager. This helps to not make mappings for the data keys.

The implementation is entirely backwards-compatible and optional, both for the in-tree and the out-of-tree implementations. The in-tree drivers have been simplified by only working with the secret data map instead of the whole `v1.Secret` object. The out-of-tree contract is untouched. If a credentials secret reference is provided then its content will be merged into the secret that is passed to the provider.

ℹ️ For simplification of the review process the PR is split into 4 different commits that could be reviewed individually.

/assign @hardikdr @prashanth26 @AxiomSamarth 
/invite @hardikdr @prashanth26 @AxiomSamarth 

/cc @vpnachev @timuthy 

**Release note**:

```feature operator
All machine classes do now support an optional `.{spec.}credentialsSecretRef` field in addition to today's `.{spec.}secretRef` field. If `.{spec.}credentialsSecretRef` is non-nil then the provider credentials will be read out of this secret. The user-data for the machine bring-up is still required to be part of the secret referenced by `.{spec.}secretRef`.
```
```feature operator
Some machine class secrets are now supporting alternative data keys:
* The machine class secret for Alicloud machines does now also accept the data keys `accessKeyID` and `accessKeySecret` as alternatives for today's keys.
* The machine class secret for AWS machines does now also accept the data keys `accessKeyID` and `secretAccessKey` as alternatives for today's keys.
* The machine class secret for Azure machines does now also accept the data keys `clientID`, `clientSecret`, `subscriptionID` and `tenantID` as alternatives for today's keys.
* The machine class secret for GCP machines does now also accept the data key `serviceaccount.json` as alternatives for today's key.
```

### DIFF
--- a/pkg/apis/machine/v1alpha1/aws_machineclass_types.go
+++ b/pkg/apis/machine/v1alpha1/aws_machineclass_types.go
@@ -67,18 +67,19 @@ type AWSMachineClassList struct {
 
 // AWSMachineClassSpec is the specification of a AWSMachineClass.
 type AWSMachineClassSpec struct {
-	AMI               string                      `json:"ami,omitempty"`
-	Region            string                      `json:"region,omitempty"`
-	BlockDevices      []AWSBlockDeviceMappingSpec `json:"blockDevices,omitempty"`
-	EbsOptimized      bool                        `json:"ebsOptimized,omitempty"`
-	IAM               AWSIAMProfileSpec           `json:"iam,omitempty"`
-	MachineType       string                      `json:"machineType,omitempty"`
-	KeyName           string                      `json:"keyName,omitempty"`
-	Monitoring        bool                        `json:"monitoring,omitempty"`
-	NetworkInterfaces []AWSNetworkInterfaceSpec   `json:"networkInterfaces,omitempty"`
-	Tags              map[string]string           `json:"tags,omitempty"`
-	SpotPrice         *string                     `json:"spotPrice,omitempty"`
-	SecretRef         *corev1.SecretReference     `json:"secretRef,omitempty"`
+	AMI                  string                      `json:"ami,omitempty"`
+	Region               string                      `json:"region,omitempty"`
+	BlockDevices         []AWSBlockDeviceMappingSpec `json:"blockDevices,omitempty"`
+	EbsOptimized         bool                        `json:"ebsOptimized,omitempty"`
+	IAM                  AWSIAMProfileSpec           `json:"iam,omitempty"`
+	MachineType          string                      `json:"machineType,omitempty"`
+	KeyName              string                      `json:"keyName,omitempty"`
+	Monitoring           bool                        `json:"monitoring,omitempty"`
+	NetworkInterfaces    []AWSNetworkInterfaceSpec   `json:"networkInterfaces,omitempty"`
+	Tags                 map[string]string           `json:"tags,omitempty"`
+	SpotPrice            *string                     `json:"spotPrice,omitempty"`
+	SecretRef            *corev1.SecretReference     `json:"secretRef,omitempty"`
+	CredentialsSecretRef *corev1.SecretReference     `json:"credentialsSecretRef,omitempty"`
 
 	// TODO add more here
 }
--- a/pkg/apis/machine/v1alpha1/aws_machineclass_types.go
+++ b/pkg/apis/machine/v1alpha1/aws_machineclass_types.go
@@ -28,6 +28,13 @@ const (
 	AWSAccessKeyID string = "providerAccessKeyId"
 	// AWSSecretAccessKey is a constant for a key name that is part of the AWS cloud credentials.
 	AWSSecretAccessKey string = "providerSecretAccessKey"
+
+	// AWSAlternativeAccessKeyID is a constant for a key name of a secret containing the AWS credentials (access key
+	// id).
+	AWSAlternativeAccessKeyID = "accessKeyID"
+	// AWSAlternativeAccessKeySecret is a constant for a key name of a secret containing the AWS credentials (access key
+	// secret).
+	AWSAlternativeSecretAccessKey = "secretAccessKey"
 )
 
 // +genclient
--- a/cmd/machine-controller-manager-cli/main.go
+++ b/cmd/machine-controller-manager-cli/main.go
@@ -99,7 +99,7 @@ func main() {
 		log.Fatalf("Could not parse machine class yaml: %s", err)
 	}
 
-	driver := driver.NewDriver(machineID, &secret, classKind, machineclass, machineName)
+	driver := driver.NewDriver(machineID, secret.Data, classKind, machineclass, machineName)
 
 	if machineID == "" {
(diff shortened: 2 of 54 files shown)


pkg/apis/machine/v1alpha1/aws_machineclass_types.go
```diff
@@ -28,6 +28,13 @@ const (
 	AWSAccessKeyID string = "providerAccessKeyId"
 	// AWSSecretAccessKey is a constant for a key name that is part of the AWS cloud credentials.
 	AWSSecretAccessKey string = "providerSecretAccessKey"
+
+	// AWSAlternativeAccessKeyID is a constant for a key name of a secret containing the AWS credentials (access key
+	// id).
+	AWSAlternativeAccessKeyID = "accessKeyID"
+	// AWSAlternativeAccessKeySecret is a constant for a key name of a secret containing the AWS credentials (access key
+	// secret).
+	AWSAlternativeSecretAccessKey = "secretAccessKey"
```
prashanth26, line 37: What if we kept the field names consistent for both the secrets. And give preference to the secrets used by `secretRef` while merging them as it is with this change. Refer - https://github.com/gardener/machine-controller-manager-provider-aws/pull/16/files#diff-107d76bae843a73a08db977af838302f8fdabea9dc6236ac560afe71b2ed1e00R47? Similar to how we are doing with this logic, however, will avoid these alternate field names and also refactoring at various MCM OOT providers.

However, one disadvantage I also see in this approach is that the (OOT) provider doesn't have a choice to switch between the two secrets, which I think should be fine.

prashanth26 (commented):
/invite @amshuman-kr

pkg/apis/machine/v1alpha1/aws_machineclass_types.go
rfranzke: In both secrets (`.secretRef` and `.credentialsSecretRef` both variants are possible, the old names and the alternative keys). If `.credentialsSecretRef` is present then the credentials will be taken from here, independent of whether they already are in `.secretRef`.
This change here is related to the last commit which is independent of the support for `.credentialsSecretRef`. I wanted to open a separate PR, however, I have built this on top of the other changes and would have ended up with conflicts, hence, I decided to put it in here as well. If you want I can move this out into a separate PR, though.

prashanth26: Okay, now I got it. 

However, I still wonder if we need these `alternate fields`? Because this would confuse the end-user more with 4 different combinations for fields to provide? 

I would prefer only giving `AWSAccessKeyID ` on both secretRefs (giving preference to `.credentialsSecretRef`). And not providing the new `AWSAlternativeAccessKeyID ` in order to reduce the different combinations and in-turn confusion? 

However, just my opinion. Open to discussions.

rfranzke: The motivation behind it is the following: When MCM supports `.credentialsRef` then the provider extensions of Gardener will make us of it by NOT creating a new secret containing the credentials but by leveraging/re-using the already existing `cloudprovider` secret in the shoot namespace. Hence, no extra handling or any implementation will be necessary other than referencing this secret. For this to work, MCM needs to understand the data keys used inside Gardener.
The alternative data keys are completely optional, nobody has to use them, so not sure where the mentioned confusion shall arise from. The DNS extension supports them as well as alternative, see e.g. https://github.com/gardener/external-dns-management/blob/master/examples/20-secret-aws-credentials.yaml#L7-L14.
If you have ideas how to improve the example secrets or whether to put some documentation I'm open to do that, of course :)

prashanth26: Okay sure. If you feel so, you probably have a better idea :). I am okay with it. 

>If you have ideas how to improve the example secrets or whether to put some documentation I'm open to do that, of course :)
I think the current documentation changes are enough.

prashanth26: And yes, i re-read your comments and it makes more sense to me now. Thanks 👍

rfranzke: Cool :)

amshuman-kr (approved):
Thanks for the quick and extensive PR @rfranzke ! LGTM

The only question I have is that while this PR retains the signature of the OOT interfaces but changes the semantics. I.e. the secret object that is passed is no longer backed by a _real_ secret object. As long as the OOT providers do not expect the secret object to be a real one in the apiserver, this should be fine. Ideally, they shouldn't. @prashanth26 Should we make this explicit in the documentation?

vpnachev (approved):
/lgtm

rfranzke:
> The only question I have is that while this PR retains the signature of the OOT interfaces but changes the semantics. I.e. the secret object that is passed is no longer backed by a real secret object. As long as the OOT providers do not expect the secret object to be a real one in the apiserver, this should be fine. Ideally, they shouldn't. @prashanth26 Should we make this explicit in the documentation?

Agreed, good point. I stumbled across this as well, and for the sake of not changing the contract I created a new `v1.Secret` object and merged all data into it. IMO the interface should be changed to only work with a secret data map and not with the whole secret object, similar to what I've done for the in-tree drivers. I guess all out-of-tree providers are only using the data of the secret anyways.

amshuman-kr:
> IMO the interface should be changed to only work with a secret data map and not with the whole secret object, similar to what I've done for the in-tree drivers. I guess all out-of-tree providers are only using the data of the secret anyways.

I agree. That way the contract will be explicit. @prashanth26 Should we consider changing the OOT signature? We need not do it in the PR, of course.

prashanth26:
I agree with you guys. Something I probably overlooked while making the initial interface. Will try to incorporate this change in the coming releases. Yes, for now, let's stick to this.

prashanth26 (approved):
/lgtm
Waiting for approval from @amshuman-kr

amshuman-kr (approved):
LGTM

> Something I probably overlooked while making the initial interfaces. 

We all did @prashanth26. Hindsight is 20-20 🙂  It is 2020 too 😉

hardikdr (approved):
/lgtm

prashanth26:
I shall trigger an MCM release as the new field will have to be added in all the provider migration codes - https://github.com/gardener/machine-controller-manager/blob/master/pkg/apis/machine/types.go#L1213,

PR #551, opened 2020-11-04
Fix the OpenAPI validation of the machine-deployment crd

hardikdr:
**What this PR does / why we need it**:  Fix the OpenAPI validation of the machine-deployment crd
Related to: https://github.com/kubernetes/kubernetes/issues/80493

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```improvement operator
Fix the OpenAPI validation of the machine-deployment crd
```

PR #593, opened 2021-03-03
Machines are reconciled only when MachineClasses have valid finalizers

### LINKED ISSUE #560: Add finalizers on machineClasses only when machines referring it
**What happened**:
Currently, MCM adds finalizers on machineClasses unless the deletionTimestamp is set. And this adds finalizers to machineClass that probably aren't used. 

MachineClasses to have finalizers only when machines backing them are present.

This would also mean that the machine reconciliation for a machine object should not be started unless there is a machineClass finalizers set by MCM. Because in the past we had made this change however, the race conditions lead to machines without machineClass objects backing it. 

**Environment**:
/area ops-productivity
/size xs
/exp beginner

AxiomSamarth:
Implements MachineClasses to have finalizers only when machines backing them are present.

Fixes #560 

The change has been introduced for OOT MCM common library. If there is a need then it shall be introduced in In-Tree too.

**Release note**:

```feature operator
Finalizers will be added to the MachineClass which is used by at least one machine. Machines whose backing MachineClass does not have finalizers shall not be reconciled.
```

### DIFF
--- a/kubernetes/crds/machine.sapcloud.io_machinedeployments.yaml
+++ b/kubernetes/crds/machine.sapcloud.io_machinedeployments.yaml
@@ -59,9 +59,6 @@ spec:
             object represents. Servers may infer this from the endpoint the client
             submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
           type: string
-        metadata:
-          description: Standard object metadata.
-          type: object
         spec:
           description: Specification of the desired behavior of the MachineDeployment.
           properties:
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -134,6 +134,7 @@ func (c *controller) reconcileClusterMachine(machine *v1alpha1.Machine) (machine
 		return machineutils.LongRetry, err
 	}
 
+	// Validate MachineClass
 	machineClass, secretData, retry, err := c.ValidateMachineClass(&machine.Spec.Class)
 	if err != nil {
 		klog.Error(err)
--- a/pkg/util/provider/machinecontroller/machine_test.go
+++ b/pkg/util/provider/machinecontroller/machine_test.go
@@ -224,8 +224,8 @@ var _ = Describe("machine", func() {
 
 	Describe("#ValidateMachineClass", func() {
 		type setup struct {
-			aws     []*v1alpha1.MachineClass
-			secrets []*corev1.Secret
+			machineClass []*v1alpha1.MachineClass
+			secrets      []*corev1.Secret
 		}
 		type expect struct {
 			machineClass interface{}
@@ -241,6 +241,12 @@ var _ = Describe("machine", func() {
 		objMeta := &metav1.ObjectMeta{
 			GenerateName: "class",
 			Namespace:    testNamespace,
+			Finalizers:   []string{MCMFinalizerName},
+		}
+
+		objMetaWithoutFinalizer := &metav1.ObjectMeta{
+			GenerateName: "class",
+			Namespace:    testNamespace,
 		}
 
 		DescribeTable("##table",
@@ -248,9 +254,9 @@ var _ = Describe("machine", func() {
 				stop := make(chan struct{})
 				defer close(stop)
 
-				machineObjects := []runtime.Object{}
-				for _, o := range data.setup.aws {
-					machineObjects = append(machineObjects, o)
+				controlObjects := []runtime.Object{}
+				for _, machineClass := range data.setup.machineClass {
+					controlObjects = append(controlObjects, machineClass)
 				}
 
 				coreObjects := []runtime.Object{}
@@ -258,7 +264,7 @@ var _ = Describe("machine", func() {
 					coreObjects = append(coreObjects, o)
 				}
 
-				controller, trackers := createController(stop, objMeta.Namespace, machineObjects, coreObjects, nil, nil)
+				controller, trackers := createController(stop, objMeta.Namespace, controlObjects, coreObjects, nil, nil)
 				defer trackers.Stop()
 
 				waitForCacheSync(stop, controller)
@@ -282,7 +288,7 @@ var _ = Describe("machine", func() {
 			},
 			Entry("non-existing machine class", &data{
 				setup: setup{
-					aws: []*v1alpha1.MachineClass{
+					machineClass: []*v1alpha1.MachineClass{
 						{
 							ObjectMeta: *newObjectMeta(objMeta, 0),
 						},
@@ -299,7 +305,7 @@ var _ = Describe("machine", func() {
 			Entry("non-existing secret", &data{
 				setup: setup{
 					secrets: []*corev1.Secret{},
-					aws: []*v1alpha1.MachineClass{
+					machineClass: []*v1alpha1.MachineClass{
 						{
 							ObjectMeta: *newObjectMeta(objMeta, 0),
 							SecretRef:  newSecretReference(objMeta, 0),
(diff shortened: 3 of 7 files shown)


prashanth26 (changes requested):
Thanks for this PR. 

The unit tests probably need refactoring as well.

AxiomSamarth:
Unit tests refactored and a new relevant test added too! 😄

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -134,6 +134,8 @@ func (c *controller) reconcileClusterMachine(machine *v1alpha1.Machine) (machine
 		return machineutils.LongRetry, err
 	}
 
+	// Validate MachineClass
+	klog.Infof("Validating Machine Class %s backing machine %s", machine.Spec.Class.Name, machine.Name)
```
prashanth26, line 138: Do we need this?

kubernetes/crds/machine.sapcloud.io_machinedeployments.yaml
```diff
@@ -59,9 +59,6 @@ spec:
             object represents. Servers may infer this from the endpoint the client
             submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
           type: string
-        metadata:
```
prashanth26, line 62: I hope this change is intensional?

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -123,6 +123,18 @@ func (c *controller) ValidateMachineClass(classSpec *v1alpha1.ClassSpec) (*v1alp
...
+	clone := machineClass.DeepCopy()
+
+	if finalizers := sets.NewString(clone.Finalizers...); !finalizers.Has(MCMFinalizerName) {
+		klog.Errorf("The machine class %s has no finalizers set. So not reconciling the machine.", machineClass.Name)
+		err = c.reconcileClusterMachineClassKey(machineClass.Name)
+		if err != nil {
+			klog.Errorf("Machine Class %q: Error occured while reconciling: %v", machineClass.Name, err)
+			return machineClass, secretData, retry, err
+		}
+		return machineClass, secretData, retry, err
```
prashanth26, line 135: The controller way of doing this is by [re-enqueing the object to the queue](https://github.com/gardener/machine-controller-manager/blob/master/pkg/util/provider/machinecontroller/machineclass.go#L87) and returning an error here.

prashanth26 (changes requested):
Apologies for the super late review. Better late than never 😄

kubernetes/crds/machine.sapcloud.io_machinedeployments.yaml
AxiomSamarth: Yes. Prashanth. It is the solution for the issue - https://github.com/gardener/machine-controller-manager/pull/551

pkg/util/provider/machinecontroller/machine.go
AxiomSamarth: Yeah, it does not really make a difference. I will remove it

pkg/util/provider/machinecontroller/machine_util.go
AxiomSamarth: Okay

kubernetes/crds/machine.sapcloud.io_machinedeployments.yaml
prashanth26: Okay. In future, let's try to move minor independent fixes into it's on PRs. Just to keep it followable via commits.

prashanth26:
The commits seem to be messed up. Can you please fix it?

prashanth26 (approved):
/lgtm

pkg/util/provider/machinecontroller/machine_test.go
```diff
@@ -241,6 +241,13 @@ var _ = Describe("machine", func() {
 		objMeta := &metav1.ObjectMeta{
 			GenerateName: "class",
 			Namespace:    testNamespace,
+			Finalizers:   []string{MCMFinalizerName},
+		}
+
+		objMetaWithoutFinalizer := &metav1.ObjectMeta{
+			GenerateName: "class",
+			Namespace:    testNamespace,
+			Finalizers:   []string{MCMFinalizerName},
```
himanshu-kun, line 250: I think `Finalizers` field shouldn't be a part of this

pkg/util/provider/machinecontroller/machine_test.go
```diff
@@ -346,6 +362,34 @@ var _ = Describe("machine", func() {
 					err:        false,
 				},
 			}),
+			Entry("machineClass without Finalizer", &data{
```
himanshu-kun, line 365: are we just trying to show that machineClass which doesn't have finalizer will be marked by finalizer in this Testcase.

pkg/util/provider/machinecontroller/machine_test.go
```diff
@@ -253,6 +260,15 @@ var _ = Describe("machine", func() {
...
 
+				machine1 := &v1alpha1.Machine{
+					ObjectMeta: *newObjectMeta(objMeta, 0),
+					Spec: v1alpha1.MachineSpec{
+						Class: *data.action,
+					},
+				}
+
+				machineObjects = append(machineObjects, machine1)
+
```
himanshu-kun, line 271: I see this `machine1` object being appended in `machineObjects` list but I don't see it being used in any testcases?

himanshu-kun (changes requested):
Thanks for giving me this opportunity to review. I think I have my doubts at some places, would be really helpful if you could clear them @AxiomSamarth . I won't be able to respond today as I am on leave.
Pls proceed with merging the PR if you think you have answered my doubts :)

pkg/util/provider/machinecontroller/machine_test.go
AxiomSamarth: Good catch! Thank you

pkg/util/provider/machinecontroller/machine_test.go
AxiomSamarth: The idea is - if a MachineClass does not have a finaliser then the machineClass should not be reconciled.

pkg/util/provider/machinecontroller/machine_test.go
AxiomSamarth: You are correct, Himanshu. It is a redundant entry. 

The idea behind creating the machine object was to check that the machineClass gets a finaliser only when there is atleast one machine using it. I have written a different test case in `MachineClass_test.go` now. Again, good catch 😉

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -123,6 +124,15 @@ func (c *controller) ValidateMachineClass(classSpec *v1alpha1.ClassSpec) (*v1alp
...
 
+	clone := machineClass.DeepCopy()
+
+	if finalizers := sets.NewString(clone.Finalizers...); !finalizers.Has(MCMFinalizerName) {
+		klog.Errorf("The machine class %s has no finalizers set. So not reconciling the machine.", machineClass.Name)
+		c.machineClassQueue.Add(machineClass.Name)
+		err := errors.New("The machine class %s has no finalizers set. So not reconciling the machine." + machineClass.Name)
+		return nil, nil, retry, err
+	}
+
```
prashanth26, line 135:
```suggestion
	if finalizers := sets.NewString(machineClass.Finalizers...); !finalizers.Has(MCMFinalizerName) {
		klog.Errorf("The machine class %s has no finalizers set. So not reconciling the machine.", machineClass.Name)
		err := errors.New("The machine class %s has no finalizers set. So not reconciling the machine." + machineClass.Name)
		c.machineClassQueue.Add(machineClass.Name)
		return nil, nil, retry, err
	}

```
Do we need to clone this machineClass object?

prashanth26 (commented):
One final minor comment

pkg/util/provider/machinecontroller/machine_util.go
AxiomSamarth: Changes made in the latest commit.

prashanth26:
/title Machines are reconciled only when MachineClasses have valid finalizers

Issue #544, opened 2020-10-23
Orphaned Azure NICs block subnet deletion afterwards


ialidzhikov:
**What happened**:
I see quite often failed Azure Shoot deletions with error similar to:

```
-> Pod 'bar.infra.tf-destroy-jv2w9' reported:
* Error deleting Subnet "shoot--foo--bar-nodes" (Virtual Network "shoot--foo--bar" / Resource Group "shoot--foo--bar"): network.SubnetsClient#Delete: Failure sending request: StatusCode=400 -- Original Error: Code="InUseSubnetCannotBeDeleted" Message="Subnet shoot--foo--bar-nodes is in use by /subscriptions/<omitted>/resourceGroups/shoot--foo--bar/providers/Microsoft.Network/networkInterfaces/shoot--foo--bar-cpu-worker-0-z1-7d4cd95c59-p5qgg-nic/ipConfigurations/shoot--foo--bar-cpu-worker-0-z1-7d4cd95c59-p5qgg-nic and cannot be deleted. In order to delete the subnet, delete all the resources within the subnet. See aka.ms/deletesubnet." Details=[]

Releasing state lock. This may take a few moments...
```

Basically the Worker (and machine-controller-manager) resources are deleted but for some reason the NICs are not deleted and block the subnet deletion afterwards (during the Infrastructure deletion).

I can see that the NIC still exists:

```
$ az network nic show -g shoot--foo--bar -n shoot--foo--bar-cpu-worker-0-z1-7d4cd95c59-p5qgg-nic
```

**What you expected to happen**:
No NICs to be orphaned.

**How to reproduce it (as minimally and precisely as possible)**:
Not clear for now.

**Environment**:
- machine-controller-manager version: v0.34.3

prashanth26:
HI @ialidzhikov ,

How often have you observed this? And do you remember when the first occurrence of something like this was? Because when I look at the history changes on the Azure driver, I don't see any significant change in the last few months (atleast for v0.34.3, as there is a minor change in the master) - https://github.com/gardener/machine-controller-manager/commits/rel-v0.34.0/pkg/driver/driver_azure.go. 

Also just checked the VM deletion logic looks like the machine deletion occurs along with Machine, Disk and NIC deletion. Also orphan NIC and disks are checked even after deletion here - https://github.com/gardener/machine-controller-manager/blob/master/pkg/driver/driver_azure.go#L308-L323. I am not sure how the situation occurs :o

[...]

PR #594, opened 2021-03-15
Improved NIC creation and deletion logic

prashanth26:
- NIC creation now adopts any existing NICs with matching names
- NIC deletion confirms deletion by performing GET on the deletion
- Improved creation logs to give more details

Fixes https://github.com/gardener/machine-controller-manager/issues/544

**Release note**:

```other operator
Improved NIC creation and deletion logic to handle NIC creation and deletions more gracefully.
```

### DIFF
--- a/pkg/driver/driver_azure.go
+++ b/pkg/driver/driver_azure.go
@@ -29,6 +29,7 @@ import (
 	"github.com/gardener/machine-controller-manager/pkg/apis/machine"
 	v1alpha1 "github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
 	"github.com/gardener/machine-controller-manager/pkg/metrics"
+	backoff "github.com/gardener/machine-controller-manager/pkg/util/backoff"
 	"github.com/prometheus/client_golang/prometheus"
 	corev1 "k8s.io/api/core/v1"
 
@@ -489,6 +490,7 @@ func (d *AzureDriver) createVMNicDisk() (*compute.VirtualMachine, error) {
 	/*
 		Subnet fetching
 	*/
+	klog.V(3).Infof("Fetching subnet details for VM %q", vmName)
 	// Getting the subnet object for subnetName
 	subnet, err := clients.subnet.Get(
 		ctx,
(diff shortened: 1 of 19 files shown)


pkg/driver/driver_azure.go
```diff
@@ -502,48 +503,60 @@ func (d *AzureDriver) createVMNicDisk() (*compute.VirtualMachine, error) {
...
-	/*
-		NIC creation
-	*/
+	NIC, err := clients.nic.Get(ctx, resourceGroupName, nicName, "")
+	//klog.Infof("Error: %v \nNIC: %v", err, NIC)
 
-	// Creating NICParameters for new NIC creation request
-	NICParameters := d.getNICParameters(vmName, &subnet)
+	// Change to something like below, however I am unable to find the helper method for this.
+	// if err != nil && management.IsResourceNotFoundError(err) {
```
prashanth26, line 510: I am having some trouble with validating the error code when NIC deosn't exist. Any suggestions? 

cc: @AxiomSamarth , @amshuman-kr

pkg/driver/driver_azure.go
```diff
@@ -1035,20 +1047,64 @@ func (clients *azureDriverClients) checkOrphanDisks(ctx context.Context, resourc
 }
 
 func (clients *azureDriverClients) deleteNIC(ctx context.Context, resourceGroupName string, nicName string) error {
+	var (
+		nicDeletionTimeout     = 10 * time.Minute
```
amshuman-kr, line 1051: Constant?

pkg/driver/driver_azure.go
```diff
@@ -1035,20 +1047,64 @@ func (clients *azureDriverClients) checkOrphanDisks(ctx context.Context, resourc
...
+
 	onARMAPISuccess(prometheusServiceNIC, "NIC deletion was successful for %s", nicName)
 	return nil
 }
 
+func (clients *azureDriverClients) waitUntilNICIsDeleted(ctx context.Context, resourceGroupName string, nicName string) error {
+	var (
+		sleepInterval    = 10 * time.Second
+		maxSleepInterval = 3 * time.Minute
+		currentSleepTime = 0 * time.Second
```
amshuman-kr, line 1079: Constants? Or even better, why not use client-go/backoff? https://github.com/kubernetes/client-go/blob/master/util/flowcontrol/backoff.go#L32-L38

pkg/driver/driver_azure.go
```diff
@@ -1035,20 +1047,64 @@ func (clients *azureDriverClients) checkOrphanDisks(ctx context.Context, resourc
...
+
+	if err := clients.waitUntilNICIsDeleted(nicDeletionCtx, resourceGroupName, nicName); err != nil {
 		return onARMAPIErrorFail(prometheusServiceNIC, err, "nic.Delete")
 	}
+
 	onARMAPISuccess(prometheusServiceNIC, "NIC deletion was successful for %s", nicName)
 	return nil
 }
 
+func (clients *azureDriverClients) waitUntilNICIsDeleted(ctx context.Context, resourceGroupName string, nicName string) error {
```
amshuman-kr, line 1075: It might be a good idea to split this function into two.

1. `waitUntil(ctx context.Context, conditionFn func() bool)` which implements just implements the loop logic (with the backoff) and just calls the `conditionFn` function inside the loop in every iteration as long as `conditionFn` return `false` and quits the loop if the `conditionFn` returns `true`.
2. `conditionNICDeleted(ctx context.Context, resourceGroupName string, nicName string) func() bool` which returns a function which implements the get call and the check if it is not found. Something like...
```go
func (clients *azureDriverClients) conditionNICDeleted(ctx context.Context, resourceGroupName string, nicName string) func() bool {
	return func() bool {
		_, err := clients.nic.Get(ctx, resourceGroupName, nicName, "")

		if err != nil && isResourceNotFoundError(err) {
			// NIC doesn't exist, hence deletion is successful
			return true
		}
		
		return true
	}
}
```

This way, you can reuse the code easily if we need to use similar logic elsewhere later (creation or ever for disks etc.). WDYT?

amshuman-kr (commented):
Thanks @prashanth26! The PR itself looks good. Some optional nit-picks and design pattern suggestions below.

pkg/driver/driver_azure.go
prashanth26: Agreed @amshuman-kr . Sure shall make this change.

pkg/driver/driver_azure.go
amshuman-kr: https://github.com/Azure/go-autorest/blob/master/autorest/error.go#L41 might be worth a shot.

pkg/driver/driver_azure.go
prashanth26: I was actually thinking of using a backoff client, didn't know which one to use. Yes, this makes sense. Let me check it out.

pkg/driver/driver_azure.go
amshuman-kr: Some of the backoff library already do something like this. Then you have to just write the condition part. Not sure of the client-go.Backoff has something like this or not.

AxiomSamarth (approved):

pkg/driver/driver_azure.go
prashanth26: Based on your suggestion i have updated the logic here to this - https://github.com/gardener/machine-controller-manager/pull/594/files#diff-36365e28193d28c39f36ac4d9172fbf8eac69553361f6179fa3184f9cad9eb0cR1329-R1342. I hope this is more appropriate now.

pkg/driver/driver_azure.go
prashanth26: Done - https://github.com/gardener/machine-controller-manager/pull/594/files#diff-36365e28193d28c39f36ac4d9172fbf8eac69553361f6179fa3184f9cad9eb0cR1077-R1083.

pkg/driver/driver_azure.go
prashanth26: I looked through using the client-go/backoff mechanism. However, it looks like it's usecase is a little different compared to ours. It seems like it's keeping a bucket for objects with different names that are being backed off. Refer below
- https://github.com/kubernetes/client-go/blob/master/util/flowcontrol/backoff.go#L37
- https://github.com/kubernetes/kubernetes/blob/019080fc4d32bc7acc994605afbc2fac416188d7/pkg/controller/daemon/daemon_controller.go#L820-L838

So looked up more appropriate libraries for our use-case. I found [this backoff library](https://github.com/cenkalti/backoff) that seems to be used by a lot of projects. And usage also seems simple and works well in our case. The Licence however is MIT, I hope that should be fine as I see other vendors using similar licenses - https://github.com/cenkalti/backoff/blob/v4/LICENSE.

pkg/driver/driver_azure.go
prashanth26: I have made `waitUntil()` as a helper package method to be able to be reused in the future [here](https://github.com/gardener/machine-controller-manager/pull/594/files#diff-9dbc206e5c2d8e9011611b1f513e2d2e73140382dae88afe6969374aab75a9edR29-R56). Also [checkNICStatus()](https://github.com/gardener/machine-controller-manager/pull/594/files#diff-36365e28193d28c39f36ac4d9172fbf8eac69553361f6179fa3184f9cad9eb0cR1050-R1075) to get the NIC status while being deleted (also can be used for creation).

prashanth26:
@AxiomSamarth - We will probably need to incorporate this change prior to the MCM azure provider release.

pkg/driver/driver_azure.go
```diff
@@ -502,48 +504,56 @@ func (d *AzureDriver) createVMNicDisk() (*compute.VirtualMachine, error) {
 	}
 	onARMAPISuccess(prometheusServiceSubnet, "subnet.Get")
 
-	/*
-		NIC creation
-	*/
+	NIC, err := clients.nic.Get(ctx, resourceGroupName, nicName, "")
```
AxiomSamarth, line 507: Hey Prashanth,

If this `Get()` call returns an error apart from 404 Not Found, it will result in nil pointer error at line 634 `VMParameters := d.getVMParameters(vmName, vmImageRef, *NIC.ID)` as it will have no reference to NIC.

Found this issue while I was incorporating this in OOT and wrote a UT for the same. Please let me know your thoughts about it.

prashanth26: Nice catch. This is why we need tests. :) 

I have tried to fix this scenario with [this commit](https://github.com/gardener/machine-controller-manager/pull/594/commits/7c9b0d9ad6f4f56b0208381404def236bc383e9d). Please let me know if it makes sense.

AxiomSamarth: Perfect! This works like a charm!

pkg/driver/driver_azure.go
```diff
@@ -1034,17 +1059,67 @@ func (clients *azureDriverClients) checkOrphanDisks(ctx context.Context, resourc
...
+		nic, err := clients.nic.Get(ctx, resourceGroupName, nicName, "")
+
+		// Case-1: If NIC should exist, check below if condition
+		if shouldExist {
+			if err == nil && nic.ID != nil {
+				// NIC exists
+				return nil
+			}
+
+			klog.V(4).Infof("NIC doesn't existance for %q", nicName)
```
AxiomSamarth, line 1073: Can we keep the log as simple as NIC <nicName> not found? The current one does not seem to be so good. Just a thought.

prashanth26: Done. Here is the change - https://github.com/gardener/machine-controller-manager/pull/594/commits/3c0e156a9f9200159cc852b45eb488cd6c153360.

amshuman-kr (approved):
Thanks a lot for the change @prashanth26! and sorry for the delay in approving the PR. LGTM

prashanth26:
Thanks @amshuman-kr . Will squash all changes, run the changes once more and then merge it.

prashanth26:
Made a final round of test. Looks good to me. Merging now.

PR #597, opened 2021-03-26
Support AWS Capacity Reservations

shin-nien:

This PR passes the ID of an [AWS Capacity Reservation (or reservation group)](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/capacity-reservations-using.html#create-cr-group) to the EC2 instances.

Sometimes AWS is unable to create new instances due to a temporary shortage of capacity in an AZ. AWS Capacity Reservations minimise the risk of stalling deployments; especially ones that use a large maxSurge (which creates more instances than you're currently running).

The default reservation target type is `open` so you can create the reservation group ahead of time without any reservations in it. If machines are created and can't match any reservations then they're created using On-Demand capacity (like the usual way).

Typically you'd create a reservation for `replicas + maxSurge` instances. Then when a deployment happens, the existing nodes are added to the reservation and now you have a buffer to rotate the new instances into the deployment.

- Didn't seem like these fields are tested for in the unit tests. Wasn't sure if that was a conscious decision or not because it's relatively simple.
- I wasn't sure if I needed to do anything with deepcopy-gen, any advice would be appreciated!

**Release note**:
```feature operator
Bumped AWS SDK version to v1.33.21
Allows passing through an AWS Capacity Reservation ID or Group ARN so capacity can be reserved to guarantee deployments that use large maxSurge values.
```

### DIFF
--- a/kubernetes/crds/machine.sapcloud.io_awsmachineclasses.yaml
+++ b/kubernetes/crds/machine.sapcloud.io_awsmachineclasses.yaml
@@ -148,6 +148,18 @@ spec:
                     name must be unique.
                   type: string
               type: object
+            capacityReservation:
+              description: The Capacity Reservation that this machine should be launched into. Either the reservation or
+                the reservation group needs to be provided - not both.
+              properties:
+                capacityReservationId:
+                  description: Target this existing Capacity Reservation.
+                  type: string
+                capacityReservationResourceGroupArn:
+                  description: Target this existing Capacity Reservation Group. Perhaps more useful than using capacity reservations
+                    directly since groups can be created much ahead of time.
+                  type: string
+              type: object
             ebsOptimized:
               type: boolean
             iam:
--- a/kubernetes/machine_classes/aws-machine-class.yaml
+++ b/kubernetes/machine_classes/aws-machine-class.yaml
@@ -27,6 +27,9 @@ spec:
 # credentialsSecretRef: # Optional - Kubernetes secret containing only provider secrets (in this case the Secret in the secretRef does not need them)
 #   name: "test-secret-credentials" # Name of the secret
 #   namespace: "default" # Namespace of secret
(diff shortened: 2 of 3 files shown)


gardener-robot-ci-1:
Thank you @shin-nien for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

prashanth26:
Hi @shin-nien ,

We really appreciate this PR. However, we are in the process of migrating all our code OOT as [mentioned here](https://github.com/gardener/machine-controller-manager/blob/master/README.md). And hence we are avoiding making any new feature addition to this code. This code would also be removed soon.

Instead I would suggest you check out the [OOT provider aws](https://github.com/gardener/machine-controller-manager-provider-aws). And file a PR over here. If you need any help with OOT understanding or Migration please let me know.

shin-nien:
Thanks @prashanth26 

Happy to close this PR but would [I still need to extend the CRD for AWSMachineClass](https://github.com/gardener/machine-controller-manager/pull/597/commits/6f96c8809d3423fed975c11c0bac00cf545178fd) in the repo? 

If so I'll drop the commits for bumping the sdk and wiring in the capacity reservation.

prashanth26:
Hi @shin-nien ,

No, the respective sample resources are present [here](https://github.com/gardener/machine-controller-manager-provider-aws/tree/master/kubernetes). You can update them here. There is only one common `MachineClass` CRD post the migration and you don't need provider-specific ones [see](https://github.com/gardener/machine-controller-manager-provider-aws/blob/master/kubernetes/machine-class.yaml).

 Also, you should find the [API specs here](https://github.com/gardener/machine-controller-manager-provider-aws/blob/master/pkg/aws/apis/aws_provider_spec.go). So in a nut shell all the changes will have to be on the OOT repo as expected. 

I had missed a few documentation files on provider AWS which I have fixed with [this PR](https://github.com/gardener/machine-controller-manager-provider-aws/pull/28).

PR #611, opened 2021-05-19
Adds finalizers on machines that are adopted by the machine controller

### LINKED ISSUE #610: MachineClass finalizers missing on migrated machine objects
**What happened**:
Azure machines were found with a few orphan NICs. On investigation, the issue was identified to machines not having finalizers after migration leading to quick deletion of machines. 

The machine to re-add finalizer when not found.

cc: @plkokanov

prashanth26:
Adds on finalizers on machines that are adopted by MCM. Without this change, it causes issues while migrating machine objects between clusters. 

Fixes #610 

- This PR basically allows adding of finalizer when when backing VM is created and status is present on it.
- I have removed the test case as it no longer makes sense in that method. We should add a test suit to the main reconcile loop in the future.

**Release note**:

```bugfix developer
Adds finalizers on machines that are adopted by the machine controller. Without this change, it causes issues while migrating machine objects between clusters. 
```

/assign @AxiomSamarth 
cc: @plkokanov

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ managevm
 controller_manager
 mcm.out
 kubectl
+.cache_ggshield
 
 # Binary files of MCM
 ./machine-controller-manager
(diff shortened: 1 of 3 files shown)


AxiomSamarth (approved):
/lgtm

PR #623, opened 2021-06-15
Restore integration test along with UTs 

AxiomSamarth:
This PR is experimental to test if the race condition really occurs during rapid scale up and scale down. The overshooting period has been modified to 2s to checkout the scenario. 

Fixes #

If this integration tests pass then we are good with replacing the integration tests with UT. Otherwise, we may have to check and fix the possible occurrence of the race condition between MachineSet Controller and MachineSafetyController.

**Release note**:

```other developer
NONE
```

### DIFF
--- a/.ci/integration-tests
+++ b/.ci/integration-tests
@@ -148,7 +148,7 @@ function run_controller() {
         --machine-safety-apiserver-statuscheck-timeout=30s \
         --machine-safety-apiserver-statuscheck-period=1m \
         --machine-safety-orphan-vms-period=30m \
-        --machine-safety-overshooting-period=300ms \
+        --machine-safety-overshooting-period=100ms \
         --leader-elect-lease-duration=1m \
         --leader-elect-renew-deadline=30s \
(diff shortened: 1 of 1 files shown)


AxiomSamarth:
Some notes for future reference of the PR/issue.

With the K8s upgrade, the rapid scale up & scale down test scenarios would fail due to `freeze` label not set rightly on the MachineSet object due to race condition between MachineSet Controller and Machine Safety Controller.

Mostly, the MachineSet Controller would win the race & scale down the machines before the MachineSafety Controller would discover the overshooted number of machines & freeze the machineDeployment and MachineSet objects. Thus the test to check the frozen state of MachineSet would fail during the integration test.

To handle/reproduce this corner case where the overlap of both controllers is expected, the `--machine-safety-overshooting-period` flag is reduced to 100ms. With this, the corresponding unit test will also continue to exist and these integration tests will continue to exist too.

prashanth26 (approved):
/lgtm

PR #627, opened 2021-06-17
Allow controlled machine drain and rollout

prashanth26:
- Machine rollouts are now more as desired with the number of replicas always maintained to `desired + maxSurge`. Earlier machines in termination were left out of this calculation but now is considered.
- Avoids blocking of drain call when the buffer is full for the `volumeAttachmentHandlers`. 
- `volumeAttachmentHandlers` are always closed now even on failure cases. 

Fixes #

**Release note**:

```feature user
Machine rollouts are now more as desired with the number of replicas always maintained to `desired + maxSurge`. Earlier machines in termination were left out of this calculation but now is considered with this change.
```
```bugfix operator
Avoids blocking of drain call when the buffer is full for the volumeAttachmentHandlers. 
```

### DIFF
--- a/pkg/controller/deployment_rolling.go
+++ b/pkg/controller/deployment_rolling.go
@@ -377,7 +377,7 @@ func (dc *controller) annotateNodesBackingMachineSets(MachineSets []*v1alpha1.Ma
 
 	for _, machineSet := range MachineSets {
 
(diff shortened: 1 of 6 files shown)


amshuman-kr (approved):
Thanks for the fixes @prashanth26. /lgtm

PR #628, opened 2021-07-09
Clarify when to apply the force deletion label

dguendisch:
The existing doc said that the label `force-deletion: "True"` should be applied when the machine object was already in deletion.
However it only works if the label is present when invoking `kubectl delete machine xxx`.

Fixes #

**Release note**:

```other operator

```


prashanth26 (approved):
/lgtm
Yes, this should be applied before the deletion. 
/invite @AxiomSamarth

PR #641, opened 2021-08-24
Fix CRD manifests

timuthy:
This PR fixes the CRD manifests for several objects that use `metav1.ObjectMeta` or `runtime.RawExtension`. It's necessary to set `x-kubernetes-preserve-unknown-fields: true` for these types because otherwise data underneath related fields is pruned and leads to validation errors in MCM.

Example:

```
I0824 09:05:30.777736       1 machine_controller_util.go:366] List machines failed on decodeProviderSpecAndSecret: machine codes error: code = [Internal] message = [Error while validating ProviderSpec [spec.disks: Required value: at least one disk is required spec.machineType: Required value: machineType is required spec.region: Required value: region is required spec.zone: Required value: zone is required spec.networkInterfaces.networkInterfaces: Required value: at least one network interface is required spec.scheduling.onHostMaintenance: Unsupported value: "": supported values: "MIGRATE", "TERMINATE" spec.serviceAccounts: Required value: at least one service account is required]]
# Please edit the object below. Lines beginning with a '#' will be ignored,
```

I added additional printer columns for `Machines` that we already use in Gardener along the way (4bb8db0636bbc35bfaa7e32e5fdf481ae746d2fb).

**Release note**:

```bugfix operator
A bug has been fixed in the pre-delivered CRD manifests for MCM (`/kubernetes/crds`). It caused data to be pruned from MCM related resources and led to reconciliation issues.
```

### DIFF
--- a/kubernetes/crds/machine.sapcloud.io_machineclasses.yaml
+++ b/kubernetes/crds/machine.sapcloud.io_machineclasses.yaml
@@ -56,6 +56,7 @@ spec:
           providerSpec:
             description: Provider-specific configuration to use during node creation.
             type: object
+            x-kubernetes-preserve-unknown-fields: true
           secretRef:
             description: SecretRef stores the necessary secrets such as credentials
               or userdata.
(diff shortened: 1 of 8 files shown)


prashanth26 (approved):
Thanks for this @timuthy . Love it <3 

/invite @AxiomSamarth

AxiomSamarth (approved):

PR #645, opened 2021-09-29
Handles Stale node object cases

himanshu-kun:
This PR implements the logic to deal with the case where a new VM starts using an old node obj which stayed in the control plane for some reason (heavy traffic on API server/ custom finalizers).
This situation happens only in AWS because the node obj name is same as domain name of the VM.
Fixes # 633

*Implemented marking machine obj as `Failed` logic
*Handled the `Adopting existing VM ...` log from reprinting.
**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -325,12 +325,12 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 			Secret:       createMachineRequest.Secret,
 		},
 	)
-	if err == nil {
+	if err == nil && (nodeName == "" || providerID == "") {
 		// Found VM with required machine name
 		klog.V(2).Infof("Found VM with required machine name. Adopting existing machine: %q with ProviderID: %s", machineName, getMachineStatusResponse.ProviderID)
 		nodeName = getMachineStatusResponse.NodeName
 		providerID = getMachineStatusResponse.ProviderID
-	} else {
+	} else if err != nil {
 		// VM with required name is not found.
 
 		machineErr, ok := status.FromError(err)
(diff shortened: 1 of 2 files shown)


pkg/util/provider/machinecontroller/machine.go
```diff
@@ -325,12 +325,12 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 			Secret:       createMachineRequest.Secret,
 		},
 	)
-	if err == nil {
+	if err == nil && (nodeName == "" || providerID == "") {
```
AxiomSamarth, line 328: I did not understand how this condition is making any difference with the previous logic as `(nodeName == "" || providerID == "")` is always `True` looking at the code flow.

himanshu-kun: I used the logic that once it enters this `if` then `nodeName` and `providerID` are sure to be initialized. so when the flow will not enter this `if` the next time because of the condition `(nodeName == "" || providerID == "")`

AxiomSamarth: But in L307 and L308, these variables are initialised with empty strings. So, I am not understanding how the subsequent reconciliation calls will behave any different.

himanshu-kun: hmm. right , I should have used 
`nodeName` -> `machine.Labels["node"]`
`providerID` -> `machine.Spec.providerID`

instead. Good catch , I hope making these changes if fine ?

himanshu-kun: I have addressed the changes here 
68a9577
It works as expected and have run the integration test locally as well.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -359,14 +354,39 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
...
+
+				// Creation was successful
+				klog.V(2).Infof("Created new VM for machine: %q with ProviderID: %q and backing node: %q", machine.Name, providerID, getNodeName(machine))
+
+				//if a stale node obj exists by the same nodeName
+				if _, err := c.nodeLister.Get(nodeName); err == nil {
+					//mark the machine obj as `Failed`
+					klog.Errorf("Stale node obj with name %q has been found", nodeName)
+
+					c.machineStatusUpdate(
```
kvmprashanth, line 366: We are updating the status of the machine without really updating the `providerID`. So the VM deletion will depend on the fact that we have tagged the VMs properly. 

The better way would be to delete the created VM at this step here, before updating the status. Or have the providerID updated so that deletion has the providerID? WDYT?

kvmprashanth:
Also please amend the release note and add an appropriate message.

PR #648, opened 2021-10-13
Logic for user triggered machine deletion

### LINKED ISSUE #390: Enable end-user to trigger the deletion/replacement of the machine.
## Improvement
Currently, the end-user of MCM only gets to see the node-objects. We have seen the cases where the end-user might want to delete or replace the specific buggy-machine in the cluster. We should probably provide a mechanism to end-user of triggering the replacement of the machine. 
- One of the seemingly natural ways could be via labels/annotations on the node-object.

AxiomSamarth:
This PR implements the logic for user triggered machine object deletion with which the user can easily replace a faulty node/unwanted node in the cluster by annotating the node with `node.machine.sapcloud.io/trigger-deletion-by-mcm="true"`

Fixes #390 

The changes are introduced only in out of tree machine controller as the in tree machine controller shall be deprecated soon.

**Release note**:

```feature user
End User can now delete the backing machine object of the node instantly by annotating the desired node with  'node.machine.sapcloud.io/trigger-deletion-by-mcm="true"`
```

### DIFF
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -234,6 +234,33 @@ func (c *controller) addNodeToMachine(obj interface{}) {
 }
 
 func (c *controller) updateNodeToMachine(oldObj, newObj interface{}) {
+
+	node := newObj.(*corev1.Node)
+	if node == nil {
+		klog.Errorf("Couldn't convert to node from object")
+		return
+	}
+
+	// check for the TriggerDeletionByMCM annotation on the node object
+	// if it is present then mark the machine object for deletion
+	if value, ok := node.Annotations[machineutils.TriggerDeletionByMCM]; ok && value == "true" {
+
+		machine, err := c.getMachineFromNode(node.Name)
+		if err != nil {
+			klog.Errorf("Couldn't fetch machine %s, Error: %s", machine.Name, err)
+			return
+		}
+
+		if machine.DeletionTimestamp == nil {
+			klog.Infof("Node %s is annotated to trigger deletion by MCM.", node.Name)
+			if err := c.controlMachineClient.Machines(c.namespace).Delete(context.Background(), machine.Name, metav1.DeleteOptions{}); err != nil {
+				klog.Errorf("Machine object %s backing the node %s could not be marked for deletion.", machine.Name, node.Name)
+				return
+			}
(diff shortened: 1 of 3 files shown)


pkg/util/provider/machinecontroller/machine.go
```diff
@@ -234,6 +234,33 @@ func (c *controller) addNodeToMachine(obj interface{}) {
...
+		if machine.DeletionTimestamp == nil {
+			if err := c.controlMachineClient.Machines(c.namespace).Delete(context.Background(), machine.Name, metav1.DeleteOptions{}); err != nil {
+				klog.Errorf("Machine object %s backing the node %s could not be marked for deletion.", machine.Name, node.Name)
+				return
+			}
+			klog.Infof("Machine object %s backing the node %s marked for deletion.", machine.Name, node.Name)
+		}
+	}
+
 	c.addNodeToMachine(newObj)
```
himanshu-kun, line 264: I have tested the code locally , it runs as expected
Just one minor pick, I see the log `Node %s is annotated to trigger deletion by MCM.` repeated multiple times in the mc-logs.
The possible reason for it might be `updateNodeToMachine` is getting called for any other update on the node obj by other controllers during deletion.
We can mitigate this by moving the log to `if machine.DeletionTimestamp == nil` code block.

himanshu-kun (changes requested):
some log specific changes , otherwise looks good, thanks for the PR

pkg/util/provider/machinecontroller/machine.go
AxiomSamarth: Good catch! Updated in the next commit

himanshu-kun:
/lgtm
test step is failing for aws with 
`Waiting 1800s for machine to join the cluster	Failed: Timeout occured while waiting for operation. Exiting Test to avoid further conflicts.`
Can you please try running the IT locally for provider-aws and verify

himanshu-kun (approved):

AxiomSamarth:
> /lgtm
> test step is failing for aws with
> `Waiting 1800s for machine to join the cluster Failed: Timeout occured while waiting for operation. Exiting Test to avoid further conflicts.`
> Can you please try running the IT locally for provider-aws and verify

I have re-triggered the build. Let's see. I am not so comfortable merging changes without successful tests on pipeline.

himanshu-kun (approved):

PR #652, opened 2021-11-07
Introduce `NodeTemplate` attribute in MachineClass

AxiomSamarth:
This PR is needed to implement the generic scale to zero feature in CA. This PR adopts the `cpu`, `gpu` and the `memory` attributes of the `expectedNodeAttribute` to be of type `resource.Quantity`.

Fixes #

The release of MCM including this change will have to be vendored in CA. So, we would want to prioritise this one.

**Release note**:

```feature user
NONE
```


pkg/apis/machine/v1alpha1/machineclass_types.go
```diff
@@ -73,20 +74,17 @@ type MachineClassList struct {
 // ExpectedNodeAttributes contains subfields to track all scale from zero resources
 type ExpectedNodeAttributes struct {
 	// CPU cores of the expected node
-	// +optional
-	CPU *int64 `json:"cpu,omitempty"`
+	CPU resource.Quantity `json:"cpu,omitempty"`
```
himanshu-kun, line 77: we can remove `omitempty` from required fields

AxiomSamarth: Yes

himanshu-kun (changes requested):
Thanks for the quick changes Samarth
* we should add a release note of change in the API maybe 
* there's not enough validation for the fields , but we have discussed that and it wouldn't be solved until we do validation by webhook
* pls let me know your comments if `instanceType` should also be made required
* Currently `region` and `zone` are part of `ExpectedNodeAttributes` but in CA they can be fetched from other places in machineClass as well. I don't remember the reason why we included them in expectedNodeAttributes. Can you explain that also.

himanshu-kun (approved):
/lgtm

pkg/apis/machine/v1alpha1/machineclass_types.go
```diff
@@ -39,18 +40,17 @@ type MachineClass struct {
 	// +optional
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
+	// +kubebuilder:validation:XPreserveUnknownFields
+	// AutoscalerNodeTemplate contains subfields to track all node resources required to scale nodegroup from zero
+	AutoscalerNodeTemplate AutoscalerNodeTemplate `json:"autoscalerNodeTemplate"`
```
himanshu-kun, line 45: we might have to make this field `optional` (and so pointer as well) for backward compatibility as stated for the `NodeTemplate` field by Tim [here](https://github.com/gardener/gardener/pull/4980#discussion_r754852941)

AxiomSamarth: Yes, I had already made a point about it.

himanshu-kun (approved):
/lgtm

AxiomSamarth:
> /lgtm

Even though this attribute is an optional attribute but the fields when mentioned are pretty critical. So, I am thinking to add a validation for the MachineClass field [here](https://github.com/gardener/machine-controller-manager/blob/f6c94b820534e7aeaf12e99f1b9fe59eea77ceb0/pkg/util/provider/machinecontroller/machine_util.go#L96) for the map `Capacity` to have the keys `cpu`, `gpu` and `memory` which are critical right now.

himanshu-kun:
> Even though this attribute is an optional attribute but the fields when mentioned are pretty critical. So, I am thinking to add a validation for the MachineClass field here for the map Capacity to have the keys cpu, gpu and memory which are critical right now.

Yes we can do that. However with this validation we won't be able to stop CA to read the machineClass. 
From what I understand,  this validation can help is when CA instructs to scale a machineDeployment from zero, and if cpu,gpu or memory not present , then machine won't scale up.
So yeah in one way it'll work.

AxiomSamarth:
Yes, that the idea. The scaleup of nodes belonging to a nodeGroup that is consuming a faulty machine class should fail and the node group should go into unhealthy state. As and when the machineClass validations are moved as a Webhook this behaviour will improve.

AxiomSamarth:
But anyway, a check will have to be added in the CA also for these attributes or else, there could be a panic.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -133,6 +133,11 @@ func (c *controller) ValidateMachineClass(ctx context.Context, classSpec *v1alph
 		return nil, nil, machineutils.ShortRetry, err
 	}
 
+	err = c.validateNodeTemplate(machineClass.NodeTemplate)
+	if err != nil {
+		return nil, nil, machineutils.ShortRetry, err
```
himanshu-kun, line 138: before return we can add an error log that node template is not valid.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -158,6 +163,25 @@ func (c *controller) getSecretData(machineClassName string, secretRefs ...*v1.Se
...
+		return nil
+	}
+
+	for _, attribute := range capacityAttributes {
+		if _, ok := nodeTemplate.Capacity[attribute]; !ok {
+			err := errors.New("MachineClass NodeTemplate Capacity should mandatorily have CPU, GPU and Memory configured")
+			return err
+		}
+	}
+
```
himanshu-kun, line 181: we can add validation for instanceType , region and zone , validating that empty string is not provided

himanshu-kun: the docstring of the method will have to be adopted accordingly

himanshu-kun (changes requested):
looks good, minor changes requested

pkg/util/provider/machinecontroller/machine_util.go
AxiomSamarth: good point.

pkg/util/provider/machinecontroller/machine_util.go
AxiomSamarth: good point too.

himanshu-kun (approved):
/lgtm

himanshu-kun (approved):
/lgtm

PR #659, opened 2021-11-14
Refactor Orphan collection logic

himanshu-kun:
* made orphan collection code more readable
* added more test cases

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/util/provider/machinecontroller/machine_safety.go
+++ b/pkg/util/provider/machinecontroller/machine_safety.go
@@ -265,10 +265,7 @@ func (c *controller) checkMachineClass(ctx context.Context, machineClass *v1alph
 	for machineID, machineName := range listMachineResponse.MachineList {
 		machine, err := c.machineLister.Machines(c.namespace).Get(machineName)
 
-		if err != nil && !apierrors.IsNotFound(err) {
-			// errors other than NotFound error
-			klog.Errorf("SafetyController: Error while trying to GET machines. Error: %s", err)
-		} else {
+		if apierrors.IsNotFound(err) || err == nil {
 			if err == nil {
 				if machine.Spec.ProviderID == machineID {
 					continue
@@ -300,8 +297,12 @@ func (c *controller) checkMachineClass(ctx context.Context, machineClass *v1alph
 			} else {
 				klog.V(2).Infof("SafetyController: Orphan VM found and terminated VM: %s, %s", machineName, machineID)
 			}
+		} else {
+			// errors other than NotFound error
+			klog.Errorf("SafetyController: Error while trying to GET machine %s. Error: %s", machineName, err)
 		}
 	}
+
 	return machineutils.LongRetry, nil
 }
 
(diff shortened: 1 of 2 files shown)


dkistner (approved):
Looks good.
Do you think that another nested if block make it really better readable? :P

himanshu-kun:
> Do you think that another nested if block make it really better readable? :P
You are right. After starting working on improving readability , I figured that there's nothing much that can be done :')

I have another flow in mind

```
if (NotFound error) or (no Error){

         if (no Error){

                //checks if machine is not being processed by Machine controller

         }

         //dummy machine obj creation and trigger deletion logic

}else{

//any other error

}
```

would this be fine?

dkistner:
Well. I have no strong opinion on that. Also your second approach seems ok for me.
As you are the primary maintainers please do the decision on your preference.

pkg/util/provider/machinecontroller/machine_safety.go
```diff
@@ -265,15 +266,18 @@ func (c *controller) checkMachineClass(ctx context.Context, machineClass *v1alph
 		machine, err := c.machineLister.Machines(c.namespace).Get(machineName)
 
 		if err != nil && !apierrors.IsNotFound(err) {
```
majst01, line 268: It would be much easier to understand without two negations, for example

```go
if err != nil {
    if apierros.IsNotFound(err) {
...
  }
}
```

himanshu-kun: Yes @majst01 , thats what I have addressed in the latest commit. Thanks for commenting.
Please let me know your thoughts regarding the new flow.

pkg/util/provider/machinecontroller/machine_safety.go
```diff
@@ -264,16 +265,16 @@ func (c *controller) checkMachineClass(ctx context.Context, machineClass *v1alph
...
+		if apierrors.IsNotFound(err) || err == nil {
+			if err == nil {
+				if machine.Spec.ProviderID == machineID {
+					continue
+				}
+				// machine obj is still being processed by the machine controller
+				if machine.Status.CurrentStatus.Phase == "" || machine.Status.CurrentStatus.Phase == v1alpha1.MachineCrashLoopBackOff {
+					klog.V(3).Infof("SafetyController: Machine object %q with backing nodeName %q , providerID %q is being processed by machine controller, hence skipping", machine.Name, getNodeName(machine), getProviderID(machine))
+					continue
+				}
```
AxiomSamarth, line 277: The first `if apierrors.IsNotFound(err) || err == nil` nests an `if err == nil` which in turn nests two if statements which will do something and continues the loop skipping everything below it. So, why not make that check of `err == nil` independent before the first if block; followed by an if block to check the NotFound error?

himanshu-kun: are you suggesting this

```
if (no error){
 // checks for continue
}
if (NotFound error){
// dummy object creation and delete
}else{
//any other error
}
```

In this flow we'll miss the case where `err==nil && machine.Spec.ProviderID!=machineID && (machine not reconciled by MC)` , and we want to remove the VM for this case also.
Yes to solve that we could follow below flow which is the same flow as I have pushed , with just the if block taken out as you suggested.

```
if (no error){
 // checks for continue
}
if (NotFound error) or (no error){
// dummy object creation and delete
}else{
//any other error
}
```

Let me know your preference :)

AxiomSamarth (approved):
/lgtm

PR #661, opened 2021-12-21
Set `CONTROL_CLUSTER_NAMESPACE` env variable for integration tests

himanshu-kun:
for IT the env variable `CONTROL_CLUSTER_NAMESPACE` can remain unset(it can be set explicitly though by user) even when the code interprets the value of cluster namespace. 
But in some other parts of code to know the cluster namespace its beneficial if `CONTROL_CLUSTER_NAMESPACE` is set.
This PR sets the env variable.
Fixes #

**Release note**:

```other operator
NONE
```


AxiomSamarth (approved):
/lgtm

PR #662, opened 2021-12-22
Switch leader election to endpointleases

acumino:
/kind enhancement

CC: @ialidzhikov 

**Release note**:

```other operator
The default leader election resource lock of `machine-controller-manager` has been changed from `endpoints` to `endpointsleases`.
```
```breaking operator
Components that deploy the `machine-controller-manager` will now have to adapt the RBAC rules to allow `machine-controller-manager` to maintain its leader election resource lock in `leases` as well.
```

### DIFF
--- a/pkg/util/client/leaderelectionconfig/config.go
+++ b/pkg/util/client/leaderelectionconfig/config.go
@@ -44,7 +44,8 @@ func DefaultLeaderElectionConfiguration() options.LeaderElectionConfiguration {
 		LeaseDuration: metav1.Duration{Duration: DefaultLeaseDuration},
 		RenewDeadline: metav1.Duration{Duration: DefaultRenewDeadline},
 		RetryPeriod:   metav1.Duration{Duration: DefaultRetryPeriod},
-		ResourceLock:  rl.EndpointsResourceLock,
+		// TODO(acumino): migrate default to `leases` in one of the next releases
+		ResourceLock: rl.EndpointsLeasesResourceLock,
 	}
 }
 
@@ -69,5 +70,6 @@ func BindFlags(l *options.LeaderElectionConfiguration, fs *pflag.FlagSet) {
 		"of a leadership. This is only applicable if leader election is enabled.")
 	fs.StringVar(&l.ResourceLock, "leader-elect-resource-lock", l.ResourceLock, ""+
 		"The type of resource object that is used for locking during "+
-		"leader election. Supported options are `endpoints` (default) and `configmap`.")
+		"leader election. Supported options are 'endpoints', 'configmaps', "+
+		"'leases', 'endpointsleases' and 'configmapsleases'.")
 }
--- a/kubernetes/deployment/in-tree/control-cluster-role.yaml
+++ b/kubernetes/deployment/in-tree/control-cluster-role.yaml
@@ -42,3 +42,19 @@ rules:
    - patch
    - update
    - watch
+-  apiGroups:
+   - coordination.k8s.io
+   resources:
+   - leases
+   verbs:
+   - create
+-  apiGroups:
+   - coordination.k8s.io
+   resources:
+   - leases
+   resourceNames:
+   - machine-controller-manager
+   verbs:
+   - get
+   - watch
+   - update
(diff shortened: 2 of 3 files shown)


pkg/util/client/leaderelectionconfig/config.go
```diff
@@ -69,5 +70,5 @@ func BindFlags(l *options.LeaderElectionConfiguration, fs *pflag.FlagSet) {
 		"of a leadership. This is only applicable if leader election is enabled.")
 	fs.StringVar(&l.ResourceLock, "leader-elect-resource-lock", l.ResourceLock, ""+
 		"The type of resource object that is used for locking during "+
-		"leader election. Supported options are `endpoints` (default) and `configmap`.")
+		"leader election. Supported options are `endpointsleases` (default).")
```
ialidzhikov, line 73: We don't have to remove already supported options from the doc string of the flag.

```suggestion
	fs.StringVar(&l.ResourceLock, "leader-elect-resource-lock", l.ResourceLock, ""+
		"The type of resource object that is used for locking during "+
		"leader election. Supported options are 'endpoints', 'configmaps', "+
		"'leases', 'endpointsleases' and 'configmapsleases'.")
```

FYI, this seems to be copied from https://github.com/kubernetes/kubernetes/blob/c75d254beb662ea73736d2f123d0382b7f91dfbd/staging/src/k8s.io/component-base/config/options/leaderelectionconfig.go#L43-L46.

ialidzhikov (changes requested):
Generally lgtm, minor nit.

ialidzhikov:
@acumino found that the RBAC under `kubernetes/deployment` has to be adapted accordingly. Nice catch!

himanshu-kun (approved):
/lgtm
is there a resource by the name `endpointsleases` also, I can find `leases` and `endpoint` resources

acumino:
> /lgtm
> is there a resource by the name `endpointsleases` also, I can find `leases` and `endpoint` resources

No, we want to migrate from endpoint to leases. 
We exactly don't want to change it randomly but in two phases:
(1) acquire old object lock, once acquired, acquire new object lock; only when both acquired we have a lock and can proceed
(2) (next release) since now master has to acquire the lock on the new object (lease) to be master, we can now delete the need to acquire the old object (endpoints or configmap).
And now we're using lease as it should be.

himanshu-kun:
> No, we want to migrate from endpoint to leases.
We exactly don't want to change it randomly but in two phases:
(1) acquire old object lock, once acquired, acquire new object lock; only when both acquired we have a lock and can proceed
(2) (next release) since now master has to acquire the lock on the new object (lease) to be master, we can now delete the need to acquire the old object (endpoints or configmap).
And now we're using lease as it should be.

Thanks @acumino , for the clear explanation:thumbsup:

pkg/util/client/leaderelectionconfig/config.go
```diff
@@ -69,5 +70,6 @@ func BindFlags(l *options.LeaderElectionConfiguration, fs *pflag.FlagSet) {
 		"of a leadership. This is only applicable if leader election is enabled.")
 	fs.StringVar(&l.ResourceLock, "leader-elect-resource-lock", l.ResourceLock, ""+
 		"The type of resource object that is used for locking during "+
-		"leader election. Supported options are `endpoints` (default) and `configmap`.")
+		"leader election. Supported options are 'endpoints', 'configmaps', "+
+		"'leases', 'endpointsleases' (default) and 'configmapsleases'.")
```
ialidzhikov, line 74:
```suggestion
		"'leases', 'endpointsleases' and 'configmapsleases'.")
```


The default value is already added/documented by the underlying library. You can check this out by running:

```
docker run eu.gcr.io/gardener-project/gardener/machine-controller-manager:v0.42.0 --help

      --leader-elect-resource-lock endpoints                    The type of resource object that is used for locking during leader election. Supported options are endpoints (default) and `configmap`. (default "endpoints")
```

There is no need to document it one more time in the doc string of this flag.

ialidzhikov (approved):
/lgtm

AxiomSamarth (approved):

ialidzhikov:
@acumino can you prepare draft PRs to provider-extensions with the required RBAC changes? Thanks in advance.

PR #667, opened 2021-12-28
Trigger orphan collection on OutOfRange machine status

himanshu-kun:
Till now orphan collection triggers on machine obj deletion or after 30min.
But to handle cases which arise due to eventual inconsistency in AWS , where multiple VMs can start backing a machine obj, in that case to stop machine-controller to reach the VM quota , orphan collection is now also triggered if machine object is updated with a `OutOfRange` error description.
The update of machine obj to `OutOfRange` can only happen by `GetMachineStatus` call , which is called only during a machine creation or deletion.

So in summary , orphan collection will be triggered in three cases now
* during creation if multiple VM are seen
* during deletion
* after 30min of last logic run for each machine object.

Fixes #

I haven't enqueued in the orphan queue if status contains OOR error now , but only if the machine obj status `earlier didn't have ` OOR error and now have it. 
This is to deal with scenario where machine obj has OOR error already and any other status update event triggers. So if we enqueue just on presence of OOR error ,then orphan logic will be triggered almost all the time

**Release note**:

```feature operator
orphan collection is also triggered if machine obj is updated with having multiple backing VMs
```

### DIFF
--- a/pkg/util/provider/machinecontroller/controller.go
+++ b/pkg/util/provider/machinecontroller/controller.go
@@ -197,7 +197,9 @@ func NewController(
 	controller.machineSafetyOrphanVMsQueue.Add("")
 	controller.machineSafetyAPIServerQueue.Add("")
 	machineInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-		// deleteMachineToSafety makes sure that orphan VM handler is invoked
+		// updateMachineToSafety makes sure that orphan VM handler is invoked on some specific machine obj updates
+		UpdateFunc: controller.updateMachineToSafety,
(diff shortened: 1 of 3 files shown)


himanshu-kun:
@dkistner could you provide you review as well:)

himanshu-kun:
Merging as a new release is awaited by Core colleagues

himanshu-kun:
@AxiomSamarth could you provide you approving review and merge?

dkistner (commented):
Hi @himanshu-kun,

sorry for the delay. From my point of view this looks good.
But I would really like to have @AxiomSamarth opinion also on this.

AxiomSamarth (approved):
/lgtm

PR #677, opened 2022-01-25
Node Name column while printing machine objects

himanshu-kun:
This PR enhances CRD so that on `k get machines` the nodeName is also printed. This is to ease getting nodeName for provider like aws
Fixes #

**Release note**:

```operator other
NONE
```


AxiomSamarth (approved):
/lgtm

Issue #679, opened 2022-02-02
API server unreachable through istio could lead to deletion of all nodes


himanshu-kun:
**How to categorize this issue?**

/area performance
/kind bug
/priority 2

**What happened**:
An issue was noticed with the health check of MCM . The kubelet of nodes couldn't reach KAPI as there was problem with istio service of the seed. The kubelet uses the internal address of api-server to reach KAPI. This lead to all machines being in unknown state for over 10 min ,and led to MCM deleting the nodes.

**What you expected to happen**:
MCM should not delete nodes in such scenario and should be aware that API server is unreachable by kubelet. Otherwise it should not mark all machines for deletion at the same time, if they are unhealthy but one by one.

**How to reproduce it (as minimally and precisely as possible)**:
Not tested but from my understanding
- Take down the istio svc, this leads to reaching using internal address impossible
- notice KCM marking nodes as unknown before its taken down by DWD (this is a race condition and this example will work only if DWD takes down KCM after KCM marks nodes as unknown)
- notice MCM marking machines as unknown because corresponding nodes had been marked unknown
- after 10 min (machineHealthTimeout) machine will be marked Failed 

**Anything else we need to know?**:

- currently DWD also doesn't have mechanism to take down MCM
- MCM has API server availability check present but that uses in-cluster svc address of KAPI and so it doesn't return failure in this case.

**Environment**:

- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- Others:

himanshu-kun:
cc @vlerenc @ashwani2k

himanshu-kun:
fixed with https://github.com/gardener/machine-controller-manager/pull/683 for MCM

PR #683, opened 2022-02-15
Limit #machines marked failed at a time due to being Unhealthy

### LINKED ISSUE #354: Handle infrastructure meltdown scenarios
## Issue
We have observed the situations where, due to infrastructure outages/issues, most of the machines in the cluster could become `unhealthy` simultaneously. In such cases, the MCM remediation system could kick-in and start trying to replace all affected machines. This would make the overall-state no better but actually worse.

## Probable Solution
We need to build a robust solution to handle such cases. One of the possible solutions could be to freeze the MCM when more than ~X% [~60%/70%] of the machines are unhealthy together.

himanshu-kun:
This PR deals with scenario where meltdown can happen . Till now MC was marking all healthTimeout machines as Failed, at the same time in case API server is not reachable through the domain which kubelet uses, or any other such reason , and this could lead to meltdown
The current PR limits the number of failed machines marking to 1 at a time per machinedeployment, using channels.

Fixes #354 
#679 

- number of locks won't exceed number of machinedeployments 
**Release note**:

```bugfix user
MCM now marks 1 machine per machineDeployment as Failed at a time in case of healthTimeout. This is introduced to deal with meltdown scenario
```


pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1356,3 +1519,16 @@ func getProviderID(machine *v1alpha1.Machine) string {
...
+func getMachineOwnerNames(machine *v1alpha1.Machine) (string, string) {
+	machineSetName := machine.OwnerReferences[0].Name
+	var machineDeployName string
+	msLen := len(machineSetName)
+	for i := msLen - 1; i >= 0; i-- {
+		if machineSetName[i] == '-' {
+			machineDeployName = machineSetName[:i]
+			break
+		}
+	}
```
AxiomSamarth, line 1532: I would recommend invoking a `Get `call on the `machineSet` and referring the owner reference of this `machineSet `to fetch the name of `MachineDeployment` instead of string handling with `-` as delimiter.

himanshu-kun: Hey Samarth :D 
Yeah I was thinking the same, but since this is a part of MC , so I would have to create a seperate clientset for machineDeployments , and since MC shouldn't be aware of machineDeployment so I wrote this way.

AxiomSamarth: Oh! That's quite tricky. In that case, I have another "more Golang" way of putting this up as a code as below.

```suggestion
	machineSetNameParts := strings.Split(machineSetName, "-")
	machineDeployName := strings.Join(machineSetNameParts[:len(machineSetNameParts)-1], "-")
```

Verify the index logic once. I have directly put up the code here.

himanshu-kun: Thanks Samarth, it looks quite clean now. I have verified its working correctly!

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -708,31 +718,78 @@ func (c *controller) reconcileMachineHealth(ctx context.Context, machine *v1alph
...
 					"Machine %s is not healthy since %s minutes. Changing status to failed. Node Conditions: %+v",
 					machine.Name,
 					timeOutDuration,
 					machine.Status.Conditions,
 				)
-			}
 
-			// Log the error message for machine failure
-			klog.Error(description)
+				machineSetName, machineDeployName := getMachineOwnerNames(machine)
```
MartinWeindel, line 746: The method `getMachineOwnerNames` uses the naming convention to extract the machine deployment name from the name of the first owner reference, which seems to be the name of the machine set. This is a rather indirect way of access.
If I understand correctly, the machine has a label `name` and its value is the machine deployment name. Wouldn't it be better to use that way? Later the label is used anyway to find pending machines for this deployment.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -708,31 +718,78 @@ func (c *controller) reconcileMachineHealth(ctx context.Context, machine *v1alph
...
+						<-ch
+						return machineutils.ShortRetry, err
+					}
+					if markable {
+						retryPeriod, updated, err := c.updateMachineToFailedState(ctx, description, machine, clone)
+						//wait for cache sync
+						if updated && !c.waitForFailedMachineCacheUpdate(machine, cacheUpdateTimeout) {
+							//waiting 1 sec since nothing else can be done if cache update is failing
+							klog.Infof("cache sync returned false, waiting 1 sec , machineName=%q", machine.Name)
+							<-time.After(1 * time.Second)
```
MartinWeindel, line 776:
```suggestion
							time.Sleep(1 * time.Second)
```
for better readability, as `<-time.After` only has an advantage if it used as a select case.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -708,31 +718,78 @@ func (c *controller) reconcileMachineHealth(ctx context.Context, machine *v1alph
...
+				}
+				ch := val.(chan struct{})
+
+				klog.Errorf("Number of goroutines now %d\n", runtime.NumGoroutine())
+				lockAcquireCtx, cancelFn := context.WithTimeout(ctx, lockAcquireTimeout)
+				defer cancelFn()
+
+				select {
+				case ch <- struct{}{}:
+					klog.Infof("Acquired lock ,machineName=%q , locks left to acquire %d", machine.Name, cap(ch)-len(ch))
```
MartinWeindel, line 763:
```suggestion
					klog.Infof("Acquired lock, machineName=%q", machine.Name)
```

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -708,31 +718,78 @@ func (c *controller) reconcileMachineHealth(ctx context.Context, machine *v1alph
...
+						klog.Errorf("Couldn't check if machine can be marked as Failed. Error: %q", err)
+						<-ch
+						return machineutils.ShortRetry, err
+					}
+					if markable {
+						retryPeriod, updated, err := c.updateMachineToFailedState(ctx, description, machine, clone)
+						//wait for cache sync
+						if updated && !c.waitForFailedMachineCacheUpdate(machine, cacheUpdateTimeout) {
+							//waiting 1 sec since nothing else can be done if cache update is failing
+							klog.Infof("cache sync returned false, waiting 1 sec , machineName=%q", machine.Name)
```
MartinWeindel, line 775:
```suggestion
							klog.Infof("cache sync returned false, waiting 1 sec, machineName=%q", machine.Name)
```

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1339,6 +1396,118 @@ func (c *controller) UpdateNodeTerminationCondition(ctx context.Context, machine
...
+					noPhase++
+				} else if machine.Status.CurrentStatus.Phase == v1alpha1.MachineCrashLoopBackOff {
+					crashLooping++
+				}
+			} else {
+				terminating++
+				// counting terminated machine twice in `inProgress` to avoid case where there is delay by MS controller
+				// in adding new machine object and terminating machines are also gone.
+				inProgress++
+			}
```
MartinWeindel, line 1471: switch/case would improve readability

MartinWeindel (commented):
Overall logic looks good to me.  I have some suggestions for improvement.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1339,6 +1396,118 @@ func (c *controller) UpdateNodeTerminationCondition(ctx context.Context, machine
...
+
+func (c *controller) canMarkMachineFailed(machineSetName, machineDeployName, namespace string, maxReplacements int) (bool, error) {
+	var (
+		list     = []string{machineDeployName}
+		selector = labels.NewSelector()
+		req, _   = labels.NewRequirement("name", selection.Equals, list)
+	)
+
+	selector = selector.Add(*req)
+	//listing all machines for machinedeployment
```
unmarshall, line 1438: Correct the comments. Add a space just after `//`

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1339,6 +1396,118 @@ func (c *controller) UpdateNodeTerminationCondition(ctx context.Context, machine
...
+	machineList, err := c.machineLister.Machines(namespace).List(selector)
+	if err != nil {
+		return false, err
+	}
+
+	klog.Infof("num machines in machineDeploy=%s is %d", machineDeployName, len(machineList))
+	//inProgress keeps count of number of machines which are counted as `getting replaced`
```
unmarshall, line 1445: Correct the comments. Add a space just after //

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1339,6 +1396,118 @@ func (c *controller) UpdateNodeTerminationCondition(ctx context.Context, machine
...
+		// Keep retrying until update goes through
+		klog.Errorf("update failed for machine %q in function. Retrying, error: %q", machine.Name, err)
+	} else {
+		updated = true
+		klog.Infof("Machine State has been updated for %q with providerID %q and backing node %q", machine.Name, getProviderID(machine), getNodeName(machine))
+		// Return error for continuing in next iteration
+		err = fmt.Errorf("machine creation is successful. Machine State has been UPDATED")
+	}
+
+	return machineutils.ShortRetry, updated, err
```
unmarshall, line 1427: If you are always going to return `machineutils.ShortRetry` then why bother returning this from this method. The caller should do that instead. It would make sense to return it in case retry could have different values or no value (in a specific condition). Any reason to always return `machineutils.ShortRetry`?

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -708,31 +718,78 @@ func (c *controller) reconcileMachineHealth(ctx context.Context, machine *v1alph
 					machine.Name,
 					timeOutDuration,
 				)
+				// Log the error message for machine failure
```
unmarshall, line 721: func `reconcileMachineHealth` is way too large. Ideally it should be broken down into smaller functions. Make this method very hard to test and also hard to read. Some quick refactoring for better readability can at least be done:

1. if err != nil attempting to execute c.nodeLister.Get then that should be on the top (early exit) from the method with smaller processing logic.
2.  combine the conditions IsNotFound and else (general error) case under this if err != nil case. Makes it much more readable.
3. `checkCreationTimeout` indicates a verb however what it holds is a value of whether the machine is in pending state. A better  name would have been `isMachinePending` . Right now the intent has been mixed with the actual value that it holds.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -708,31 +718,78 @@ func (c *controller) reconcileMachineHealth(ctx context.Context, machine *v1alph
...
-			}
 
-			// Log the error message for machine failure
-			klog.Error(description)
+				machineSetName, machineDeployName := getMachineOwnerNames(machine)
+				//getting lock for machineDeployment this machine belongs to, creating lock if absent
+				tempCh := make(chan struct{}, 1)
+				val, loaded := c.safetyOptions.HealthChanMap.LoadOrStore(machineDeployName, tempCh)
+				if loaded {
+					close(tempCh)
```
unmarshall, line 751: any reason that you close the channel if loaded is true?

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -708,31 +718,78 @@ func (c *controller) reconcileMachineHealth(ctx context.Context, machine *v1alph
...
 					machine.Status.Conditions,
 				)
-			}
 
-			// Log the error message for machine failure
-			klog.Error(description)
+				machineSetName, machineDeployName := getMachineOwnerNames(machine)
+				//getting lock for machineDeployment this machine belongs to, creating lock if absent
+				tempCh := make(chan struct{}, 1)
+				val, loaded := c.safetyOptions.HealthChanMap.LoadOrStore(machineDeployName, tempCh)
```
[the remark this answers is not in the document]
himanshu-kun: Yes I thought about that , but that would add further complexity in the code. I didn't do it for the following reasons:
* a channel of struct consumes almost zero memory according to some blogs
* Generally number of machineDeployments don't exceed 50 even in largest of clusters, so map size won't be more than 50
* a restart of MCM (which generally occurs due to hibernation) would release the memory by map.
* to delete an entry , we'll need a watch on machineDeployment to see if they are deleted, and as part of machineController we don't have watchers and listers for it.

Please let me know if you still have concerns

pkg/util/provider/machinecontroller/machine_util.go
himanshu-kun: Yes you are right, I'll use the label instead then

pkg/util/provider/machinecontroller/machine_util.go
himanshu-kun: In `waitForFailedMachineCacheUpdate()` I explicitly wait for the cache update as the reflector cache update is not that reliable. If it doesn't get updated in `waitForFailedMachineCacheUpdate()` (either due to timeout or any error) , then I wait for this `1sec` and thus depending only on the reflector update now.

pkg/util/provider/machinecontroller/machine_util.go
himanshu-kun: This log was introduced for debugging purposes, so I'll remove this at the end.

pkg/util/provider/machinecontroller/machine_util.go
himanshu-kun: Yes these counters are for better logging. I would use switch case for the same.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -56,6 +60,12 @@ var (
 	emptyMap = make(map[string]string)
 )
 
+const (
+	maxReplacements    = 1
```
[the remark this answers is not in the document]
himanshu-kun: Sure, I'll open an issue to make this value configurable and discussion on this, as I am still unsure on the best way to pass down the values.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -56,6 +60,12 @@ var (
...
+	lockAcquireTimeout = 1 * time.Second
```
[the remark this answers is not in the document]
himanshu-kun: I have tested with 100 node cluster(with around 5 machineDeployment) and no timeout log was found, so I think its sufficient. One reason is because we have lock per machineDeployment and number of worker goroutines are 50 , so there are not more than 50 goroutines waiting to acquire lock per machineDeployment in worst case.
I'll test this worst case again and will post the observation here.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -56,6 +60,12 @@ var (
...
+	cacheUpdateTimeout = 1 * time.Second
```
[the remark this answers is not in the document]
himanshu-kun: For now it was a guess work . In reality its around 2sec , because we give the reflector update further 1 sec in case this `cacheUpdateTimeout` happens in `waitForFailedMachineCacheSync`.
Open for suggestions

pkg/util/provider/options/types.go
```diff
@@ -123,6 +124,9 @@ type SafetyOptions struct {
 	// MachineControllerFrozen indicates if the machine controller
 	// is frozen due to Unreachable APIServers
 	MachineControllerFrozen bool
+	//HealthChanMap store the channel per machinedeployment which is used to limit removal of `health timed out` machines
+	//Default limit is 1 machine per machinedeployment
+	HealthChanMap *sync.Map
```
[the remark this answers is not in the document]
himanshu-kun: 1. I kept it in SafetyOptions as meltdown protection is more of a safety feature of MCM, and other safety related variables like `MachineControllerFrozen` is also there.
2. normal maps in golang are not thread safe. And in `HealthChanMap` we are doing both read and write(when channel is not present in map for a machineDeployment, so we create it and write to map) . So it was like high read and less write like situation. Yes there is a performance impact no doubt.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1339,6 +1396,118 @@ func (c *controller) UpdateNodeTerminationCondition(ctx context.Context, machine
...
+
+	klog.Errorf("machineDeployName=%q , terminating=%d , failed=%d , pending=%d , noPhase=%d , crashLooping=%d , extraCountedProgress=%d", machineDeployName, terminating, failed, pending, noPhase, crashLooping, terminating)
+
+	if inProgress < maxReplacements {
+		return true, nil
+	}
+	return false, nil
+}
+
+func (c *controller) waitForFailedMachineCacheUpdate(machine *v1alpha1.Machine, timeout time.Duration) bool {
```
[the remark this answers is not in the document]
himanshu-kun: ok, will do that.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1339,6 +1396,118 @@ func (c *controller) UpdateNodeTerminationCondition(ctx context.Context, machine
...
+	_, err := c.controlMachineClient.Machines(clone.Namespace).UpdateStatus(ctx, clone, metav1.UpdateOptions{})
+	updated := false
+	if err != nil {
+		// Keep retrying until update goes through
+		klog.Errorf("update failed for machine %q in function. Retrying, error: %q", machine.Name, err)
+	} else {
+		updated = true
+		klog.Infof("Machine State has been updated for %q with providerID %q and backing node %q", machine.Name, getProviderID(machine), getNodeName(machine))
+		// Return error for continuing in next iteration
+		err = fmt.Errorf("machine creation is successful. Machine State has been UPDATED")
```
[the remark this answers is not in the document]
himanshu-kun: will remove the error message.

pkg/util/provider/machinecontroller/machine_util.go
himanshu-kun: thanks, will add these refactoring

pkg/util/provider/machinecontroller/machine_util.go
himanshu-kun: `HealthChanMap` is a map with key as machineDeployment name and value as the lock/channel for that machineDeployment. In these line we try to create entry for the machineDeployment if not there in the map. If the entry is there then I delete the tempCh (which I created to make entry in the map for that machineDeployment).

pkg/util/provider/machinecontroller/machine_util.go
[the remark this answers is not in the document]
himanshu-kun: issue opened -> https://github.com/gardener/machine-controller-manager/issues/688

pkg/util/provider/machinecontroller/machine_util.go
himanshu-kun: addressed in 2aa4416f08be317fc00c9b12355fe15ea986b5a8

pkg/util/provider/machinecontroller/machine_util.go
himanshu-kun: addressed in 2aa4416f08be317fc00c9b12355fe15ea986b5a8

pkg/util/provider/machinecontroller/machine_util.go
himanshu-kun: addressed in 2aa4416f08be317fc00c9b12355fe15ea986b5a8

pkg/util/provider/machinecontroller/machine_util.go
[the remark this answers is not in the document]
himanshu-kun: addressed in 2aa4416f08be317fc00c9b12355fe15ea986b5a8

pkg/util/provider/machinecontroller/machine_util.go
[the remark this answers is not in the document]
himanshu-kun: addressed in 2aa4416f08be317fc00c9b12355fe15ea986b5a8

pkg/util/provider/machinecontroller/machine_util.go
himanshu-kun: addressed in 2aa4416f08be317fc00c9b12355fe15ea986b5a8

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -708,31 +718,78 @@ func (c *controller) reconcileMachineHealth(ctx context.Context, machine *v1alph
...
+					State:          v1alpha1.MachineStateFailed,
+					Type:           machine.Status.LastOperation.Type,
+					LastUpdateTime: metav1.Now(),
+				}
+				clone.Status.CurrentStatus = v1alpha1.CurrentStatus{
+					Phase: v1alpha1.MachineFailed,
+					//TimeoutActive:  false,
+					LastUpdateTime: metav1.Now(),
+				}
+				objectRequiresUpdate = true
```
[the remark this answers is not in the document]
himanshu-kun: addressed in 2aa4416f08be317fc00c9b12355fe15ea986b5a8

pkg/util/provider/options/types.go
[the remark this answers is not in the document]
unmarshall: `SafetyOptions` struct contains configurable parameters but `HealthChanMap` is implementation details on how do you stagger the removal of `health timed out` machines. This is an implementation leak.

pkg/util/provider/machinecontroller/machine_util.go
himanshu-kun: increased the wait time to 10sec to deal with it currently.

pkg/util/provider/options/types.go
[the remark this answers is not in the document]
MartinWeindel: I agree, the `HealthChanMap` contains state and should be moved to the other state of the `machinecontroller.controller`

pkg/util/provider/machinecontroller/machine_util.go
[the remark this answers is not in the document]
unmarshall: If the number of machine deployment are going to be somewhere around 50 then i do not see much of an issue, however this is not in your control. The cost of having channels is surely not zero. For this PR we can go with the assumption that this cost is tolerable. However care should be taken to release channels and free memory when its no longer required.

You have mentioned 2 options: Using lister for `MachineDeployment` or lister for 'Machine` - get deployments from labels. Either of these can be used to determine deployments which are no longer there.

If you wish to implement a solution without depending upon lister/watcher then you can do the following (high level approach) - can be detailed if desired.

- Create an interface - DeploymentPermit which has a single method - tryPermit(timeout time.Duration) or tryPermit(ctx contex.Context)
- Have a sync.Map - key = deploymentName, value = struct { c chan struct{} lastLockAcquisitionTime time.Time}
- Implement getPermit on sync.Map which will attempt to acquire the lock within the given duration
- Implement a janitor time.NewTicker which checks periodically for any entry whose lastLockAcquisitionTIme is beyond a threshold (lets say 30 mins), if yes checks that length of channel = 0 (safety check) and removes this entry from the Map after closing the channel.

In your code you anyways use sync.Map.loadOrStore which will again create an entry if its not present.

pkg/util/permits/permits.go
```diff
@@ -0,0 +1,175 @@
...
+	DeletePermits(key string)
+	Close()
+}
+
+type permit struct {
+	c                      chan struct{}
+	lastAcquiredPermitTime time.Time
+}
+
+func NewPermitGiver(stalePermitKeyTimeout time.Duration, janitorFreqency time.Duration) PermitGiver {
```
AxiomSamarth, line 43: Nit - Typo in the variable name `janitorFrequency`

himanshu-kun: addressed in [59d31c0](https://github.com/gardener/machine-controller-manager/pull/683/commits/59d31c00569227a7e0466c1b4da4fe981f412562)

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1349,10 +1473,81 @@ func setTerminationReasonByPhase(phase v1alpha1.MachinePhase, terminationConditi
...
+		case lockC <- struct{}{}:
+			return true
+		case <-ctx.Done():
+			return false
+		}
+	}
+}
+
+func (c *controller) tryMarkingMachineFailed(ctx context.Context, machine, clone *v1alpha1.Machine, machineDeployName, description string, lockAcquireTimeout time.Duration) (machineutils.RetryPeriod, error) {
+	defer c.permitGiver.ReleasePermit(machineDeployName)
```
unmarshall, line 1491: Shouldn't this be only done if you were able to acquire a permit? So essentially move this inside the if condition after you are able to acquire a permit

himanshu-kun: I was of the opinion that anyways `ReleasePermit` has the logic to deal with permit not acquired case, but yes moving it inside would be better.

himanshu-kun: addressed in 
d4e34d8

unmarshall (approved):
Looks good.

PR #685, opened 2022-03-03
Scale down annotation value corrected

himanshu-kun:
The annotation value of scale down annotation provided by CA has been changed from `True` to `true`. This will enable MCM to stop CA from scaling down during rolling update of machines.
Fixes #

**Release note**:

```bugfix user
The value for key `cluster-autoscaler.kubernetes.io/scale-down-disabled` placed by MCM is now `true` and not `True`. This typo stopped MCM from disabling CA from scaling down during rolling update.
```


unmarshall (approved):
/lgtm

PR #697, opened 2022-04-07
Refactor ClusterAutoscaler annotations to a separate package

kon-angelo:

Refactors autoscaler annotations to a separate package. This is necessary because with the inclusion of https://github.com/gardener/machine-controller-manager/pull/683 build targets for mcm-extensions that  depend on https://github.com/gardener/machine-controller-manager/tree/master/pkg/util/provider/app to start the machine controller will now include `gardener/machine-controller-manager/tree/master/pkg/driver` in their build.

This is problematic because the extensions now have a dependency on the infrastructure provider clients bundled with MCM that may conflict with their versions. It is better if this import chain is eliminated.

Fixes #

**Release note**:

```other operator
Refactor autoscaler annotation to a separate package
```

### DIFF
--- a/pkg/controller/autoscaler/types.go
+++ b/pkg/controller/autoscaler/types.go
@@ -1,0 +1,12 @@
+package autoscaler
+
+const (
+	// ClusterAutoscalerScaleDownDisabledAnnotationKey annotation to disable the scale-down of the nodes.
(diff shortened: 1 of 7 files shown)


himanshu-kun (approved):
/lgtm

PR #707, opened 2022-04-21
Update stale node check to check if node names are equal

kon-angelo:

In Azure the call to create a VM is async. The operation is periodically polled until Azure reports a successful VM creation. Because of the delay between polling periods it may be that the node has joined the cluster by the time the creation op succeeds. In that case, because of #645 the VM would be immediately deleted and a brief loop of VM "thrashing" starts where MCM will retry creating the machine with the same name and be immediately deleted. The extra condition added here prevents this behaviour.  

Fixes #

**Release note**:

```other operator
Stale node detector skips marking a node as stale if the node name and machine name are equal.
```


himanshu-kun (approved):
/lgtm

PR #709, opened 2022-04-22
[rel-v0.44] Automated cherry pick of #707: Update stale node check to check if node names are equal

kon-angelo:
/kind/bug

Cherry pick of #707 on rel-v0.44.

#707: Update stale node check to check if node names are equal

**Release Notes:**
```other operator
Stale node detector skips marking a node as stale if the node name and machine name are equal.
```


himanshu-kun (approved):
/lgtm

PR #711, opened 2022-05-03
Migrate leader election to leases

acumino:
/kind enhancement

Part of https://github.com/gardener/gardener/issues/4742

/cc @ialidzhikov 

**Release note**:

```breaking operator
The default leader election resource lock of `machine-controller-manager` has been changed from `endpointsleases` to `leases`.
Please make sure, that you had at least `machine-controller-manager@v0.43.0` running before upgrading to `v0.46.0`, so that it has successfully acquired leadership with the hybrid resource lock (`endpointsleases`) at least once.
```

### DIFF
--- a/kubernetes/deployment/in-tree/control-cluster-role.yaml
+++ b/kubernetes/deployment/in-tree/control-cluster-role.yaml
@@ -31,7 +31,6 @@ rules:
    - nodes/status
    - configmaps
    - secrets
-   - endpoints
    - events
    verbs:
    - create
--- a/kubernetes/deployment/in-tree/target-cluster-role.yaml
+++ b/kubernetes/deployment/in-tree/target-cluster-role.yaml
@@ -7,7 +7,6 @@ rules:
    - ""
    resources:
    - nodes
-   - endpoints
    - replicationcontrollers
    - pods
    verbs:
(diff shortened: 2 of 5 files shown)


himanshu-kun (approved):
/lgtm

ialidzhikov:
> We could also check whether the RBAC rules for endpoints can be now removed.

@acumino , any comments on this one?

acumino:
> > We could also check whether the RBAC rules for endpoints can be now removed.
> 
> @acumino , any comments on this one?

There was not any specific RBAC for endpoints for leader election. I am not sure if mcm still need endpoints apart from the use for leader election if not we can drop it else not.

ialidzhikov:
> There was not any specific RBAC for endpoints for leader election. I am not sure if mcm still need endpoints apart from the use for leader election if not we can drop it else not.

That was the whole point of my comment - to check it :)

I recall some cases when client-go or controller-runtime can try to list endpoints for the kubernetes service in the default namespace. At least we should be able to remove create/patch/update verbs for endpoints from the example RBAC.

himanshu-kun:
I have tested by removing the `endpoints` rules from the clusterrole, and tested the following MCM scenarios:
* scaling-up machinedeployment
* scaling-down
* rolling update

There were no logs found stating cannot list `endpoint` or similar, and there were no restarts of MCM.
Hope these are enough @ialidzhikov

ialidzhikov:
Okay, my comment was to adapt the example RBAC and remove the `endpoints` rule. In https://github.com/gardener/machine-controller-manager/pull/662 we adapted the example RBAC to add rules for leases, I guess now we can remove the rules for endpoints.

ialidzhikov (approved):
/lgtm

PR #712, opened 2022-05-05
update local_setup.sh

Mkmittal:
settles a problem with local_setup.sh working with the providers where it allowed only a single space before and after `:=`. But now, it'll allow for any number of spaces b/w `variable_name` and `:=` in the provider's makefile and mcm's makefile.
Now provider is to be passed as a separate entity from the seed name, which was earlier passed as one. This makes the local setup ready to work on other providers, like azure and vsphere, where the seed name can be az or anything, but the provider name is azure.
Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,9 @@ download-kubeconfigs:
 	read SEED; \
 	echo "enter shoot name"; \
 	read SHOOT; \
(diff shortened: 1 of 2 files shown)


gardener-robot-ci-3:

himanshu-kun (approved):
/lgtm

PR #725, opened 2022-05-30
Added changes from PR #567

Mkmittal:
adds changes from PR #567 and makes appropriate changes to work with the latest versions and updates
Fixes #

In the test case `rolling-update case: both new and old machineSet should have appropriate desired replicas` in deployment_test.go the new machine set gets initialised with replicas =  maxSurge + 1, instead of maxSurge and there is no reduction in number of replicas of the older machine set. This requires debugging as it should not be case and because of this the unit test fails.

**Release note**:

```other operator
NONE
```


gardener-robot-ci-3:

himanshu-kun (approved):
/lgtm
Thanks for the changes @Mkmittal !

PR #728, opened 2022-06-13
Upgrade K8s Version to 0.23.7

Mkmittal:
Upgrades K8s version to 0.23.7.
moves from policy v1beta1 to policy v1 as the beta1 version was deprecated and would be gone in K8s v1.25.
Fixes #

--v flag is not working, the error message is an unknown flag, that needs to be debugged.
need to add the discovery check for the policyv1beta1 to policyv1 change as it is only supported from Kubernetes v1.21.

**Release note**:

```other operator

```


gardener-robot-ci-2:

PR #733, opened 2022-07-14
Mark machine failed during rollout also

himanshu-kun:
A bug where MCM was not marking machine as failed during rollout has been resolved.
In situation where the all the machines in the new or the old machineset go into `Unknown` state , as per the current logic the rollout would stop and won't proceed until we manually delete the machine.
This happens because machineSet controller is not aware of a `Unknown` state and so takes no action, and only machine controller could resolve this situation by marking the machine `Failed`.

Fixes #

- this erroneous behavior was introduced recently as a part of the meltdown fix [PR](https://github.com/gardener/machine-controller-manager/pull/683) , to avoid making meltdown logic clash with the rollout logic as they are performed by two seperate controllers and also so that we don't surpass the `maxReplacement` value, but after looking back , machine were always marked `Failed` during rollout previously also , so this PR just kind of goes back to the previous logic
**Release note**:

```bugfix user
Rollout freeze won't happen due to `Unknown` machines now.
```

### DIFF
--- a/pkg/controller/controller_suite_test.go
+++ b/pkg/controller/controller_suite_test.go
@@ -18,7 +18,9 @@ package controller
 
 import (
 	"context"
+	"flag"
 	"fmt"
+	"io"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


kon-angelo:
Do I get it correctly: 
+ the `machineSet` controller can't handle machines in `Unknown` state.
+ the `machine` controller does not mark `Unknown` machines as failed during rollout. 
+ This creates a deadlock when all machines are marked as `Unknown` during rollout.

himanshu-kun:
Yes that is correct

kon-angelo (approved):
/lgtm

PR #738, opened 2022-08-05
Update version of clusterroles, add resoureName for leases

### LINKED ISSUE #674: Update clusterRole for local deployment of MCM and MC

/area usability
/kind enhancement
/priority 3

resourceName `machine-controller` needs to be added over [here](https://github.com/gardener/machine-controller-manager/blob/1417b92329616a3f92aeb40aadbb2b8c6655e393/kubernetes/deployment/out-of-tree/control-cluster-role.yaml#L60-L61)

**Why is this needed**:
When we deploy MCM pod using files [here](https://github.com/gardener/machine-controller-manager/blob/1417b92329616a3f92aeb40aadbb2b8c6655e393/kubernetes/deployment/out-of-tree)(for roles and rolebindings) and [here](https://github.com/gardener/machine-controller-manager-provider-gcp/blob/master/kubernetes/deployment.yaml)(assuming we are deploying OOT GCP) like in local integration tests, and if we use version of OOT providers which vendors MCM v0.43.0 , then there is a possiblity that MC is not able to become the leader as we have moved to endpointleases with [this](https://github.com/gardener/machine-controller-manager/pull/662) PR which is released with MCM v0.43.0

rishabh-11:
This PR updates the rule for leases in cluster role and moves clusterroles/clusterrolebindings to v1 from v1beta1. This is needed for the mcm-provider IT to work when running the controllers inside a cluster and not as a local process. 

Fixes #674 

**Release note**:

```bugfix operator
resourceName `machine-controller` added for leases in clusterrole. Updated version of Clusterroles and Clusterrolebindings to v1.
```

### DIFF
--- a/kubernetes/deployment/in-tree/control-cluster-role-binding.yaml
+++ b/kubernetes/deployment/in-tree/control-cluster-role-binding.yaml
@@ -1,4 +1,4 @@
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
   name: machine-controller-manager-control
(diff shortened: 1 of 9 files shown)


himanshu-kun (approved):
/lgtm

PR #744, opened 2022-09-01
Handle `PodDisruptionBudget` `policy/v1`

shafeeqes:
The policy/v1beta1 API version of PodDisruptionBudget will no longer be served in v1.25.
We should migrate manifests and API clients to use the policy/v1 API version, available since v1.21

Part of https://github.com/gardener/gardener/issues/6567

**Release note**:

```other operator
Migrated clients to use `policy/v1` `PodDisruptionBudget` for kubernetes versions >= 1.21. `policy/v1beta1` PDB is also supported but for k8s < 1.21
```


himanshu-kun (approved):
/lgtm
Thnx for the changes @shafeeqes

shafeeqes:
Thanks for the review, Can we have a release with this PR included?

himanshu-kun:
I am planning a release by EoD

PR #746, opened 2022-09-21
Add nodeName and providerID column while listing machines(s)

### LINKED ISSUE #737: Print node and instance details when printing machines on console.

/area usability 
/kind enhancement
/priority 3

As an operator I would like to see more details about the machine. 
In concrete 2 fields will be really helpful namely - 
* Node Name captured under `node`
* InstanceDetails captured under `providerId` (to map them on the Infra)

To not bloat up the width we can also consider providing these additional columns via   `-o wide` option. This can keep the design consistent with the current usage allowing still an ease to get the details when required using the `-o wide` option. 

**Why is this needed**:
Currently when we fetch machine details using kubectl we get the following output - 
```
NAME                                                 STATUS    AGE
shoot--dev--ashlclkyv-ash-local-kyv-z1-66666-pmb9t   Running   12d
```
To map the machine to the corresponding k8s node object as well the infra instance details one needs to search it through the yaml which can be made easier by having them default printed as a response to fetch machines.

rishabh-11:
WIth this PR, when one runs the` kubectl get machines` command,`Node name` will be displayed as a column. If the `-owide` option is used then `ProviderID` will also be displayed as a column. The output will be as follows:
Screenshot 2022-09-21 at 1 58 33 PM

Screenshot 2022-09-29 at 3 26 25 PM

Fixes #737 

**Release note**:

```feature operator
Using `kubectl get machines` will display `Node` of the corresponding machine as a column. If `-owide` flag is used then the corresponding `ProviderID` will also be displayed.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,7 +149,7 @@ test-clean:
 	@rm -f $(COVERPROFILE)
 
 generate: controller-gen
-	$(CONTROLLER_GEN) crd paths=./pkg/apis/machine/v1alpha1... output:crd:dir=kubernetes/crds output:stdout
+	$(CONTROLLER_GEN) crd paths=./pkg/apis/machine/v1alpha1/... output:crd:dir=kubernetes/crds output:stdout
 	@./hack/generate-code
 	@./hack/api-reference/generate-spec-doc.sh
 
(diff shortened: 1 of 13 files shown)


himanshu-kun (approved):
/lgtm
Thanks for the changes!

PR #759, opened 2022-11-08
Print errors properly

### LINKED ISSUE #756: Ensure all errors for eviction or any other k8s client api call are logged at the default log level `v3`

/area control-plane
/kind bug
/priority 1

**What happened**:
We came across an issue where eviction of pods by MCM were not successful when there were more than one PDB's acting for a given Pod. However the issue was that MCM was not logging anything to identify why is the eviction failing. 

So the PDB would work if the evictions are done by kube-controller-manager. But since this was a rolling update of Gardener shoot we had a seperate "custom" component `MCM` evicting Pods via the API-Server. This eviction will not work if there are two PDB's existing. [See here in the docs](https://kubernetes.io/docs/tasks/run-application/configure-pdb/#arbitrary-controllers-and-selectors) 

On analysing we found  that MCM does logs all errors but they are [logged iat V4](https://github.com/gardener/machine-controller-manager/blob/d9682e5ccbd260201d7777816b4244842f356d71/pkg/controller/drain.go#L663-L664) and the default verbosity is v3.

We expect MCM to log all relevant errors from kapi calls at the default log level 3. 

Although the issue talks about one such occurrence but we can use this to also cover all other code paths where we see similar patterns of error being logged at v4 level and bring them to the default log level to avoid similar problems for other scenarios which are either not reported or have not yet surfaced in the usage. 

**Environment**:

- Others:

himanshu-kun:

Fixes #756 

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/util/provider/machinecontroller/machine_bootstrap_token.go
+++ b/pkg/util/provider/machinecontroller/machine_bootstrap_token.go
@@ -72,7 +72,7 @@ func (c *controller) addBootstrapTokenToUserData(ctx context.Context, machineNam
 		klog.V(4).Infof("replacing url encoded placeholder %s with %s in user-data!", urlEncodedPlaceholder, url.QueryEscape(token))
 		userDataS = strings.ReplaceAll(userDataS, urlEncodedPlaceholder, url.QueryEscape(token))
 	} else {
-		klog.V(4).Info("no bootstrap token placeholder found in user-data, nothing to replace!")
+		klog.Warningf("no bootstrap token placeholder found in user-data, nothing to replace!")
 	}
 
 	secret.Data["userData"] = []byte(userDataS)
--- a/cmd/machine-controller-manager/app/controllermanager.go
+++ b/cmd/machine-controller-manager/app/controllermanager.go
@@ -299,7 +299,9 @@ func StartControllers(s *options.MCMServer,
 }
 
 // TODO: In general, any controller checking this needs to be dynamic so
-//  users don't have to restart their controller manager if they change the apiserver.
+//
+//	users don't have to restart their controller manager if they change the apiserver.
+//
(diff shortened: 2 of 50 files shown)


pkg/util/provider/machinecontroller/machine_bootstrap_token.go
```diff
@@ -72,7 +72,7 @@ func (c *controller) addBootstrapTokenToUserData(ctx context.Context, machineNam
 		klog.V(4).Infof("replacing url encoded placeholder %s with %s in user-data!", urlEncodedPlaceholder, url.QueryEscape(token))
 		userDataS = strings.ReplaceAll(userDataS, urlEncodedPlaceholder, url.QueryEscape(token))
 	} else {
-		klog.V(4).Info("no bootstrap token placeholder found in user-data, nothing to replace!")
+		klog.Warningf("no bootstrap token placeholder found in user-data, nothing to replace!")
```
ashwani2k, line 75: Do we need to take action on this warning. 
Can we may be also include a small description of impact of this missing bootstrap token.

ashwani2k (commented):
Thanks @himanshu-kun for doing the changes. 
Just a small clarification otherwise looks good to me.

pkg/util/provider/machinecontroller/machine_bootstrap_token.go
himanshu-kun: We will still be creating machine in this case , as user might configure other ways of making bootstrap token available.
I have improved the message though , 8f6021a981cb977964ca97d6dc827f718c0314ff

ashwani2k: Thanks.
/lgtm

rishabh-11 (approved):

Issue #754, opened 2022-10-24
Apply machine deployment changes only to latest machine set, never both (e.g. during update)


vlerenc:
## What happened:
While investigating issues with pending workload during a rolling update that looked as if drain started on nodes without sufficient new capacity, we found out that changes to the machine deployment (e.g. cluster auto-scaler out- or in-scaling) are applied to both machine sets, the old one and the new one. This is at least a bug, but it is also the most likely culprit for the pending workload:
- If the cluster scales out, this bug may lead to one more node that shouldn't be added (to the old machine set). This node will become `Ready` and will be tainted immediately, but it may confuse the progress tracking and may result in a wrong node being drained.
- If the cluster scales in, this bug may lead to one more node that shouldn't be removed (from the old machine set). This node will be drained regardless of the ongoing rolling update and the workload become pending.

### Scale-Out
3 x 3 nodes, 2 zones completed their rolling update, 1 zone hangs, `maxSurge` is 6 / 3 = 2 (`maxUnavailable` is 0):
<img width="822" alt="18-pods" src="https://user-images.githubusercontent.com/4974203/197480295-3c993003-1aa4-4c25-b6e3-fe898829da35.png">

After adding more workload, all machine deployments scale 3 -> 4, but the backing machine sets for the blocked zone are both incremented, leading to not 3 (as per machine deployment requested), but 4 new machines:
<img width="819" alt="24-pods" src="https://user-images.githubusercontent.com/4974203/197480510-5f58fa34-a5cf-49f8-a528-3eeb7faaf78c.png">

### Scale-In
Same situation as above, now manually reducing the machine deployment replica count 3 -> 2:
<img width="705" alt="At-3-Scaling-To-2" src="https://user-images.githubusercontent.com/4974203/197481326-04ad7e82-0c9f-4b0c-9146-8d84ed2c6aed.png">

Outcome after the 3 -> 2 change, the old machine set replica count was reduced, now violating the rolling update contract (2 and no longer the old 3 nodes are `Ready` even though no new machine has joined the cluster yet). Now further reducing the machine deployment replica count 2 -> 1:
<img width="706" alt="At-2-Scaling-To-1" src="https://user-images.githubusercontent.com/4974203/197481746-38c1486e-9c3b-4e84-9a57-b8fc8deb2ece.png">

Outcome after the 2 -> 1 change, now even both machine set replica counts were reduced:
[...]

PR #762, opened 2022-12-13
Update `Progressing` condition on machineDeployment for nil `progressDeadlineSeconds`

himanshu-kun:
Currently when `progressDeadlineSeconds` is nil , then no `Progressing` condition is added to machineDeployment. This PR adds that condition now. 
In gardener, currently we deploy machineDeployments with this field as `nil` , but we would also like to see if the deployment is rollingUpdating or not. Now that this condition would be present, we could achieve that.

Fixes #

- done as a pre-step for solving https://github.com/gardener/machine-controller-manager/issues/754

**Release note**:

```feature user
MachineDeployment would now have `Progressing` condition even when no progress Deadline is specified. This condition would never go to the reason `ProgressDeadlineExceeded` in that case.
```


kon-angelo (approved):
/lgtm

PR #773, opened 2023-01-27
Fix machine creation bootstrap tokens expiration

### LINKED ISSUE #772: Node cannot join the cluster due to expired bootstrap token

/kind bug
/priority 3

**What happened**:
Mostly always my machines take more than 20 minutes to be created in the cloudprovider.
Therefore the workers are configured to with a specific `machineCreationTimeout` of `50m`.
When the nodes come up after approx. 30min the node is unable to join cluster due to a expired bootstrap token.

Bootstrap token for node does not expired before the configured `spec.worker[].machineController.machineCreationTimeout`

1. Create a shoot and configure a worker group with a `machineCreationTimeout` of more than 20m
2. let the machine need more than 20 minutes to be created

**Environment**:

- Kubernetes version (use `kubectl version`): 1.22.17
- Cloud provider or hardware configuration: Equinix

schrodit:

The expiration of bootstrap tokens is hardcoded in the machine controller deployment. (depending on the provider but mostly approx. 20min).
There is an option to overwrite this default per worker group with `machineCreationTimeout`.
This option is considered during the machine creation but not in the bootstrap token.
So this means that if the machine takes more time than 20 minutes to be created, the bootstrap token is already expired and the needed configs cannot be fetched form the cluster.

This PR makes the bootstrap token respect the machines `machineCreationTimeout` option.

Fixes #772

Tested with the equinix provider

**Release note**:

```bugfix user
Fix a bug in the bootstrap token creation that caused node to not be able to join the cluster due to an expired bootstrap token.
```


gardener-robot-ci-3:
Thank you @schrodit for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

pkg/util/provider/machinecontroller/machine_bootstrap_token.go
```diff
@@ -90,11 +91,15 @@ func (c *controller) getBootstrapTokenOrCreateIfNotExist(ctx context.Context, ma
 			if err != nil {
 				return nil, err
 			}
+			expiration := c.safetyOptions.MachineCreationTimeout.Duration
+			if machine.Spec.MachineConfiguration != nil && machine.Spec.MachineConfiguration.MachineCreationTimeout != nil {
+				expiration = machine.Spec.MachineConfiguration.MachineCreationTimeout.Duration
+			}
```
himanshu-kun, line 97: we can use the helper function `getEffectiveCreationTimeout` for this.

himanshu-kun (changes requested):
It is a serious bug , Thanks for the PR Tim!

pkg/util/provider/machinecontroller/machine_bootstrap_token.go
schrodit: ah didn't know that helper function exists.

Updated the code PTAL.

himanshu-kun (approved):
/lgtm

PR #778, opened 2023-02-02
Consider critical-components-not-ready taint when checking machine health

SimonKienzler:

A node taint (`node.gardener.cloud/critical-components-not-ready`) is introduced in https://github.com/gardener/gardener/pull/7406 as part of https://github.com/gardener/gardener/issues/7117. This taint needs to be considered by the MCM when it checks the machine health, as a machine is only supposed to be `Ready` if the introduced taint is not present.

Co-Authored-By: @timebertt

Part of https://github.com/gardener/gardener/issues/7117

The change has been tested locally in combination with with `machine-controller-manager-provider-local` in order to verify correct behavior.

**Release note**:
```feature user
Machine object won't turn from `Pending`  to `Running` state if `node.gardener.cloud/critical-components-not-ready` taint is there on the corresponding node.
```


gardener-robot-ci-3:
Thank you @SimonKienzler for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -645,7 +646,7 @@ func (c *controller) reconcileMachineHealth(ctx context.Context, machine *v1alph
 			}
 			cloneDirty = true
 
-		} else if c.isHealthy(clone) && clone.Status.CurrentStatus.Phase != v1alpha1.MachineRunning {
+		} else if c.isHealthy(clone, node) && clone.Status.CurrentStatus.Phase != v1alpha1.MachineRunning {
```
himanshu-kun, line 649: With the suggested way we won't block the `Unknown` machines(primarily) from turning to `Ready` because of the taint.
```suggestion
		} else if c.isHealthy(clone, node) {
		      if clone.Status.CurrentStatus.Phase == v1alpha1.MachinePending{
		       if !c.criticalComponentsNotReadyTaintPresent(node) {
		      ... // turn to Running
		      } else if clone.Status.CurrentStatus.Phase != v1alpha1.MachineRunning{
		      ...// turn to Running
		      }
		}
```

himanshu-kun (changes requested):
From the gardener issue I understand that this taint is relevant only during the first time the machine becomes `Running`. Current changes in the PR also lets the taint affect the health mechanism of MCM. Health mechanism is only relevant after the machines have become `Running` once. So I have suggested changes , not to affect that.
Please let me know if you think otherwise.

SimonKienzler:
@himanshu-kun thanks for your feedback, I gave it a go - please let me know what you think.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -645,6 +645,23 @@ func (c *controller) reconcileMachineHealth(ctx context.Context, machine *v1alph
 			}
 			cloneDirty = true
 
+		} else if c.isHealthy(clone) && clone.Status.CurrentStatus.Phase == v1alpha1.MachinePending {
+			// when checking if a healthy machine in Pending is ready to be Running,
+			// we need to take into account the CriticalNodeComponentsNotReadyTaint.
+			if !c.criticalComponentsNotReadyTaintPresent(node) {
+				// Machine is ready and has joined/re-joined the cluster
```
himanshu-kun, line 652:
```suggestion
				// Machine is ready and has joined the cluster
```

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -645,6 +645,23 @@ func (c *controller) reconcileMachineHealth(ctx context.Context, machine *v1alph
 			}
 			cloneDirty = true
 
+		} else if c.isHealthy(clone) && clone.Status.CurrentStatus.Phase == v1alpha1.MachinePending {
+			// when checking if a healthy machine in Pending is ready to be Running,
+			// we need to take into account the CriticalNodeComponentsNotReadyTaint.
```
himanshu-kun, line 650:
```suggestion
			// when checking if a healthy machine in `Pending` state is ready to be marked `Running`,
			// we need to take into account the CriticalNodeComponentsNotReadyTaint.
```

himanshu-kun (changes requested):
could you also add 3 test cases in a different `Context` block under `##General Machine Health Reconciliation` in `pkg/util/provider/machinecontroller/machine_util_test.go`
* "Pending machine which is healthy but node without taint shouldn't be marked Running"
* "Pending machine which is healthy and node has taint should be marked Running"
* "Unknown machine which is healthy , and node with the taint should be marked Running"

SimonKienzler:
@himanshu-kun I added the test cases. `##General Machine Health Reconciliation` is a `DescribeTable` test, I'm not aware of a way to subdivide the entries into different `Context`s. Hope it's alright like this.

pkg/util/provider/machinecontroller/machine_util_test.go
```diff
@@ -2309,6 +2309,92 @@ var _ = Describe("machine_util", func() {
...
+					},
+					targetMachineName: machineSet1Deploy1 + "-" + "0",
+				},
+				expect: expect{
+					retryPeriod:   machineutils.ShortRetry,
+					err:           errors.New("machine creation is successful. Machine State has been UPDATED"),
+					expectedPhase: v1alpha1.MachineRunning,
+				},
+			}),
+			Entry("unknown machine is healthy, and node doesn't have `critical-components-not-ready` taint, should be marked Running", &data{
```
himanshu-kun, line 2371: Could you modify it (or add another one) with the situation of `node has critical-components-not-ready taint` and machine becomes Running.
This case would make sure , health mechanism is not affected by the taint.

himanshu-kun (changes requested):
Asked for one test case. otherwise looks good.

SimonKienzler:
@himanshu-kun Unfortunately, I cannot (see https://github.com/orgs/community/discussions/5634). Can we make it work in another way?

himanshu-kun:
Okay, I'll push my branch to my fork and then could you cherry-pick the changes to your PR?

himanshu-kun:
My Branch -> https://github.com/himanshu-kun/machine-controller-manager/tree/feature/wait-node-critical

timebertt:
/assign
I will take over this PR while @SimonKienzler is unavailable.

timebertt:
@himanshu-kun I cherry-picked your commits. I think, this PR should be ready to merge :)

himanshu-kun:
We discussed internally and have a common concern that should the `node.gardener.cloud/critical-components-not-ready` be deciding the `Phase` of a machine at all?
`Running` machine itself means that the node is healthy with all the relevant node conditions true, kubelet runnning etc. Taint is just a constraint from the kubernetes side for scheduler. From what I can think of , keeping the machine `Pending` because of the taint,  has only one benefit that the `EveryNodeReady` condition of shoot would not turn green until critical shoot system components are ready. But then we also have `SystemComponentsRunning` condition for all system components.
@timebertt do you see any other benefits which I might have missed?

timebertt:
I would say this is necessary to prevent rolling updates from tumbling. Otherwise, mcm would continue the rolling update even if the customer workload cannot be scheduled on the node yet because it is not fully ready.

From https://github.com/gardener/gardener/issues/7117:

> make machine-controller-manager aware of the new checks and only continue rolling updates once all node-critical pods got ready in addition to waiting for the Ready condition (already done today)

timebertt:
@himanshu-kun please tell me if there is anything missing to merge this PR :)

himanshu-kun:
No there isn't @timebertt . thanks for the explanation.

himanshu-kun (approved):
/lgtm

PR #792, opened 2023-03-05
Bump `k8s.io/* ` and other dependencies

afritzler:
This PR bumps the Kubernetes related dependencies to v1.26.2. 

**Release note**:

```other developer
Bump `k8s.io/*` dependencies to v1.26.2
```


himanshu-kun (approved):
/lgtm

himanshu-kun:
@AndreasBurger we see that after revendoring these MCM changes in mcm-providers (like mcm-aws) and running the provider , its crashing with the following error
```
unknown flag: --v
Usage of /var/folders/gc/zljv6g5j22v9ympp0dj42w_r0000gn/T/go-build711335284/b001/exe/main:
      --address ip                                              The IP address to serve on (set to 0.0.0.0 for all interfaces) (default 0.0.0.0)
      --bootstrap-token-auth-extra-groups string                Comma-separated list of groups to set bootstrap token's "auth-extra-groups" field to
      --cloud-provider string                                   The provider for cloud services.  Empty string for no provider.
      --concurrent-syncs int32                                  The number of nodes that are allowed to sync concurrently. Larger number = more responsive service management, but more CPU (and network) load (default 50)
      --contention-profiling                                    Enable lock contention profiling, if profiling is enabled
      --control-kubeconfig string                               Filepath to the control cluster's kubeconfig where machine objects would be created. Optionally you could also use 'inClusterConfig' when pod is running inside control kubeconfig. (Default value is same as target-kubeconfig)
      --controller-start-interval duration                      Interval between starting controller managers.
      --kube-api-burst int32                                    Burst to use while talking with kubernetes apiserver (default 30)
      --kube-api-content-type string                            Content type of requests sent to apiserver. (default "application/vnd.kubernetes.protobuf")
      --kube-api-qps float32                                    QPS to use while talking with kubernetes apiserver (default 20)
      --leader-elect                                            Start a leader election client and gain leadership before executing the main loop. Enable this when running replicated components for high availability. (default true)
      --leader-elect-lease-duration duration                    The duration that non-leader candidates will wait after observing a leadership renewal until attempting to acquire leadership of a led but unrenewed leader slot. This is effectively the maximum duration that a leader can be stopped before it is replaced by another candidate. This is only applicable if leader election is enabled. (default 15s)
      --leader-elect-renew-deadline duration                    The interval between attempts by the acting master to renew a leadership slot before it stops leading. This must be less than or equal to the lease duration. This is only applicable if leader election is enabled. (default 10s)
      --leader-elect-resource-lock string                       The type of resource object that is used for locking during leader election. Supported options are 'endpoints', 'configmaps', 'leases', 'endpointsleases' and 'configmapsleases'. (default "leases")
      --leader-elect-retry-period duration                      The duration the clients should wait between attempting acquisition and renewal of a leadership. This is only applicable if leader election is enabled. (default 2s)
      --machine-creation-timeout duration                       Timeout (in durartion) used while joining (during creation) of machine before it is declared as failed. (default 20m0s)
      --machine-drain-timeout duration                          Timeout (in durartion) used while draining of machine before deletion, beyond which MCM forcefully deletes machine. (default 2h0m0s)
      --machine-health-timeout duration                         Timeout (in durartion) used while re-joining (in case of temporary health issues) of machine before it is declared as failed. (default 10m0s)
      --machine-max-evict-retries int32                         Maximum number of times evicts would be attempted on a pod before it is forcibly deleted during draining of a machine. (default 360)
      --machine-pv-detach-timeout duration                      Timeout (in duration) used while waiting for detach of PV while evicting/deleting pods (default 2m0s)
      --machine-pv-reattach-timeout duration                    Timeout (in duration) used while waiting for reattach of PV onto a different node (default 1m30s)
      --machine-safety-apiserver-statuscheck-period duration    Time period (in duration) used to poll for APIServer's health by safety controller (default 1m0s)
      --machine-safety-apiserver-statuscheck-timeout duration   Timeout (in duration) for which the APIServer can be down before declare the machine controller frozen by safety controller (default 30s)
      --machine-safety-orphan-vms-period duration               Time period (in durartion) used to poll for orphan VMs by safety controller. (default 30m0s)
      --min-resync-period duration                              The resync period in reflectors will be random between MinResyncPeriod and 2*MinResyncPeriod (default 12h0m0s)
      --namespace string                                        Name of the namespace in control cluster where controller would look for CRDs and Kubernetes objects (default "default")
      --node-conditions string                                  List of comma-separated/case-sensitive node-conditions which when set to True will change machine to a failed state after MachineHealthTimeout duration. It may further be replaced with a new machine if the machine is backed by a machine-set object. (default "KernelDeadlock,ReadonlyFilesystem,DiskPressure,NetworkUnavailable")
      --port int32                                              The port that the controller-manager's http service runs on (default 10259)
      --profiling                                               Enable profiling via web interface host:port/debug/pprof/ (default true)
      --target-kubeconfig string                                Filepath to the target cluster's kubeconfig where node objects are expected to join
unknown flag: --v
exit status 2
make[1]: *** [start] Error 1
```

Are you aware of the reason / remedies ?

needs/second-opinion @elankath

afritzler:
As most of the provider still depend on v0.45 I would check if this problem occurred in one of the previous versions as I didn't change anything in the flagset initialization.

rishabh-11:
Hi @afritzler, Can you add the following line over [here](https://github.com/afritzler/machine-controller-manager/blob/d4df935993dc9cb647646a4794c4325e2317da63/cmd/machine-controller-manager/app/options/options.go#L98)
```go
logs.AddFlags(fs) // Here `logs` is `k8s.io/component-base/logs`.
```
With this change, I am able to run the integration tests for `mcm-provider-aws`.

afritzler:
@rishabh-11 done.

PR #794, opened 2023-03-05
Fix linting issues in `controller` package

afritzler:

- Fix linting issues in controller package
- Fix unhandled errors
- Remove unused code

I noticed that a lot of `errors` are not properly handled in the `controller` package. For now I just logged out the errors when they occur.  As the general error handling flow is not clear to me this should be definitely looked into. Ideally you want to wrap and return an `error` where it occurs.   

**Release note**:

```other developer
NONE
```


himanshu-kun (approved):
/lgtm

PR #795, opened 2023-03-05
Fix linting issues in `test`, `handlers` and `features` packages

afritzler:
Fix linting issues in `test`, `handlers` and `features` packages.

**Release note**:

```other developer
NONE
```


afritzler:
What is the reasoning behind forbidding dot imports?
```
./pkg/test/integration/common/framework.go:18:2: should not use dot imports
./pkg/test/integration/common/framework.go:19:2: should not use dot imports
./pkg/test/integration/common/framework.go:20:2: should not use dot imports
```
In the context of `gomega` and `ginkgo` dot imports IMO are pretty handy.

unmarshall:
>In the context of gomega and ginkgo dot imports IMO are pretty handy.

I agree, it should not be removed in tests. It should however not be used in non-test code.

himanshu-kun:
> should not use dot imports

For now I don't find any way to add it as an exception to `golint` . We will switch to golangci-lint where we could configure for this particular error. 
For the time being lets avoid dot imports @afritzler . Kindly update the PR , I'll merge it quickly.

pkg/test/integration/common/framework.go
```diff
@@ -1242,10 +1250,10 @@ func (c *IntegrationTestFramework) Cleanup() {
 				Machines(controlClusterNamespace).
 				Watch(ctx, metav1.ListOptions{TimeoutSeconds: &timeout}) //ResourceVersion: machineObj.ResourceVersion
 			for event := range watchMachines.ResultChan() {
-				c.ControlCluster.McmClient.
+				Expect(c.ControlCluster.McmClient.
```
himanshu-kun, line 1253: can you ignore `NotFound` errors here and similar other cases . 
Example https://github.com/gardener/gardener/blob/d36d8944d19040f4f6ddd2043dbe93c401291a13/test/integration/gardenlet/bastion/bastion_test.go#L265

himanshu-kun (changes requested):
kindly also remove the dot imports for now, as we don't have a way to ignore those in `golint`

afritzler:
Regarding the dot-imports. I can revert this part. Didn't mean to break things on your side.

unmarshall:
>kindly also remove the dot imports for now, as we don't have a way to ignore those in golint 

Just to add to this point. We are not against usage of dot-imports in tests as it is widely used across several gardener projects and is also listed as an exception to the general rule for using dot-imports.
See https://go.dev/doc/go1compat 
```
Dot imports. If a program imports a standard package using import . "path", additional names defined in the imported package in future releases may conflict with other names defined in the program. We do not recommend the use of import . outside of tests, and using it may cause a program to fail to compile in future releases.
```
https://github.com/golang/lint is deprecated and once we move to `golangci-lint` then we can explicitly add a rule to allow dot-imports in tests. Till then just have a shorter alias for ginkgo and gomega.

himanshu-kun:
> Regarding the dot-imports. I can revert this part. Didn't mean to break things on your side.
@afritzler waiting for your changes. We are planning a release , and would like to get this change in.

afritzler:
Give me a sec here.

pkg/test/integration/common/framework.go
afritzler: Using `.To(Or(Succeed(), BeNotFoundError()))` the same way here will require a `gardener/gardener` import which is a huge change. Let me refactor that in a way that we can live here without for now.

afritzler:
@himanshu-kun I am not sure I like this change as it now includes the `gardener/gardener` dependency which results in a huge PR. change.

himanshu-kun:
> @himanshu-kun I am not sure I like this change as it now includes the `gardener/gardener` dependency which results in a huge PR. change.

Yes definitely, we shouldn't add g/g as dependency. My proposal was more towards copying the relevant code from g/g in some util file, so that we could use the matcher

afritzler:
Alright, I copied the matches over into `test/utils/matchers`.

himanshu-kun (approved):
thanks for the changes!
/lgtm

PR #803, opened 2023-03-29
Edge cases corrected where all machineSets were scaling down to zero

### LINKED ISSUE #802: Edge Case:- Rolling Update scales all machine sets to `0`

/area robustness
/kind bug
/priority critical

**What happened**:
During the live update, we observed that both new and old machine sets were scaled down to `0`.  This happened because during the rolling update, the number of machines for the new machine set reached the `allowedLimit` and so it cannot add more machines to it [code link](https://github.com/gardener/machine-controller-manager/blob/dbafe5d3899fa5f79d10a70cb84d6bcdca7e7b12/pkg/controller/deployment_sync.go#L444).  If we look at the code [here](https://github.com/gardener/machine-controller-manager/blob/dbafe5d3899fa5f79d10a70cb84d6bcdca7e7b12/pkg/controller/deployment_sync.go#L465-L484), the `nameToSize` map is not populated for the corresponding machineSet, and [here](https://github.com/gardener/machine-controller-manager/blob/dbafe5d3899fa5f79d10a70cb84d6bcdca7e7b12/pkg/controller/deployment_sync.go#L479), we call to scale the machine set and pass `0` as a scale value.

The PR which caused this regression - https://github.com/gardener/machine-controller-manager/pull/765

**Environment**:

- Others:

himanshu-kun:

Fixes #802 

**Release note**:

```bugfix user
An edge case where all the machineSets were scaled down to zero has been dealt with.
```

### DIFF
--- a/pkg/controller/controller_suite_test.go
+++ b/pkg/controller/controller_suite_test.go
@@ -29,6 +29,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/watch"
 	coreinformers "k8s.io/client-go/informers"
 	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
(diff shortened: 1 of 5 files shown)


rishabh-11 (approved):
/lgtm

PR #807, opened 2023-04-17
Automated cherry pick of #806: Turn `CrashloopBackoff` machines to `Running` quicker

### LINKED ISSUE #805: [Regression] CrashloopBackoff don't turn to Running quickly

/area quality
/kind bug
/priority 3

**What happened**:

Due to the PR https://github.com/gardener/machine-controller-manager/pull/745 , `CrashloopBackoff` machines don't turn to Running as soon as the node object becomes `Ready`. 

**Details** :

When a machine creation fails , the machine turn to `Crashloopbackoff`. This leads to retrying of reconciliation and `triggerCreationFlow` is called again in every reconciliation until  VM creation succeeds.
In our case, once its succeding, the spec related fields like node label, providerID are updated and the machine is sent for reconciliation again.
In this next reconciliation , it enters `reconcileMachineHealth()` , but due to node object not yet registered , a `LongRetry`(10min) is done.

Also when the node object is created, the event is ignored because of [this check](https://github.com/gardener/machine-controller-manager/blob/657ffc949c7018d73c972575481937215127d8ab/pkg/util/provider/machinecontroller/machine.go#L228) in place.

This leads to machine being in `Crashloopbackoff` phase, and it turns to `Running` only after 10min (given the node object has successfully joined and Ready)

**Environment**:

- Others:

rishabh-11:
This PR turns `CLBF` machines to `Pending` if the `driver.CreateMachine` call was successful in the previous reconciliation. This will help the `reconcileMachineHealth` to turn the `Pending` to the `Running` phase when the machine is reconciled due to node object creation. 

Fixes #805 

**Release note**:

```bugfix user
`CrashloopBackoff` machines will turn to `Running` quicker .
```

### DIFF
--- a/pkg/apis/machine/types.go
+++ b/pkg/apis/machine/types.go
@@ -220,7 +220,7 @@ const (
 	// MachineFailed means operation failed leading to machine status failure
 	MachineFailed MachinePhase = "Failed"
 
-	// MachineCrashLoopBackOff means creation or deletion of the machine is failing. It means that machine object is present but there is no corresponding node object.
+	// MachineCrashLoopBackOff means creation or deletion of the machine is failing. It means that machine object is present but there is no corresponding VM.
 	MachineCrashLoopBackOff MachinePhase = "CrashLoopBackOff"
 )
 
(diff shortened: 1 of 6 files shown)


himanshu-kun (approved):
/lgtm

PR #821, opened 2023-06-05
Keep `DesiredReplicas` annotation up to date even if no scaling happens

### LINKED ISSUE #815: [Edge case] Stale `deployment.kubernetes.io/desired-replicas` annotation block rollout

/area robustness
/kind bug
/priority 3

**What happened**:

A case was seen where there was a single machineSet under the machineDeployment with `deployment.kubernetes.io/desired-replicas` not up-to-date with `machinedeloyment.spec.replicas` , but `machineSet.spec.Replicas ==machineDeployment.spec.replicas`

This led to scaling being detected for this machineDeployment in every reconcile but [`scale`](https://github.com/gardener/machine-controller-manager/blob/7b6f3a000d6e2f3f73b798437ba10b899dedcdc7/pkg/controller/deployment_sync.go#L406-L408) not actually happening.

Later when the consumer triggered a rollout (machineDeployment started referring new machineClass) , the rollout never starts as [rollout logic](https://github.com/gardener/machine-controller-manager/blob/7b6f3a000d6e2f3f73b798437ba10b899dedcdc7/pkg/controller/deployment.go#L541) is called after [scaling](https://github.com/gardener/machine-controller-manager/blob/7b6f3a000d6e2f3f73b798437ba10b899dedcdc7/pkg/controller/deployment.go#L532-L535C1) logic 

- `deployment.kubernetes.io/desired-replicas` to re-sync 
- rollout not to get blocked

I am not sure how this situation was reached. Ideally , on any machineSet update , all the machineSets are updated, which should update `deployment.kubernetes.io/desired-replicas` also.

But these set of events could lead to this situation:
image

**Environment**:
mcm 0.49.0

- Others:

rishabh-11:

Fixes #815 

**Release note**:

```bugfix user
An edge case where outdated DesiredReplicas annotation blocked a rolling update is fixed.
```

### DIFF
--- a/pkg/controller/deployment_sync_test.go
+++ b/pkg/controller/deployment_sync_test.go
@@ -841,7 +841,7 @@ var _ = Describe("deployment_sync", func() {
 					err: false,
 				},
 			}),
-			Entry("if no scaling happened but DesiredReplicas Annotation for the only active machineSet is outdated, it should be updated with the correct value", &data{
+			Entry("if no scaling needed but DesiredReplicas Annotation for the only active machineSet is outdated, it should be updated with the correct value", &data{
 				setup: setup{
 					machineDeployment: newMachineDeployment(mDeploymentSpecTemplate, 1, 500, 2, 0, nil, nil, nil, nil),
 					oldISs: []*machinev1.MachineSet{
--- a/pkg/controller/deployment_sync_test.go
+++ b/pkg/controller/deployment_sync_test.go
@@ -455,8 +455,9 @@ var _ = Describe("deployment_sync", func() {
 					controlObjects = append(controlObjects, data.setup.oldISs[i])
 				}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/deployment_sync_test.go
```diff
@@ -840,6 +841,21 @@ var _ = Describe("deployment_sync", func() {
 					err: false,
 				},
 			}),
+			Entry("if no scaling happened but DesiredReplicas Annotation for the only active machineSet is outdated, it should be updated with the correct value", &data{
```
himanshu-kun, line 844:
```suggestion
			Entry("if no scaling needed but DesiredReplicas Annotation for the only active machineSet is outdated, it should be updated with the correct value", &data{
```

himanshu-kun (approved):
/lgtm

PR #830, opened 2023-06-14
Updated the docforge manifest & added index files

n-boshnakov:
This PR updates the file structure in the Machine controller Manager section of the public Gardener website, implementing the changes asked for in #829. It is connected to a [PR in gardener/documentation](https://github.com/gardener/documentation/pull/417) that modifies the manifest file there.

Fixes #

**Release note**:

```doc developer
NONE
```

### DIFF
--- a/.docforge/manifest.yaml
+++ b/.docforge/manifest.yaml
@@ -1,57 +1,9 @@
+nodesSelector:
+  path: /docs
 structure:
 - name: _index.md
-  source: /README.md
-- name: docs
   properties:
-      frontmatter:
-        title: Documentation
-  nodesSelector:
-    path: /docs
-  nodes:
-  - name: deployment
-    properties:
-      frontmatter:
-        title: Deployment
-        weight: 1
-        categories:
-          - Developers
-          - Operators
-    nodesSelector:
-      path: /docs/deployment
-  - name: development
-    properties:
-      frontmatter:
-        title: Development
-        weight: 2
-        categories:
-          - Developers
-    nodesSelector:
-      path: /docs/development
-  - name: documents
-    properties:
-      frontmatter:
-        title: Documents
-        weight: 3
-        categories:
-          - Developers
-    nodesSelector:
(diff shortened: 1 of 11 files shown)


ashwani2k (commented):
Thanks @n-boshnakov  for the PR. 
I've a concern as the original idea for raising #829 was that we don't need to maintain the hierarchy but it gets implicitly picked as I realized that if we add more docs then we need to also maintain docforge for them. 
This is a clear overhead. 

As mentioned in #829, I saw that other repos like [extension-gcp docforge](https://github.com/gardener/gardener-extension-provider-gcp/blob/master/.docforge/manifest.yaml) which renders everything under the docs (https://github.com/gardener/gardener-extension-provider-gcp/blob/master/.docforge/manifest.yaml) which renders everything under the docs it looks different.

This still leaves the overhead of managing the page. 
So the problem of FAQ not included in the docs will continue till we explicitly create the below mentioned structure for the FAQ doc as well. 

Is that the preferred way now?

n-boshnakov:
> Thanks @n-boshnakov for the PR. I've a concern as the original idea for raising #829 was that we don't need to maintain the hierarchy but it gets implicitly picked as I realized that if we add more docs then we need to also maintain docforge for them. This is a clear overhead.
> 
> As mentioned in #829, I saw that other repos like [extension-gcp docforge](https://github.com/gardener/gardener-extension-provider-gcp/blob/master/.docforge/manifest.yaml) which renders everything under the docs. Can't we do the same here. So the problem of FAQ not included in the docs will continue till we explicitly create the below mentioned structure for the FAQ doc as well.
> 
> Is that the preferred way now?

No, there is a nodeSelector that automatically includes all the content inside the /docs folder into the website. Since the FAQ file is in the /docs folder, it will render automatically, without us needing to add it manually to the manifest. The README file has to be added manually to the manifest, since it is located outside of the /docs folder.

ashwani2k (approved):

PR #831, opened 2023-06-19
Nit: Fix the format verb

ialidzhikov:
```
{"log":"(SetReplicasAnnotations) ms.Name: shoot--foo--bar-default-z3-797b7 desired: 0 , max: 0 , updated: %!d(bool=true)","pid":"1","severity":"INFO","source":"deployment_util.go:615"}
```

The `updated` var is boolean, hence `%t` should be used instead of `%d`. Ref https://pkg.go.dev/fmt

N/A

N/A

**Release note**:

```other operator
NONE
```


unmarshall (approved):
/lgtm

PR #833, opened 2023-06-21
Include `UnavailableReplicas` in MCD status update check

### LINKED ISSUE #832: `UnavailableReplicas` not checked for machine deployment status update

/area robustness
/kind bug
/priority 2

**What happened**:
The current check for whether the status update for machine deployment is needed does not consider `UnavailableReplicas` (See [this](https://github.com/gardener/machine-controller-manager/blob/65b5699827ffc081c088e0ccc91e9dee2f0a13d4/pkg/controller/deployment_util.go#L1187)). This can lead to an incorrect value of `UnavailableReplicas` in the status that will never be updated unless a scaling event happens or the machine deployment is paused or deleted. 

`UnavailableReplicas` in the machine deployment status should be updated correctly. 

Landscape issue no. 108 keeps a track of such occurrences.

rishabh-11:
Includes `UnavailableReplicas` in determining if the status of machine deployment needs to be updated.

Fixes #832 

**Release note**:

```bugfix operator
Included `UnavailableReplicas` in determining if a machine deployment status update is needed
```

### DIFF
--- a/pkg/controller/deployment_util.go
+++ b/pkg/controller/deployment_util.go
@@ -1192,6 +1192,7 @@ func statusUpdateRequired(old v1alpha1.MachineDeploymentStatus, new v1alpha1.Mac
 		old.ReadyReplicas == new.ReadyReplicas &&
 		old.Replicas == new.Replicas &&
 		old.UpdatedReplicas == new.UpdatedReplicas &&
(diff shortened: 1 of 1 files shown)


elankath (approved):
Thanks for investigation and fix.

PR #836, opened 2023-07-01
Introduce persona-based docs

vlerenc:
We would like to introduce persona-based docs focussing on the already known release notes personas:
- users
- operators
- developers

@n-boshnakov Can you please share your thoughts?
@himanshu-kun It would be great, if you could (later, after this PR/eventually, when you have some time) split the FAQ into 3 for `users`, `operators`, and `developers`, because there is good content there (and MCM has no other end-users docs at the moment).
@n-boshnakov @himanshu-kun FYI: For consistency reasons, I removed the local `_index.md` files in favor of the manifest file (most other repos only use the manifest file, which is good (everything is central, weights can be better checked, consistency issues pop up easier) and bad (decentral/modular puts the metadata where the data is), but it seems people have settled on the manifest approach mostly).

**Release note**:
None as ~30 PRs have/will be opened and the message would be the same everywhere.


.docforge/manifest.yaml
```diff
@@ -5,5 +5,60 @@ structure:
...
+        - Developers
+  nodesSelector:
+    path: /docs/development
+- name: documents
+  properties:
+    frontmatter:
+      title: Documents
+      weight: 3
+      categories:
+        - Developers
```
himanshu-kun, line 26: Just a question about how the categories are useful ? Are they used by `docforge` internally somehow, because I don't see any effect of these categories (which were also available in `_index.md`) in the `gardener.cloud` website

himanshu-kun (approved):
/lgtm

> @himanshu-kun It would be great, if you could (later, after this PR/eventually, when you have some time) split the FAQ into 3 for users, operators, and developers, because there is good content there (and MCM has no other end-users docs at the moment).

Sure I'll do that.

himanshu-kun:
waiting for review by @n-boshnakov before merging

.docforge/manifest.yaml
vlerenc: @himanshu-kun  See initial invitation/info e-mail with screenshot. Yes, the categories shall be made available on the documentation pages:
- Externally: Visitors can switch the persona to see different docs, for developers, operators, and/or users
- Internally: Only the user persona tagged docs will be shown, because developers and operators are not in scope of that page

vlerenc:
Thx @himanshu-kun

PR #842, opened 2023-08-24
Introduce new metrics to track API call duration and Update status.Status to capture underlying cause

### LINKED ISSUE #841: Introduce metric to capture duration for Provider API and Driver API invocations, capture underline error in status.Status

/area metering
/kind enhancement
/priority 3

Introduce following metrics:

- Duration for successful provider API calls. Each API call should be metered. This will help fine tune timeouts/deadline.
- Total duration required for a successful execution of `Driver` methods.

**Why is this needed**:
Helps identify and record times it takes to execute provider and driver API calls. Helps developers to look for optimizations and also track any deterioration in performance from providers.

unmarshall:

Two new metrics have been introduced
- Duration of all successful provider API calls
- Total duration taken to execute [Driver](https://github.com/gardener/machine-controller-manager/blob/dbbaa4526084d2d5e5d210721be186a5ee72c68f/pkg/util/provider/driver/driver.go#L28-L39) methods.

In addition to the metrics, `status.Status` has been enhanced to capture the underline `cause error`. A convenience function `WrapError` has also been added which captures the underline cause. This will help consumers(mcm provider implementations) to introspect the error to extract underline error code, especially during writing unit tests

Fixes #841 

**Release note**:

```other operator
New metrics introduced: 
- api_request_duration_seconds -> tracks time taken for successful invocation of provider APIs. This metric can be filtered by provider and service.
- driver_request_duration_seconds -> tracks total time taken to successfully complete driver method invocation. This metric can be filtered by provider and operation.
- driver_requests_failed_total -> records total number of failed driver API requests. This metric can be filtered by provider, operations and error_code.
```

```other developer
status.Status now captures underline cause, allowing consumers to introspect the error returned by the provider. WrapError() function could be used to wrap the provider error
```

### DIFF
--- a/pkg/util/provider/machinecodes/status/status.go
+++ b/pkg/util/provider/machinecodes/status/status.go
@@ -48,6 +48,8 @@ type Status struct {
 	// [google.rpc.Status.details][google.rpc.Status.details] field, or localized
 	// by the client.
 	message string
+	// cause captures the underlying error
+	cause error
 }
 
 // Code returns the status code contained in status.
@@ -63,7 +65,15 @@ func (s *Status) Message() string {
 	return s.message
 }
 
+// Cause returns the underlying error if captured.
+func (s *Status) Cause() error {
+	return s.cause
+}
+
 // Error returns the error message for the status.
+// WARNING: There is an unwritten contract for anyone using status.Status. One MUST never change
+// the message text. It expects error code to be in the first square brackets and error message in the next. Therefore,
+// any change made here should never change that. Any square brackets added after code and error are ignored when parsing.
 func (s *Status) Error() string {
 	return fmt.Sprintf("machine codes error: code = [%s] message = [%s]", s.Code(), s.Message())
 }
@@ -78,6 +88,15 @@ func Error(c codes.Code, msg string) error {
 	return New(c, msg)
 }
 
+// WrapError creates an instance of status.Status wrapping the underlying cause along with the code and custom error message.
+func WrapError(c codes.Code, msg string, cause error) *Status {
+	return &Status{
+		code:    int32(c),
+		message: msg,
+		cause:   cause,
+	}
+}
+
 // FromError returns a Status representing err if it was produced from this
 // package or has a method `GRPCStatus() *Status`. Otherwise, ok is false and a
 // Status is returned with codes.Unknown and the original error message.
@@ -88,10 +107,17 @@ func FromError(err error) (s *Status, ok bool) {
 
 	if matches, errInFind := findInString(err.Error()); errInFind == nil {
 		code := codes.StringToCode(matches[0])
-		return New(code, matches[1]), true
+		return &Status{
+			code:    int32(code),
+			message: matches[1],
+		}, true
 	}
 
-	return New(codes.Unknown, err.Error()), false
+	return &Status{
+		code:    int32(codes.Unknown),
+		message: err.Error(),
+		cause:   err,
+	}, false
 }
 
 // findInString need to check if this logic can be optimized
--- a/pkg/util/provider/metrics/metrics.go
+++ b/pkg/util/provider/metrics/metrics.go
@@ -81,6 +81,8 @@ var (
 	}, []string{"provider", "service"},
 	)
 
+	// APIRequestDuration records duration of all successful provider API calls.
+	// This metric can be filtered by provider and service.
 	APIRequestDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
 		Namespace: namespace,
 		Subsystem: cloudAPISubsystem,
@@ -88,12 +90,23 @@ var (
 		Help:      "Time(in seconds) it takes for a provider API request to complete",
 	}, []string{"provider", "service"})
 
+	// DriverAPIRequestDuration records duration of all successful driver API calls.
+	// This metric can be filtered by provider and operation.
 	DriverAPIRequestDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
 		Namespace: namespace,
 		Subsystem: cloudAPISubsystem,
 		Name:      "driver_request_duration_seconds",
-		Help:      "Total time (in seconds) taken for a driver request to complete",
+		Help:      "Total time (in seconds) taken for a driver API request to complete",
 	}, []string{"provider", "operation"})
+
+	// DriverFailedAPIRequests records number of failed driver API calls.
+	// This metric can be filtered by provider, operation and error code.
+	DriverFailedAPIRequests = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: namespace,
+		Subsystem: cloudAPISubsystem,
+		Name:      "driver_requests_failed_total",
+		Help:      "Number of failed Driver API requests, partitioned by provider, operation and error code",
+	}, []string{"provider", "operation", "error_code"})
 )
 
 // variables for subsystem: misc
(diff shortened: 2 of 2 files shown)


pkg/util/provider/machinecodes/status/status.go
```diff
@@ -88,10 +107,18 @@ func FromError(err error) (s *Status, ok bool) {
 
 	if matches, errInFind := findInString(err.Error()); errInFind == nil {
 		code := codes.StringToCode(matches[0])
-		return New(code, matches[1]), true
+		return &Status{
+			code:    int32(code),
+			message: matches[1],
+			cause:   err,
+		}, true
```
aaronfern, line 114:
```suggestion
		return WrapError(code, matches[1], err), true
```

Something like this should be equivalent,

pkg/util/provider/machinecodes/status/status.go
```diff
@@ -88,10 +107,18 @@ func FromError(err error) (s *Status, ok bool) {
...
 	}
 
-	return New(codes.Unknown, err.Error()), false
+	return &Status{
+		code:    int32(codes.Unknown),
+		message: err.Error(),
+		cause:   err,
+	}, false
```
aaronfern, line 121:
```suggestion
	return WrapError(codes.Unknown, err.Error(), err), false
```
Same as above

pkg/util/provider/machinecodes/status/status.go
```diff
@@ -48,6 +48,8 @@ type Status struct {
 	// [google.rpc.Status.details][google.rpc.Status.details] field, or localized
 	// by the client.
 	message string
+	// cause captures the underline error
```
aaronfern, line 51:
```suggestion
	// cause captures the underlying error
```
Can you please update all other similar occurrences as well?

pkg/util/provider/machinecodes/status/status.go
unmarshall: It is golang good practices to keep the keys. This will also be highlighted and often auto-corrected by goland IDE as well. Let us keep it this way as it also helps with readability.

pkg/util/provider/machinecodes/status/status.go
unmarshall: Same reason as above. Let us keep it this way.

pkg/util/provider/machinecodes/status/status.go
aaronfern: Fair enough, lets keep it that way then
Thanks!

pkg/util/provider/metrics/metrics.go
```diff
@@ -80,6 +80,20 @@ var (
...
+		Namespace: namespace,
+		Subsystem: cloudAPISubsystem,
+		Name:      "api_request_duration",
+		Help:      "Time it takes for a provider API request to complete",
+	}, []string{"provider", "service"})
+
+	DriverAPIRequestDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: namespace,
+		Subsystem: cloudAPISubsystem,
+		Name:      "driver_request_duration",
```
himanshu-kun, line 94:
```suggestion
		Name:      "driver_request_duration_seconds",
```
Same as above

pkg/util/provider/metrics/metrics.go
```diff
@@ -80,6 +80,20 @@ var (
 		Help:      "Number of Failed Cloud Service API requests, partitioned by provider, and service.",
 	}, []string{"provider", "service"},
 	)
+
+	APIRequestDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: namespace,
+		Subsystem: cloudAPISubsystem,
+		Name:      "api_request_duration",
```
himanshu-kun, line 87:
```suggestion
		Name:      "api_request_duration_seconds",
```
duration metrics usually have the time unit in their name

pkg/util/provider/machinecodes/status/status.go
```diff
@@ -88,10 +107,18 @@ func FromError(err error) (s *Status, ok bool) {
 
 	if matches, errInFind := findInString(err.Error()); errInFind == nil {
 		code := codes.StringToCode(matches[0])
-		return New(code, matches[1]), true
+		return &Status{
+			code:    int32(code),
+			message: matches[1],
+			cause:   err,
```
himanshu-kun, line 113: here the `cause` shouldn't be filled with `err` as its the `status.Status` object only. in case we can't decode the `err` then it should be filled, like you have done when the `if condition` fails.
I think we should keep it nil and explain with a comment abt the reason, WDYT?

pkg/util/provider/metrics/metrics.go
```diff
@@ -80,6 +80,20 @@ var (
...
+		Help:      "Time it takes for a provider API request to complete",
+	}, []string{"provider", "service"})
+
+	DriverAPIRequestDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: namespace,
+		Subsystem: cloudAPISubsystem,
+		Name:      "driver_request_duration",
+		Help:      "Total time taken for a driver request to complete",
```
himanshu-kun, line 95:
```suggestion
		Help:      "Total time (in seconds) taken for a driver request to complete",
```

pkg/util/provider/metrics/metrics.go
```diff
@@ -80,6 +80,20 @@ var (
 		Help:      "Number of Failed Cloud Service API requests, partitioned by provider, and service.",
 	}, []string{"provider", "service"},
 	)
+
+	APIRequestDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: namespace,
+		Subsystem: cloudAPISubsystem,
+		Name:      "api_request_duration",
+		Help:      "Time it takes for a provider API request to complete",
```
himanshu-kun, line 88:
```suggestion
		Help:      "Time(in seconds) it takes for a provider API request to complete",
```

himanshu-kun (changes requested):
thanks for the changes!
minor nits requested.

pkg/util/provider/metrics/metrics.go
```diff
@@ -80,6 +80,20 @@ var (
...
+	}, []string{"provider", "service"})
```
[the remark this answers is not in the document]
unmarshall: `virtual_machine_get`, `virtual_machine_update`, `virtual_machine_delete`, `virtual_machine_create`, `subnet_get`, `nic_get`, `nic_delete`, `nic_create` and similar for all other resource APIs. 
Using `service` label is not accurate but it was already used for other metrics like `APIRequestCount` and `APIFailedRequestCount`. 

I would rather want the following labels:
- provider
- resource_api
- status_code // Right now we only capture duration metrics for successful calls. If we wish to capture all calls then its essential to capture the returned status code. If only successful calls needs to be metered then this can be ommitted.

pkg/util/provider/machinecodes/status/status.go
unmarshall: Ok that is right.

pkg/util/provider/metrics/metrics.go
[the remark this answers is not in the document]
himanshu-kun: since we already send the status code for the driver calls , lets introduce `status_code` for them? would help us track the number of errors due to `ResourceExhausted` also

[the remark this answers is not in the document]
himanshu-kun: I think it would require two additional lables:
- `success` - to tell if the call was successful
- `status_code` - for the returned status code on an error

himanshu-kun (approved):
/lgtm

PR #843, opened 2023-08-28
Minor updates to integration tests doc

aaronfern:
This PR updates the integration tests doc to make it a bit more clear to a first time reader 

Fixes #

**Release note**:

```other operator

```


elankath (approved):
Thanks for the corrections
- everyone in MCM uses `make download-kubeconfigs` so the manual way to download cluster configs was ignored.
- Good catch on the `ginkgo` expect assertion correction.  It doesn't take the error but only first argument.

himanshu-kun (approved):
/lgtm

PR #849, opened 2023-09-12
Reconcile machine before creation deadline

### LINKED ISSUE #790: Update CrashLoopBackOff condition as soon as timeout expires

/area robustness
/kind bug
/priority 2

**What happened**:

Currently MCM doesn't turn `CrashLoopBackoff` machines to `Failed` after the `creationTimeout` expires. A delay as big as 2 min has been observed. This is primarily happening as we don't reconcile the machine on timeout expiry, but reconcile it whenever it comes into the queue due to periodic requeue or CRUD events, which don't happen on a `creationTimeout` expiry.

Turn to `Failed` as soon as timeout expires.

We could take inspiration from machineDeployment logic which turns `Progressing` condition to `False` with reason `ProgressDeadlineExceeded` as soon as the deadline exceeds.

**Environment**:

- Others:

aaronfern:
This PR adds the following 
1. Adapts the retry period for machines in creation so that machines are always reconciled before the creation deadline 
2. Adds a context.WithDeadline to the `CreateMachine()` call so that this call can time out before the creation deadline expires 

Fixes #790 

**Release note**:

```other operator

```


pkg/util/provider/machinecontroller/machine_util_test.go
```diff
@@ -2725,4 +2725,71 @@ var _ = Describe("machine_util", func() {
...
+
+				c, trackers := createController(stop, testNamespace, nil, nil, nil, nil)
+				defer trackers.Stop()
+
+				c.safetyOptions.MachineCreationTimeout = data.setup.mcCreateTimeout
+
+				adjustedRetry := c.adjustCreateRetryRequired(data.setup.machine, machineutils.RetryPeriod(data.setup.createOpRetry))
+
+				if data.expect.retryAdjusted {
+					Expect(adjustedRetry).Should(BeNumerically("<=", machineutils.RetryPeriod(c.getEffectiveCreationTimeout(data.setup.machine).Duration)))
```
himanshu-kun, line 2758:
```suggestion
// Considering time.Now() = 0
					Expect(adjustedRetry).Should(BeNumerically("<=", machineutils.RetryPeriod(c.getEffectiveCreationTimeout(data.setup.machine).Duration)))
```
for ease of understanding :)

himanshu-kun (changes requested):
Considering that at this moment when mcm-provider Create() calls don't respect the passed context, we have solved [point 2)](https://github.com/gardener/machine-controller-manager/issues/790#:~:text=even%20if%20the%201)%20doesn%27t%20exist%20to%20contribute%20to%20the%20delay%2C%20the%20retry/re%2Dpush%20to%20the%20queue%20can%20also%20introduce%20the%20delay%20of%20ShortRetry(3min).%20This%20happens%20because%20the%20retry%20period%20is%20not%20calculated%20considering%20the%20time%20left%20before%20timeout%20but%20is%20just%20a%20constant%20value%20of%20ShortRetry(3min)) of the issue

I think the only neat way to solve point 1) is to enable ctx awareness in Create() calls.

But that aside, I have one concern that with CrashLoopBackOff machines getting marked `Failed` soon could lead to more chances of the race observed in issue https://github.com/gardener/autoscaler/issues/118

Also if we solve the above issue, by considering `Failed` as `Terminating` that would basically mean that most of the times MCM is quicker to mark the machine `Failed` / `Terminating` and create a new replacement machine(have the same CLBF fate)  and CA will never actually be able to remove this kind of machine (though it'll backoff from node-grp)

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -355,9 +355,10 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 				// If node label is not present
 				klog.V(2).Infof("Creating a VM for machine %q, please wait!", machine.Name)
 				klog.V(2).Infof("The machine creation is triggered with timeout of %s", c.getEffectiveCreationTimeout(createMachineRequest.Machine).Duration)
-				createMCCtx, cancelFunc := c.getCreationContext(ctx, machine)
+				createMachineCtx, cancelFunc := c.getCreationContext(ctx, machine)
 				defer cancelFunc()
-				createMachineResponse, err := c.driver.CreateMachine(createMCCtx, createMachineRequest)
+				//TODO Adapt all mcm-providers driver method implementations to use the passed context
```
himanshu-kun, line 360:
```suggestion
// TODO: Adapt all mcm-providers driver method implementations to use the passed context
```

himanshu-kun (changes requested):
just a nit

Issue #590, opened 2021-01-25
Add support for error codes


ialidzhikov:
**What would you like to be added**:
Under the gardener org various CRs and API resources support error codes - `Shoot`, `Etcd`, extension CRs.
It will be very helpful if the Machine status also supports error codes.

Currently a Machine that fails to be created for example because of invalid credentials has the following example status:

```yaml
status:
  lastOperation:
    description: "Cloud provider message - AuthFailure: AWS was not able to validate
      the provided access credentials\n\tstatus code: 401, request id: <omitted>"
    lastUpdateTime: "2021-01-25T17:49:06Z"
    state: Failed
    type: Delete
```

There is nothing wrong with the above status but it is not very helpful when another component needs to interpret the error  and to mark it properly (whether it is `ERR_INFRA_UNAUTHORIZED`, `ERR_INFRA_QUOTA_EXCEEDED` or any other error code).

We could add lastError to the Machine status that can hold the error code related to the last failed operation:

```yaml
status:
  lastError:
    description: "Cloud provider message - AuthFailure: AWS was not able to validate
      the provided access credentials\n\tstatus code: 401, request id: <omitted>"
    code: ERR_INFRA_UNAUTHORIZED
  lastOperation:
    description: "Cloud provider message - AuthFailure: AWS was not able to validate
      the provided access credentials\n\tstatus code: 401, request id: <omitted>"
    lastUpdateTime: "2021-01-25T17:49:06Z"
    state: Failed
    type: Delete
```

Initially this enhancement is requested in https://github.com/gardener/gardener/issues/3020 where the concrete requirement for machine-controller-manager is to properly detect and flag misconfigured PodDisruptionBudgets that require zero voluntary Pod evictions and prevent graceful Node drain.

**Why is this needed**:
- To establish a proper contract between machine-controller-manager and component that heavily rely on Machine status (extension controllers, gardenlet)
- To enrich the Machine status with valuable information

/area ops-productivity

ialidzhikov:
> Hm, how would MCM do it? It cannot do it better compared to how Gardener/the extension do it, or?

[...]

PR #851, opened 2023-09-15
Add `errorCode` field in `LastOperation` struct

rishabh-11:
This PR introduces an error code field in the last operation struct inside the machine status. This error code is only populated for the case of the `CreateMachine` call in the `triggerCreationFlow` for resource exhaustion cases as a part of 
https://github.com/gardener/autoscaler/issues/154

Fixes # part of #590 

**Release note**:

```other operator
Added `errorCode` field in the `LastOperation` struct. This is implemented only for the `CreateMachine` call in the `triggerCreationFlow`.
```

### DIFF
--- a/pkg/util/provider/machinecodes/status/status.go
+++ b/pkg/util/provider/machinecodes/status/status.go
@@ -32,7 +32,6 @@ package status
 
 import (
 	"fmt"
-	"regexp"
 	"strings"
 
 	"github.com/gardener/machine-controller-manager/pkg/util/provider/machinecodes/codes"
@@ -105,7 +104,7 @@ func FromError(err error) (s *Status, ok bool) {
 		return nil, true
 	}
 
-	if matches, errInFind := findInString(err.Error()); errInFind == nil {
+	if matches, errInFind := findCodeAndMessage(err.Error()); errInFind == nil {
 		code := codes.StringToCode(matches[0])
 		return &Status{
 			code:    int32(code),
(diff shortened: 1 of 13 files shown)


pkg/util/provider/machinecodes/status/status.go
```diff
@@ -120,21 +119,34 @@ func FromError(err error) (s *Status, ok bool) {
...
+		switch char {
+		case '[':
+			counter++
+			temp = append(temp, char)
+		case ']':
+			if counter > 0 {
+				counter--
+				temp = append(temp, char)
+				if counter == 0 {
+					decoded = append(decoded, strings.Trim(string(temp), "[]"))
```
himanshu-kun, line 137: I had seen a small edge case , kindly adopt this change, as it solves that.
```suggestion
					tempString := string(temp)
					decoded = append(decoded, tempString[1:len(tempString)-1])
```

pkg/util/provider/machinecontroller/machine_test.go
```diff
@@ -774,6 +779,63 @@ var _ = Describe("machine", func() {
...
+							},
+						},
+					}, nil, nil, nil, nil, true, metav1.Now()),
+				},
+				action: action{
+					machine: "machine-0",
+					fakeDriver: &driver.FakeDriver{
+						VMExists:   false,
+						ProviderID: "fakeID-0",
+						NodeName:   "fakeNode-0",
```
himanshu-kun, line 819: I don't think `NodeName` should be filled as VM can't get created

pkg/util/provider/machinecontroller/machine_safety.go
```diff
@@ -94,6 +94,7 @@ func (c *controller) reconcileClusterMachineSafetyAPIServer(key string) error {
 					}
 					machine.Status.LastOperation = v1alpha1.LastOperation{
 						Description:    "Machine Health Timeout was reset due to APIServer being unreachable",
+						ErrorCode:      "",
```
himanshu-kun, line 97: I think we should remove these `""` ErrorCode initialization from everywhere, will keep the code cleaner

himanshu-kun (changes requested):
a few changes requested, otherwise looks good

himanshu-kun (approved):
/lgtm

PR #868, opened 2023-11-10
Proposal: Post Create Initialization

### LINKED ISSUE #840: Machine Update: Stage 1: Introduce InitializeMachine to Driver interface

/area control-plane
/kind enhancement
/priority 3

Currently [Driver](https://github.com/gardener/machine-controller-manager/blob/master/pkg/util/provider/driver/driver.go#L28-L39) implemented by all providers does not have a way to initialize a machine after creation. We now have two use cases for which `InitializeMachine` is now a requirement:

1. https://github.com/gardener/machine-controller-manager-provider-aws/pull/128 - assigns IPV6 addresses to the instance post creation of the instance.
2. https://github.com/gardener/machine-controller-manager-provider-aws/pull/39 - enable/disable source/destination check on an instance.

`CreateMachine` should be an atomic operation and should only do one thing - create a new machine. [Issue#comment](https://github.com/gardener/machine-controller-manager-provider-aws/pull/128#pullrequestreview-1558207610) highlights this requirement. We have already broken the contract of `CreateMachine` once (see 2 above) and as a consequence also broken the contract of `GetMachineStatus` (see 2 above). 

**Why is this needed**:
There is a need to introduce a new method `InitializeMachine` which only does a one-time update of an existing instance. Changes that are done in (1) and (2) above should be moved to this new method.

### LINKED ISSUE #871: Machine Update: Stage 1: Post Creation Instance Initialization

/area control-plane
/kind enhancement
/priority 3

Enhance the Machine Controller `triggerCreationFlow` to correctly handle post-creation instance initialization steps in `Driver.InitializeMachine` with appropriate retry handling.

**Why is this needed**:

Today the [driver.Driver](https://github.com/gardener/machine-controller-manager/blob/rel-v0.49/pkg/util/provider/driver/driver.go#L28) facade represents the boundary between the the `machine-controller` and its various provider specific implementations.

We have abstract operations for creation/deletion and listing of machines (actually compute instances) but we do not correctly handle post-creation startup logic. Nor do we provide an  abstract operation to represent the hot update of an instance after creation.

We have found this to be necessary for several use cases. Today in the MCM AWS Provider, we already misuse `driver.GetMachineStatus`  which is supposed to be a read-only operation obtaining the status of an instance. 

Each AWS EC2 instance performs source/destination checks by default.
For [EC2 NAT](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_NAT_Instance.html#EIP_Disable_SrcDestCheck) instances
these should be disabled. This is done by issueing
a [ModifyInstanceAttribute](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ModifyInstanceAttribute.html) request with the `SourceDestCheck` set to `false`. The MCM AWS Provider, decodes the  [AWSProviderSpec](https://github.com/gardener/machine-controller-manager-provider-aws/blob/39318bb2b5b4a573fdc77eaf400839d12c4abf59/pkg/aws/apis/aws_provider_spec.go#L63), reads `providerSpec.SrcAndDstChecksEnabled` and correspondingly issues the call to modify the already launched instance.  owever, this should be done as an action after creating the instance and should not be part of the VM status retrieval.

Similarly, there is a [pending PR](https://github.com/gardener/machine-controller-manager-provider-aws/pull/128) to add the `Ipv6AddessCount` and `Ipv6PrefixCount` to enable the assignment of an ipv6 address and an ipv6 prefix to instances. This requires constructing and issueing an [AssignIpv6Addresses](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_AssignIpv6Addresses.html) request after the instance is started.

To mark this issue as completed, following components should be updates:-

- [x] mcm https://github.com/gardener/machine-controller-manager/pull/898
- [x] mcm-provider-aws https://github.com/gardener/machine-controller-manager-provider-aws/pull/162
- [x] mcm-provider-az https://github.com/gardener/machine-controller-manager-provider-azure/pull/144
- [x] mcm-provider-gcp https://github.com/gardener/machine-controller-manager-provider-gcp/pull/118
- [x] mcm-provider-alicloud https://github.com/gardener/machine-controller-manager-provider-alicloud/pull/75
- [x] mcm-provider-openstack https://github.com/gardener/machine-controller-manager-provider-openstack/pull/124

elankath:

See #871

Fixes #871

**Release note**:

```doc developer
NONE
```


docs/proposals/update-machine.md
```diff
@@ -0,0 +1,163 @@
...
+
+
+
+2. **Stage-B**: Introduction of `Driver.UpdateMachine` and enhancing the MCM, MCM providers and gardener extension providers to support hot update of instances through `Driver.UpdateMachine`.  The MCM [triggerUpdationFlow](https://github.com/gardener/machine-controller-manager/blob/v0.50.1/pkg/util/provider/machinecontroller/machine.go#L531) - a reconciliation sub-flow of the MCM which is supposed to be responsible for orchestrating instance update - but currently not used, will be updated to invoke the provider `Driver.UpdateMachine` on hot-updates to to the `Machine` object
```
unmarshall, line 28: This should now be `Driver.HotUpdateMachine`.

docs/proposals/update-machine.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Stage-A Proposal
+
+
+### Instance Not Ready Taint 
+
+- Due to the fact that creation flow for machines will now be enhanced to correctly support post-creation startup logic, we should not scheduled workload until this startup logic is complete.  Even without this feature we have a need for such a taint as described in [MCM#740](https://github.com/gardener/machine-controller-manager/issues/740)
```
unmarshall, line 35: These are not really post creation startups, but post-creation VM updates. I am not convinced that we should be calling this `Startup`

docs/proposals/update-machine.md
```diff
@@ -0,0 +1,163 @@
...
+- We propose a new taint `node.machine.sapcloud.io/instance-not-ready` which  will be added as a node startup taint in gardener core [KubeletConfiguration.RegisterWithTaints](https://github.com/gardener/gardener/blob/v1.83.1/pkg/component/extensions/operatingsystemconfig/original/components/kubelet/config.go#L101)
+- The will will then removed by MCM in health check reconciliation, once the machine becomes fully ready. (when moving to `Running` phase)
```
unmarshall, line 37:
```suggestion
- This taint will then removed by MCM in health check reconciliation, once the machine becomes fully ready. (when moving to `Running` phase)
```

docs/proposals/update-machine.md
```diff
@@ -0,0 +1,163 @@
...
+Phase empty or CrashLoopBackOff ?
+"}--yes-->triggerCreationFlow
+chk--noo-->LongRetry["return machineutils.LongRetry"]
+```
+
+Today, the `triggerCreationFlow` is illustrated below with some minor details omitted/compressed for brevity 
+
+*NOTES*
+* The `lastop` below is an abbreviation for `machine.Status.LastOperation`. This, along with the machine phase is generally updated on the `Machine` object just before returning from the method.
+* `machineCreateErrorHandler` is a separate method on the controller that handles
```
unmarshall, line 65: Sentence is incomplete

docs/proposals/update-machine.md
```diff
@@ -0,0 +1,163 @@
...
+1. Observe that after the call to a successful `Driver.CreateMachine`, the machine phase is set to `Pending`, the `LastOperation.Type` is currently set to `Create` and the `LastOperation.State` set to `Processing`before returning with a `ShortRetry`. The `LastOperation.Description` is (unfortunately) set to the fixed message: `Creating machine on cloud provider`.
+1. Observe that after an erroneous call to `Driver.CreateMachine`, the machine phase is set to `CrashLoopBackOff` or `Failed`  (in case of creation timeout).
+1. We propose introducing a new MC error code `codes.Startup` indicating that the VM Instance was created but there was a startup error in post VM creation activities. The implementor of `Driver.CreateMachine` can return this error code, indicating that `CreateMachine` needs to be called again. However, the Machine Controller will keep the machine phase for such a situation as  `Pending` and not change the phase to `CrashLoopBackOff` when encountering a `codes.Startup` error. (The instance for the machine HAS been created hence `CrashLoopBackoff` is inappropriate)
```
unmarshall, line 124: If we decide to use another error code then do change it here as well. Startup signals that the VM was created and the startup of the VM which involves running the user config script, starting up kubelet and other k8s components failed. But here we are hinting at an update call to the VM post create which does not signal `Startup`.

unmarshall: One can follow simple terminology - PreCreate, Create, PostCreate.

docs/proposals/update-machine.md
```diff
@@ -0,0 +1,163 @@
...
+1. We will introduce a new _machine operation_ stage `InitiateStartup`. In case of an `codes.Startup` error, the `machine.Status.LastOperation.Description` will be set to this stage, the `LastOperation.Type` will be set to `Create` and the `LastOperation.State` set to `Failed` before returning with a `ShortRetry`
```
unmarshall, line 125: I thought we said that if a `driver.CreateMachine` call fails then we simply stick to the current behavior and always move it to either `CrashLoopBackoff` or `Failed` right? `Pending` unfortunately is way too generic and today it means that `CreateMachine` call was successful but in this case it was not.

docs/proposals/update-machine.md
```diff
@@ -0,0 +1,163 @@
...
+
+Today the [driver.Driver](https://github.com/gardener/machine-controller-manager/blob/rel-v0.49/pkg/util/provider/driver/driver.go#L28) facade represents the boundary between the the `machine-controller` and its various provider specific implementations.
+
+We have abstract operations for creation/deletion and listing of machines (actually compute instances) but we do not correctly handle post-creation startup logic. Nor do we provide an  abstract operation to represent the hot update of an instance after creation.
+
+We have found this to be necessary for several use cases. Today in the MCM AWS Provider, we already misuse `driver.GetMachineStatus`  which is supposed to be a read-only operation obtaining the status of an instance. 
+
+Each AWS EC2 instance performs source/destination checks by default.
+For [EC2 NAT](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_NAT_Instance.html#EIP_Disable_SrcDestCheck) instances
+these should be disabled. This is done by issueing
```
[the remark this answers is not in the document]
elankath: typo fixed

docs/proposals/update-machine.md
himanshu-kun: I think `UpdateMachine` sounds better and inline with already existing driver names. We could also update docstring to tell that its for hot-updatable... since we have moved all the startup related update operation to `CreateMachine` anyone reading `UpdateMachine` wouldn't be that confused. WDYT?

docs/proposals/update-machine.md
```diff
@@ -0,0 +1,163 @@
...
+
+We have other uses-cases such as  [MCM Issue#750](https://github.com/gardener/machine-controller-manager/issues/750) where there is a requirement to provide a way for consumers to add tags which can be hot-updated onto instances. This requirement can be generalized to also offer a convenient way to specify tags which can be applied to VMs, NICs, Devices etc.
+
+## Objectives
+
+We will split the fulfilment of this overall need into 2 stages of implementation.
+
+1. **Stage-A**: Support post-VM creation logic in implementations of `Driver.CreateMachine` by permitting provider implementors to add one-time logic after VM creation, return with special new error code `codes.Startup` for post-VM errors and correspondingly support a new machine operation stage `InitiateStartup`. The [triggerCreationFlow](https://github.com/gardener/machine-controller-manager/blob/rel-v0.50/pkg/util/provider/machinecontroller/machine.go#L310) - a reconciliation sub-flow of the MCM responsible for orchestrating instance creation and updating machine status will be changed to support this behaviour.
```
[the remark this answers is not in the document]
elankath: 1. One-time logic means logic is executed only once after creation.
2. will change `codes.Startup` to `codes.PostCreate` 
3. will change `InitiateStartup` to `PostCreate`

himanshu-kun (changes requested):
The proposal looks good, as it tries to make minimum changes to the logic, plus keep the flow logical. I have added a corner case which we should think of handling, if its valid

docs/proposals/update-machine.md
elankath: (convo from slack)

> It was said in the meeting not to do this as `CrashLoopBackoff` is meant for VMs that don’t get created and `Failed` is meant for VMs that don’t get created after machine creation timeout.  For cases where the VM gets created but post-creation (aka startup activities) failed, we still retain the `Pending` phase

From Madhav: There was too much back and forth on this unfortunately and what i recollect is that we should retain the current behavior of moving the VM to CrashLoopBackoff because today it signifies failure in creating the VM which now involves one or more steps. Maybe @rishabh-11 and @himanshu-kun  can confirm.

himanshu-kun, line 125: There are situations which we should think of dealing with. For example, what if following condition happens
- VM was created && 
- there was a startup failure &&,
-  `status.LastOperation` couldn't be updated due to KAPI rate limiting or MCM restart 

then in the next reconcile, we wouldn't know at what step we are

`GetMachineStatus` is used today to see if VM exists ,but it won't be able to help in identifying if startup failure happened or not.
This is happening as the `CreateMachine` call won't be atomic as per the proposal (its not atomic in Azure as well presently, but there we separately list nic to see where we have reached in create operation)

Probable solutions to the above situation:
* always assume StartupFailure happened in such a situation, let `CreateMachine()` be called , but enhance `CreateMachine()` to first check if the VM has all the startup settings done(which I'm not sure is possible for every startup setting), if not then call startup related calls

docs/proposals/update-machine.md
unmarshall: This is currently a point of contention since we have now changed the boundaries for `UpdateMachine`. Since this is part of Stage-B i would keep it `Out-Of-Scope` completely in this document. When we take up hot updates of tags then we can rethink on it.

From POV it is a bit un-clean (maybe i am not convinced now and some thinking from my end is required). Reasons for hesitation:
- Create machine creates an instance and also does `update` on the created instance. This is part of `create machine`.
- Update machine also now does an update but you wish to differentiate it by having a doc string? Why have one name of a method and do something different which is then described as part of docstring?
- When will `Update machine` be used other than hot-updating the VM? If it results in rolling update of the VM then it would most probably be part of `CreateMachine` itself now, right?

docs/proposals/update-machine.md
unmarshall: Ideally `GetMachineStatus` should be enhanced to give the status of the machine keeping in mind the desired state of the machine vs as-is. This means that in cases where a `Create` involves creating the VM + updating properties on the VM (e.g. ipv6 specific configuration or source-destination-checks) then that should be separately sent back as part of the response. 

MCM can then decide what call to make based on the status returned.

rishabh-11: >i recollect is that we should retain the current behavior of moving the VM to CrashLoopBackoff because today it signifies failure in creating the VM.

Yes, this is correct.

> However in terms of MCM our usage is plain and simply wrong. We do not have a crashing VM

I don't think it is wrong. MCM uses `CrashloopBackoff` to say that VM creation requests are failing. if we compare it to a pod, then it is similar to pod restarts due to failure. For a VM creation, we only have success or failure, no restart happens in creation, so multiple restarts can be compared to the CreateMachine calls that MCM makes in each reconciliation until it succeeds.

elankath: Agree it the clearly the job of `GetMachineStatus` to report this correctly. Provider implementations can simply return the same error proposed error code when the VM exists but startup not fully done.  `codes.Startup(Error)` (Himanshu preference) or `codes.PostCreate(Error)`. (Madhav preference)
 
 Please note that suffixing `Error` to the error codes currently breaks the convention in `codes`. There is no `Error` suffix in any of the error codes and I would like to preserve existing convention.

rishabh-11: >Maybe @rishabh-11 and @himanshu-kun can confirm.

We decided to use a separate stage for the error in startup updates. Now if we want to follow the definition of CLBF, then it can be argued that startup activities should be included in it. Then in that case, we need to have proper logs that can help us differentiate why the machine is in CLBF and this should be reflected in the status of the Machine object as well.

unmarshall: >I don't think it is wrong. MCM uses CrashloopBackoff to say that VM creation requests are failing. if we compare it to a pod, then it is similar to pod restarts due to failure. For a VM creation, we only have success or failure, no restart happens in creation, so multiple restarts can be compared to the CreateMachine calls that MCM makes in each reconciliation until it succeeds.

`CrashLoopBackoff` has a word crash in it for a very simple reason that the failure to start a container within a pod is that it is crashing. This means that the container gets created but it then crashes. This is not the same as failing to create a VM because the VM does not get created and then crashes.

rishabh-11: Then we should not use CLBF at all and replace it with something else

unmarshall: >Then we should not use CLBF at all and replace it with something else

This is why i started with `It is a pity that we use Crashlookbackoff` in my original comment.

elankath: I suggest leaving these phase renaming decisions to the overhaul. After reading the arguments above and thinking on it a bit, I agree that we can simply preserve the current phase change from `<empty>` to `CrashLoopBackoff` (yes, naming is not appropriate but this was decision taken before us) when there is an error not just on  VM instance creation but on post-creation/startup logic. 

The argument supporting this is that this is consistent phase handling on error wrt API. The argument against this is that it can confuse operators since they expect such an error code to indicate that the VM it-self was not created. However, we can mitigate this by documenting change of expected behavior in FAQ/release-notes.

unmarshall: >We decided to use a separate stage for the error in startup updates. Now if we want to follow the definition of CLBF, then it can be argued that startup activities should be included in it. Then in that case, we need to have proper logs that can help us differentiate why the machine is in CLBF and this should be reflected in the status of the Machine object as well.

In the `triggerCreationFlow` when a call is made to `driver.CreateMachine`. If an error is returned then that is handled in method [machineCreateErrorHandler](https://github.com/gardener/machine-controller-manager/blob/master/pkg/util/provider/machinecontroller/machine_util.go#L480-L515) which then calls [getCreateFailurePhase](https://github.com/gardener/machine-controller-manager/blob/master/pkg/util/provider/machinecontroller/machine_util.go#L569-L582) to get the phase. With this proposal, if an update failed then an error will be returned which will have the same control flow.  The error code will be used to determine the retry interval and if the timeout has not happened then it will simply return `CrashLoopBackoff` as the phase which will then trigger invocation of `driver.CreateMachine` again.
So with minimal changes you will achieve what you wanted to, right?

rishabh-11: >So with minimal changes you will achieve what you wanted to, right?

yes. We just need to add the reason for CLBF in the machine status and have additional logs as well.

elankath: > > So with minimal changes you will achieve what you wanted to, right?
> 
> yes. We just need to add the reason for CLBF in the machine status and have additional logs as well.

Will take note of this.

docs/proposals/update-machine.md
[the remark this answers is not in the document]
elankath: addressed review comments

docs/proposals/update-machine.md
elankath: Calling this `InstanceInitialization`

docs/proposals/update-machine.md
elankath: Fixed

elankath: Addressed point of modification to `GetMachineStatus` semantics

elankath: Guys, after thinking on this a lot today and yesterday and realizing that provider implementations are now forced to check whether there is a VM exists but not is fully initialized and that this logic is now forced into BOTH `GetMachineStatus` and `CreateMachine`, I am proposing that we stick to a new method `InitializeMachine` which is distinct from `UpdateMachine` and is meant for instance initialization. This makes logic must simpler since `CreateMachine` complexity is kept minimal.

elankath: Due to the point that Himanshu, mentioned above, the job of `Driver.CreateMachine` is becoming more and more complex. After thinking on this yesterday and today, I would propose simplifying this by introducing a new `Driver.InitializeMachine` which is separate from `Driver.UpdateMachine`. All post-create initialization logic set such as disabling src-dest checks (aws) and ipv6 address assignment (aws), etc. will be put in here. This method will be invoked only once after `Driver.CreateMachine` is invoked successfully. If `Driver.InitializeMachine` fails it will return `status.Error(codes.Initialization)` as the error. `Driver.GetMachine` also will return `status.Error(codes.Initialization)` if the VM is not fully initialized. The MCM will invoke `Driver.InitializeMachine` if `Driver.GetMachine` returns such a status.

rishabh-11 (approved):
/lgtm

himanshu-kun (approved):
/lgtm

PR #879, opened 2023-12-11
Fixed MCM restart during cleanup

sssash18:
This PR fixes the issue where MCM was unable to restart during cleanup which lead to cleanup failure

Fixes #

The logs for the restarted MCM and MC will now be appended to the existing logs itself.

Screenshot 2023-12-12 at 10 37 28 AM

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/test/integration/common/framework.go
+++ b/pkg/test/integration/common/framework.go
@@ -689,9 +689,9 @@ func (c *IntegrationTestFramework) setupMachineClass() error {
 
 // rotateLogFile takes file name as input and returns a file object obtained by os.Create
 // If the file exists already then it renames it so that a new file can be created
-func rotateAndAppendLogFile(fileName string, isRotate bool) (*os.File, error) {
-	if !isRotate {
-		if _, err := os.Stat(fileName); err != nil {
+func rotateOrAppendLogFile(fileName string, shouldRotate bool) (*os.File, error) {
+	if !shouldRotate {
+		if _, err := os.Stat(fileName); os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


pkg/test/integration/common/framework.go
```diff
@@ -689,7 +689,13 @@ func (c *IntegrationTestFramework) setupMachineClass() error {
 
 // rotateLogFile takes file name as input and returns a file object obtained by os.Create
 // If the file exists already then it renames it so that a new file can be created
-func rotateLogFile(fileName string) (*os.File, error) {
+func rotateAndAppendLogFile(fileName string, isRotate bool) (*os.File, error) {
```
himanshu-kun, line 692: `isRotate` -> `shouldRotate`

himanshu-kun, line 692: function could be named better, `rotateOrAppendLogFile()`

pkg/test/integration/common/framework.go
```diff
@@ -689,7 +689,13 @@ func (c *IntegrationTestFramework) setupMachineClass() error {
...
+	if !isRotate {
+		if _, err := os.Stat(fileName); err != nil {
```
himanshu-kun, line 694: we should check that the error is `file not found` then only create one

himanshu-kun (changes requested):
looks good, hope you have tested the cases of file not present and others..
Also pls add the logs showing the appending of new set of logs on restart..in either MC or MCM

himanshu-kun (approved):
/lgtm thanks for the changes!

PR #881, opened 2023-12-13
Made the gen-crd-api-reference-docs binary platform independent

### LINKED ISSUE #694: `make generate` doesn't work on Linux

/kind bug
/priority 3

**What happened**:

The following command fails on Linux:

```shell
$ make generate
```

The error:

```
./hack/api-reference/generate-spec-doc.sh: 2: ./gen-crd-api-reference-docs: Exec format error
```

The problem seems to stem from the `gen-crd-api-reference-docs` binary, because it's built for Mac:

```shell
$ file hack/api-reference/gen-crd-api-reference-docs 
hack/api-reference/gen-crd-api-reference-docs: Mach-O 64-bit x86_64 executable
```

`make generate` shouldn't fail.

Run `make generate` on a system that does not support `Mach-O`.

The problem can be fixed by getting the correct binary from the [official releases](https://github.com/ahmetb/gen-crd-api-reference-docs/releases/), but I think this can be solved more elegantly. Potentially by adding the Go module in `tools.go` and referencing it from there?

sssash18:
This PR installs the binary for the gen-crd-api-reference-docs each time so it is now platform independent.

Fixes #694 

**Release note**:

```other operator
NONE
```

### DIFF
--- a/hack/api-reference/generate-spec-doc.sh
+++ b/hack/api-reference/generate-spec-doc.sh
@@ -14,5 +14,6 @@
 
 cd ./hack/api-reference
 GO111MODULE=on go install github.com/ahmetb/gen-crd-api-reference-docs@latest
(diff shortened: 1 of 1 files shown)


aaronfern (approved):
/lgtm

PR #884, opened 2023-12-19
Env var  for IT tests

sssash18:
This PR introduces a new env var `TARGET_RESOURCE_GROUP`
This PR also corrects the use of `IS_CONTROL_CLUSTER_SEED` in the docs
Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/docs/development/integration_tests.md
+++ b/docs/development/integration_tests.md
@@ -31,8 +31,9 @@ If the Control Cluster is a Gardener Shoot cluster then,
 
 If the Control Cluster is a Gardener SEED cluster, then the suite ideally employs the already existing `MachineClass` and Secrets. However,
 
-1. Define the variable `IS_CONTROL_SEED` in the `.env` file and set it to `true`. 
+1. Define the variable `IS_CONTROL_CLUSTER_SEED` in the `.env` file and set it to `true`. 
 `Warning:` Make sure to set the `CONTROL_NAMESPACE` variable to the shoot namespace where the control plane of the target resides.
+1. Please pass `TARGET_RESOURCE_GROUP` in the `.env` file and set it to the name of the target cluster. This is only required for Azure clusters.
 1. (Optional) User can employ a custom `MachineClass` for the tests using below steps:
     1. Deploy a `Secret` named `test-mc-secret` (that contains the provider secret and cloud-config) in the shoot namespace of the Control Cluster. That is, the value of `metadata.namespace` should be `technicalID` of the Shoot and it will be of the pattern `shoot--<project>--<shoot-name>`.
     1. Create a `dev/machineclassv1.yaml` file and add an entry in the `.env` file with `MACHINECLASS_V1=dev/machineclassv1.yaml`.
--- a/docs/development/integration_tests.md
+++ b/docs/development/integration_tests.md
@@ -33,7 +33,7 @@ If the Control Cluster is a Gardener SEED cluster, then the suite ideally employ
 
 1. Define the variable `IS_CONTROL_CLUSTER_SEED` in the `.env` file and set it to `true`. 
 `Warning:` Make sure to set the `CONTROL_NAMESPACE` variable to the shoot namespace where the control plane of the target resides.
-1. Please pass `TARGET_RESOURCE_GROUP` in the `.env` file and set it to the name of the target cluster. This is compulsory for Azure clusters.
+1. Please pass `TARGET_RESOURCE_GROUP` in the `.env` file. It will be used for the `ResourceGroupName` in Azure clusters. Keep it as target cluster name for gardener shoot clusters.This is compulsory for Azure clusters.
 1. (Optional) User can employ a custom `MachineClass` for the tests using below steps:
     1. Deploy a `Secret` named `test-mc-secret` (that contains the provider secret and cloud-config) in the shoot namespace of the Control Cluster. That is, the value of `metadata.namespace` should be `technicalID` of the Shoot and it will be of the pattern `shoot--<project>--<shoot-name>`.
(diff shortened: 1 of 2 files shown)


aaronfern (commented):
Thanks for the changes @sssash18!

Just a quick question. Where is this new env var used?

sssash18:
> Thanks for the changes @sssash18!
> 
> Just a quick question. Where is this new env var used?

Now the use has been added @aaronfern

pkg/test/integration/common/framework.go
```diff
@@ -44,14 +44,13 @@ import (
...
-
-	// Suffix for the`kubernetes-io-cluster` tag and cluster name for the orphan resource tracker
-	targetClusterPlaceholder = "integration-test-cluster"
 )
 
 var (
 	// path for storing log files (mcm & mc processes)
 	targetDir = filepath.Join("..", "..", "..", ".ci", "controllers-test", "logs")
+	// Suffix for the`kubernetes-io-cluster` tag and cluster name for the orphan resource tracker
+	targetClusterName = os.Getenv("TARGET_RESOURCE_GROUP")
```
aaronfern, line 53: What happens if the env var is not provided? is there a default value?

docs/development/integration_tests.md
```diff
@@ -31,8 +31,9 @@ If the Control Cluster is a Gardener Shoot cluster then,
 
 If the Control Cluster is a Gardener SEED cluster, then the suite ideally employs the already existing `MachineClass` and Secrets. However,
 
-1. Define the variable `IS_CONTROL_SEED` in the `.env` file and set it to `true`. 
+1. Define the variable `IS_CONTROL_CLUSTER_SEED` in the `.env` file and set it to `true`. 
 `Warning:` Make sure to set the `CONTROL_NAMESPACE` variable to the shoot namespace where the control plane of the target resides.
+1. Please pass `TARGET_RESOURCE_GROUP` in the `.env` file and set it to the name of the target cluster. This is only required for Azure clusters.
```
aaronfern, line 36: Is this env var mandatory for azure clusters? if so maybe need to be a bit more clear

aaronfern (commented):
Thanks for addressing my question @sssash18!
I have a couple more questions, please address them

pkg/test/integration/common/framework.go
sssash18: This needs to be provided as we can't have a deafult value for the target cluster

docs/development/integration_tests.md
sssash18: yes this is mandatory for Azure

pkg/test/integration/common/framework.go
aaronfern: yes, but if it's not compulsory for other cloud providers then does the absence of this env var not impact anything?

sssash18: no because it is not even used in other cloudprovider code , so no impact whether it's present or absent.

sssash18: it might impact some tags, but testing wise other providers don't need it so tests are running fine.

aaronfern: Okay, fair enough then

aaronfern (commented):
Looks good in principle 
Can you change the description and the title as it's now not only a doc change?

sssash18:
> Looks good in principle Can you change the description and the title as it's now not only a doc change?

Done @aaronfern

aaronfern (approved):
Thank you! 
/lgtm

docs/development/integration_tests.md
```diff
@@ -31,8 +31,9 @@ If the Control Cluster is a Gardener Shoot cluster then,
 
 If the Control Cluster is a Gardener SEED cluster, then the suite ideally employs the already existing `MachineClass` and Secrets. However,
 
-1. Define the variable `IS_CONTROL_SEED` in the `.env` file and set it to `true`. 
+1. Define the variable `IS_CONTROL_CLUSTER_SEED` in the `.env` file and set it to `true`. 
 `Warning:` Make sure to set the `CONTROL_NAMESPACE` variable to the shoot namespace where the control plane of the target resides.
+1. Please pass `TARGET_RESOURCE_GROUP` in the `.env` file and set it to the name of the target cluster. This is compulsory for Azure clusters.
```
himanshu-kun, line 36: the doc is designed for opensource use of IT of MCM. So its not necessary to set this variable to the `target cluster name`, but it should be set to the name of resource group (an azure specific concept) where the resources of the cluster are made by MCM . Its only in gardener shoot cluster cases this variable is set to the cluster name..
kindly improve the statement keeping the above in mind.

pkg/test/integration/common/framework.go
```diff
@@ -44,14 +44,13 @@ import (
...
 	dwdIgnoreScalingAnnotation = "dependency-watchdog.gardener.cloud/ignore-scaling"
-
-	// Suffix for the`kubernetes-io-cluster` tag and cluster name for the orphan resource tracker
-	targetClusterPlaceholder = "integration-test-cluster"
 )
 
 var (
 	// path for storing log files (mcm & mc processes)
 	targetDir = filepath.Join("..", "..", "..", ".ci", "controllers-test", "logs")
+	// Suffix for the`kubernetes-io-cluster` tag and cluster name for the orphan resource tracker
```
himanshu-kun, line 52:
```suggestion
	// Suffix for the`kubernetes-io-cluster` tag and cluster name for the orphan resource tracker. Currently relevant only for Azure
```

himanshu-kun (changes requested):
looks good, a small update is requested for better open source consumption.

docs/development/integration_tests.md
sssash18: Done

himanshu-kun (approved):
/lgtm

PR #894, opened 2024-01-11
Added `Architecture` in NodeTemplate

sssash18:
This PR adds `architecture` field in the node template
Fixes part of #https://github.com/gardener/autoscaler/issues/122

**Release note**:

```other operator
Architecture field added in the nodetemplate. This will allow CA to pickup architecture from machine class and schedule pods on relevant arch nodes.
```

### DIFF
--- a/kubernetes/crds/machine.sapcloud.io_machineclasses.yaml
+++ b/kubernetes/crds/machine.sapcloud.io_machineclasses.yaml
@@ -56,7 +56,7 @@ spec:
               and other node info required to scale nodegroup from zero
             properties:
               architecture:
-                description: Architecture of the node belonging to nodeGroup
+                description: CPU Architecture of the node belonging to nodeGroup
                 type: string
               capacity:
                 additionalProperties:
--- a/kubernetes/crds/machine.sapcloud.io_machineclasses.yaml
+++ b/kubernetes/crds/machine.sapcloud.io_machineclasses.yaml
@@ -78,7 +78,6 @@ spec:
                 description: Zone of the expected node belonging to nodeGroup
                 type: string
             required:
-            - architecture
             - capacity
             - instanceType
             - region
--- a/kubernetes/crds/machine.sapcloud.io_machineclasses.yaml
+++ b/kubernetes/crds/machine.sapcloud.io_machineclasses.yaml
@@ -55,6 +55,9 @@ spec:
             description: NodeTemplate contains subfields to track all node resources
               and other node info required to scale nodegroup from zero
             properties:
+              architecture:
+                description: Architecture of the node belonging to nodeGroup
+                type: string
               capacity:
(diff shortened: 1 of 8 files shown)


rishabh-11 (changes requested):
You have added architecture as a required field. Is it a required field, though? It is needed only in case of arm64 nodes, right? Can we keep it as an optional field?

sssash18:
> You have added architecture as a required field. Is it a required field, though? It is needed only in case of arm64 nodes, right? Can we keep it as an optional field?

Yes could be done , because by default the node will always have amd64 as arch label. This is true in gardener scenario.
But for standalone mcm , this can lead to a case when pod with amd64 affinity might not get scheduled as the node does not have any arch field even though the node is amd64.

unmarshall:
Looks good. Please add a release note in the PR.

kubernetes/crds/machine.sapcloud.io_machineclasses.yaml
```diff
@@ -55,6 +55,9 @@ spec:
             description: NodeTemplate contains subfields to track all node resources
               and other node info required to scale nodegroup from zero
             properties:
+              architecture:
+                description: Architecture of the node belonging to nodeGroup
```
unmarshall, line 59: This is actually a processor/CPU architecture. Should we call it that?

pkg/apis/machine/v1alpha1/machineclass_types.go
```diff
@@ -84,4 +84,8 @@ type NodeTemplate struct {
 
 	// Zone of the expected node belonging to nodeGroup
 	Zone string `json:"zone"`
+
+	// +optional
+	// Architecture of the node belonging to nodeGroup
```
unmarshall, line 89: same as the above comment.

unmarshall (commented):
Looks good, just a minor comment.

kubernetes/crds/machine.sapcloud.io_machineclasses.yaml
elankath: The field is called architecture by convention in k8s. Example [NodeSystemInfo.Architecture](https://pkg.go.dev/k8s.io/api/core/v1#NodeSystemInfo.Architecture)

unmarshall: Maybe i was not clear. I was not referring to the name of the property which should be called `Architecture` and that is the reason my comment is for the `description` line where it refers to architecture as `Architecture of the node`

sssash18: Sure will make the changes.

unmarshall (approved):

rishabh-11 (approved):
/lgtm

PR #906, opened 2024-04-16
Diagnostics Logs for MCM

### LINKED ISSUE #905: Add diagnostic logs for MCM

/area logging
/kind enhancement
/priority 3

1. Log when machine phase changes to `Failed`
1. Log machines that are marked to be deleted by CA.

**Why is this needed**:
For further information when debugging issues.

elankath:

Fixes #905

**Release note**:

```other developer
NONE
```

### DIFF
--- a/pkg/controller/machineset.go
+++ b/pkg/controller/machineset.go
@@ -152,7 +152,7 @@ func (c *controller) machineSetUpdate(old, cur interface{}) {
 	// that bad as MachineSets that haven't met expectations yet won't
 	// sync, and all the listing is done using local stores.
 	if oldMachineSet.Spec.Replicas != currentMachineSet.Spec.Replicas {
-		klog.V(4).Infof("%v updated. Desired machine count change: %d->%d", currentMachineSet.Name, oldMachineSet.Spec.Replicas, currentMachineSet.Spec.Replicas)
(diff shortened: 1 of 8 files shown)


pkg/controller/machineset.go
```diff
@@ -427,7 +427,9 @@ func (c *controller) manageReplicas(ctx context.Context, allMachines []*v1alpha1
 		}
 		klog.V(2).Infof("Too many replicas for %v %s/%s, need %d, deleting %d", machineSet.Kind, machineSet.Namespace, machineSet.Name, (machineSet.Spec.Replicas), diff)
 
+		logMachinesWithPriorityAnnotation(activeMachines)
```
rishabh-11, line 430: Actually, each machine will have a priority annotation set. Should we just log machines with priority set to `1` instead?

rishabh-11 (changes requested):
one minor comment. rest looks good. thanks for the PR.

pkg/controller/machineset.go
elankath: Hmm..ok, I thought it might be useful to see if there was something strange/diff about priority. But yes, to avoid overlog we can limit this to p1.

rishabh-11 (approved):
/lgtm

rishabh-11 (approved):
/lgtm

PR #920, opened 2024-06-24
Improved context with timeout for DrainNode

### LINKED ISSUE #785: Drain Timeout is not respected during client side throttling

/area robustness
/kind bug
/priority 2

**What happened**:
A case was seen where drain timeout was set to 2hrs but the drain ended up going till 11hrs.
If drainTimeout is 2hrs, then using podEvictionInterval(20sec) , we calculate `maxEvictRetries` as 360.
In case of client side thottling , the evict call takes huge time and thus the interval b/w 2 pod eviction requests becomes more than 100sec also.
Since currently we don't have context based cancellation for `evictPodWithoutPVInternal` , we rely just on `maxEvictRetries` to exhaust to come out of the loop.  This can happen in the case where pod eviction runs into a timeout. 
This leads to 11hrs or more of drain, and no force delete of the machine is done.

Drain timeout to be respected in every situtation.

Context based cancellation should be used effectively.

**Environment**:

- Others:

sssash18:
This PR introduces a drain Context in `RunDrain` with a timeout so that context cancels and machine can be force deleted later.
Fixes #785 

**Release note**:

```other operator
Drain timeout is now correctly honored for Pod eviction during Machine Drain
```

### DIFF
--- a/pkg/util/provider/drain/drain.go
+++ b/pkg/util/provider/drain/drain.go
@@ -996,12 +996,6 @@ func (o *Options) evictPodWithoutPVInternal(ctx context.Context, attemptEvict bo
 		if i >= nretries {
 			attemptEvict = false
 		}
-		select {
-		case <-ctx.Done():
-			// Timeout occurred. Abort and report the remaining pods.
-			returnCh <- fmt.Errorf("timeout occured while waiting for pod %q terminating scheduled on node %v", pod.Name, pod.Spec.NodeName)
-		default:
-		}
 		if attemptEvict {
 			err = o.evictPod(ctx, pod, policyGroupVersion)
 		} else {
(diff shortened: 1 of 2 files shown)


aaronfern (commented):
/lgtm

aaronfern:
Please add a release note, something like below or feel free to come up with a different one
`A bug in the drain process where the drain timeout was not respected has been fixed` or
`Timeout based context cancellation will now be followed for machine drains`

elankath (approved):
You can change the release note to : "Drain timeout is now correctly honored for Pod eviction during Machine Drain" ? 

Can you do a test of this with some short timeouts and Pods with large number of PVs to see if there are no surprises ?

sssash18:
> You can change the release note to : "Drain timeout is now correctly honored for Pod eviction during Machine Drain" ?
> 
> Can you do a test of this with some short timeouts and Pods with large number of PVs to see if there are no surprises ?

Performed the testing, the context cancellation leads to force delete even if pod with PVs are in process of eviction.

rishabh-11 (approved):
/lgtm

PR #932, opened 2024-07-23
Add namespace to dummy machine object so that driver can rely on the namespace being set

ScheererJ:

Add namespace to dummy machine object so that driver can rely on the namespace being set.

Some drivers may assume that the namespace of the machine resources is correctly set when a request is sent to them, e.g. see https://github.com/gardener/gardener/pull/10176. As the namespace of the machine is known due to its machine class, it should be always safe to set the correct namespace.

Fixes #

**Release note**:

```other operator

```


rfranzke (approved):
/lgtm

rishabh-11 (approved):
/lgtm

Issue #936, opened 2024-08-20
Ensure that `DeleteMachine` call is made even in case of a `NotFound` error from `Driver.GetMachineStatus` in the deletion flow


rishabh-11:
**How to categorize this issue?**

/area control-plane
/area robustness
/kind enhancement
/priority 2

**What would you like to be added**:
Remove `getVMStatus` from `triggerDeletionFlow`. The consequence is that no other step in the deletion flow will be skipped and the call to `Driver.DeleteMachine` will be made. This will ensure that there are no orphan resources left and we don't have to rely on the orphan collection logic of MCM.

**Why is this needed**:

In azure, the creation of VM and NIC is done separately (it cannot be done together as the cloud provider does not have this functionality). In this case, the NIC gets created but the VM does not
```
2024-08-08 05:45:16	
{"log":"Successfully created NIC: [ResourceGroup: shoot--hc-ap21--340-83-hdl, NIC: [Name: shoot--hc-ap21--340-83-hdl-edge-b-z2-75c67-hqw5h-nic, ID: /subscriptions/a14b094d-362f-4924-b343-f3a65dd9424a/resourceGroups/shoot--hc-ap21--340-83-hdl/providers/Microsoft.Network/networkInterfaces/shoot--hc-ap21--340-83-hdl-edge-b-z2-75c67-hqw5h-nic]]","pid":"1","severity":"INFO","source":"driver.go:391"}

[...]

PR #919, opened 2024-06-21
Introduce a feature to propagate the machine name to user data and as label to the node

oliver-goetz:
**What this PR does / why we need it**:
This PR adds two new features to MCM which are required to enable `node-agent-authorizer` webhook based authentication for `gardener-node-agent`. Please see https://github.com/gardener/gardener/pull/10014

- MCM replaces the magic string `<<MACHINE_NAME>>` in user data with the name of the machine when it is bootstrapped.
- MCM adds the name of the machine to `node.gardener.cloud/machine-name` label of the corresponding node.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```feature developer
MCM replaces the magic string `<<MACHINE_NAME>>` in user data with the name of the machine when it is bootstrapped.
```
```feature developer
MCM adds the name of the machine to `node.gardener.cloud/machine-name` label of the corresponding node.
```

PR #940, opened 2024-09-05
Modify Machine Creation flow to make sure node label is updated before initialization of VM. Modify Deletion flow to call DeleteMachine even if VM is not found.

### LINKED ISSUE #933: Machine is never initialised if `Driver.InitializeMachine` returns `NotFound` error code for VM

/area robustness
/kind bug
/priority 1

**What happened**:
According to https://github.com/gardener/machine-controller-manager/blob/ff8261398277c3e5a481f06cfb57c417dfd07754/pkg/util/provider/machinecontroller/machine.go#L609-#L611, if `NotFound` error code is returned by `driver.InitialiseMachine`, the initialisation of the VM is skipped. This can lead to problems in the following case:- 

```
2024-08-08 07:00:21 | {"log":"Creating a VM for machine \"shoot--garden--aws-ha-eu4-cpu-worker-z1-6dc96-9j7pm\", please wait!","pid":"1","severity":"INFO","source":"machine.go:392"}Show context
  |   | 2024-08-08 07:00:21 | {"log":"Machine creation request has been recieved for \"shoot--garden--aws-ha-eu4-cpu-worker-z1-6dc96-9j7pm\"","pid":"1","severity":"INFO","source":"core.go:82"}
  |   | 2024-08-08 07:00:22 | {"log":"Waiting for VM with Provider-ID \"aws:///eu-west-1/i-0cf8cb1aa8a61dc8d\", for machine \"shoot--garden--aws-ha-eu4-cpu-worker-z1-6dc96-9j7pm\" to be visible to all AWS endpoints","pid":"1","severity":"INFO","source":"core.go:238"}
  |   | 2024-08-08 07:00:22 | {"log":"VM with Provider-ID \"aws:///eu-west-1/i-0cf8cb1aa8a61dc8d\", for machine \"shoot--garden--aws-ha-eu4-cpu-worker-z1-6dc96-9j7pm\" should be visible to all AWS endpoints now","pid":"1","severity":"INFO","source":"core.go:249"}
  |   | 2024-08-08 07:00:22 | {"log":"VM with Provider-ID: \"aws:///eu-west-1/i-0cf8cb1aa8a61dc8d\" created for Machine: \"shoot--garden--aws-ha-eu4-cpu-worker-z1-6dc96-9j7pm\"","pid":"1","severity":"INFO","source":"core.go:250"}
  |   | 2024-08-08 07:00:22 | {"log":"Created new VM for machine: \"shoot--garden--aws-ha-eu4-cpu-worker-z1-6dc96-9j7pm\" with ProviderID: \"aws:///eu-west-1/i-0cf8cb1aa8a61dc8d\" and backing node: \"\"","pid":"1","severity":"INFO","source":"machine.go:405"}
  |   | 2024-08-08 07:00:22 | {"log":"Initializing VM instance for Machine \"shoot--garden--aws-ha-eu4-cpu-worker-z1-6dc96-9j7pm\"","pid":"1","severity":"INFO","source":"machine.go:596"}
  |   | 2024-08-08 07:00:22 | {"log":"Error occurred while initializing VM instance for machine \"shoot--garden--aws-ha-eu4-cpu-worker-z1-6dc96-9j7pm\": machine codes error: code = [NotFound] message = [AWS plugin is returning no VM instances backing this machine object]","pid":"1","severity":"ERR","source":"machine.go:604"}
  |   | 2024-08-08 07:00:22 | {"log":"No VM instance found for machine \"shoot--garden--aws-ha-eu4-cpu-worker-z1-6dc96-9j7pm\". Skipping VM instance initialization.","pid":"1","severity":"WARN","source":"machine.go:610"}
  |   | 2024-08-08 07:00:22 | {"log":"Machine labels/annotations UPDATE for \"shoot--garden--aws-ha-eu4-cpu-worker-z1-6dc96-9j7pm\"","pid":"1","severity":"INFO","source":"machine.go:552"}Show context
  |   | 2024-08-08 07:00:22 | {"log":"reconcileClusterMachine: Stop for \"shoot--garden--aws-ha-eu4-cpu-worker-z1-6dc96-9j7pm\"","pid":"1","severity":"INFO","source":"machine.go:178"}
```

Here, for machine `shoot--garden--aws-ha-eu4-cpu-worker-z1-6dc96-9j7pm`, the VM was successfully created, but the initialisation failed as the VM was not found at a later instant. We know that this is an issue from the cloud provider side, but it can happen. In this case, the initialisation is skipped, and the machine object is updated with the `providerID` and `NodeName` label.

In the next reconciliation, the `GetMachineStatus` also has the same transient issue on the cloud provider, and the VM has still not been found. But because the node label was set in the previous reconciliation, https://github.com/gardener/machine-controller-manager/blob/ff8261398277c3e5a481f06cfb57c417dfd07754/pkg/util/provider/machinecontroller/machine.go#L390 is never executed and hence the VM is never initialised, but the machine will be moved to `Pending` state (and eventually to `Running` once the Node is registered)

```
2024-08-08 07:00:22 | {"log":"reconcileClusterMachine: Start for \"shoot--garden--aws-ha-eu4-cpu-worker-z1-6dc96-9j7pm\" with phase:\"\", description:\"\"","pid":"1","severity":"INFO","source":"machine.go:116"}
  |   | 2024-08-08 07:00:23 | {"log":"Get request has been recieved for \"shoot--garden--aws-ha-eu4-cpu-worker-z1-6dc96-9j7pm\"","pid":"1","severity":"INFO","source":"core.go:411"}
  |   | 2024-08-08 07:00:23 | {"log":"For machine \"shoot--garden--aws-ha-eu4-cpu-worker-z1-6dc96-9j7pm\", obtained VM error status as: machine codes error: code = [NotFound] message = [AWS plugin is returning no VM instances backing this machine object]","pid":"1","severity":"WARN","source":"machine.go:382"}
  |   | 2024-08-08 07:00:23 | {"log":"Machine/status UPDATE for \"shoot--garden--aws-ha-eu4-cpu-worker-z1-6dc96-9j7pm\" during creation","pid":"1","severity":"INFO","source":"machine.go:578"}
```
 
This leads to a problem because things like `sourceDestCheck` can be enabled/disabled during the initialisation of VM. If the initialization is not done, the pods running on them can go into CLBF, as seen in canary issue no. 5533. 

Another problem is that if the `driver.InitialiseMachine` method keeps on failing, it is still possible for the kubelet to run on the created VM and register the corresponding Node object. This will lead to the scheduler seeing the node and scheduling pods on it, which will go into CLBF as the VM has not been properly initialised. 

VM initialisation should be retried in case of `NotFound` errors and pods should not scheduled on the node until the initialisation of the corresponding VM is successful. 

### LINKED ISSUE #934: Improve error handling when `InitializeMachine` method is not yet implement by a provider

/area logging|ops-productivity
/kind bug
/priority 2

**What happened**:
Currently the InitializeMachine is not implemented by every provider other than AWS. 
However as the call to the provider implementation results in `Unimplemented` error type, MCM logs the error and doesn't check the semantics to rather ignore this error. [https://github.com/gardener/machine-controller-manager/blob/ff8261398277c3e5a481f0[…]fb57c417dfd07754/pkg/util/provider/machinecontroller/machine.go](https://github.com/gardener/machine-controller-manager/blob/ff8261398277c3e5a481f06cfb57c417dfd07754/pkg/util/provider/machinecontroller/machine.go#L596-L608).
```go
  klog.V(3).Infof("Initializing VM instance for Machine %q", machine.Name)
    resp, err := c.driver.InitializeMachine(ctx, req)
    if err != nil {
        errStatus, ok := status.FromError(err)
        if !ok {
            klog.Errorf("Cannot decode Driver error for machine %q: %s. Unexpected behaviour as Driver errors are expected to be of type status.Status", machine.Name, err)
            return machineutils.LongRetry, err
        }
        klog.Errorf("Error occurred while initializing VM instance for machine %q: %s", machine.Name, err)
        switch errStatus.Code() {
        case codes.Unimplemented:
            klog.V(3).Infof("Provider does not support Driver.InitializeMachine - skipping VM instance initialization for %q.", machine.Name)
```
Can we mark this as an Info/Warning message instead of error. 

The system to ignore the error when the method is not yet implemented and just log the behavior as an Info log. 
Create a machine in Azure where this method is not implemented and you should see the logs containing this error. 

**Environment**:

- Others: azure|gcp|openstack

thiyyakat:

The PR changes `triggerCreationFlow` to update machine labels before proceeding to initialization.  This was done to make sure that the labels are always updated even in the case of initialization failures. 

It also changes `getVMStatus` to update the labels in the following way:
1. list all the nodes and find matching machine name label ( effective when #919 and https://github.com/gardener/gardener/pull/10014 are released)
2. If the above does not find a node name, do a `GetMachineStatus` call and populate the node name. 

In addition, `getVMStatus` will always redirect to `initiateDrain`. 

A change was also made to remove error logs when `InitializeMachine` is not implemented by a provider.

Fixes #934 #936
Fixes part of #933 

The changes in the PR were tested by doing the following:
Manually returned an error from machine-controller-manager-provider-azure's `CreateMachine` code after NIC is created, such that the VM is not created. The machine was then marked for deletion. The logs showed that the NIC was deleted successfully.

Manually returned codes.Uninitialized error code from InitializeMachine for AWS. On triggering creation of machine, found that initialization is tried in a loop after shortRetry. 

Manually returned codes.Uninitialized from `initializeMachine` after the call to `driver.initializeMachine`. The logs show that the machine state update is successful. In the next reconciliation, since the machine is found to be initialized, machine goes to running.

**Release note**:

```other operator
`getVMStatus` always redirects to `InitiateDrain`. It also populates the node label on the machine object by checking `node.gardener.cloud/machine-name` label on the nodes. 
```

```bugfix operator
Fixed a bug where failure of machine initialization caused label updates to not happen. 
```

### DIFF
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -521,6 +521,8 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 		if machine.Labels[v1alpha1.NodeLabelKey] == "" || machine.Spec.ProviderID == "" {
 			klog.V(2).Infof("Found VM with required machine name. Adopting existing machine: %q with ProviderID: %s", machineName, getMachineStatusResponse.ProviderID)
 		}
+		//clean me up. I'm dirty.
+		//TODO@thiyyakat add a pointer to a boolean variable indicating whether initialization has happened successfully.
 		nodeName = getMachineStatusResponse.NodeName
 		providerID = getMachineStatusResponse.ProviderID
 	}
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -545,8 +545,6 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 			klog.Warningf("Machine UPDATE failed for %q. Retrying, error: %s", machine.Name, err)
 		} else {
 			klog.V(2).Infof("Machine labels/annotations UPDATE for %q", machine.Name)
-			// Return error even when machine object is updated
-			err = fmt.Errorf("Machine creation in process. Machine UPDATE successful")
 		}
 	}
 	if uninitializedMachine {
@@ -554,6 +552,8 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 		if err != nil {
 			return retryPeriod, err
 		}
+		// Return error even when machine object is updated
+		err = fmt.Errorf("Machine creation in process. Machine UPDATE successful")
 		return machineutils.ShortRetry, err
 	}
 	if machine.Status.CurrentStatus.Phase == "" || machine.Status.CurrentStatus.Phase == v1alpha1.MachineCrashLoopBackOff {
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -528,13 +528,7 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 		providerID = getMachineStatusResponse.ProviderID
 	}
 
-	if uninitializedMachine {
-		retryPeriod, err := c.initializeMachine(ctx, createMachineRequest.Machine, createMachineRequest.MachineClass, createMachineRequest.Secret)
-		if err != nil {
-			return retryPeriod, err
-		}
-	}
-
+	//Update labels
 	_, machineNodeLabelPresent := createMachineRequest.Machine.Labels[v1alpha1.NodeLabelKey]
 	_, machinePriorityAnnotationPresent := createMachineRequest.Machine.Annotations[machineutils.MachinePriority]
 
@@ -590,10 +584,15 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 
 		return machineutils.ShortRetry, err
 	}
+	if uninitializedMachine {
+		retryPeriod, err := c.initializeMachine(ctx, createMachineRequest.Machine, createMachineRequest.MachineClass, createMachineRequest.Secret)
+		if err != nil {
+			return retryPeriod, err
+		}
+	}
 
 	return machineutils.LongRetry, nil
 }
-
 func (c *controller) initializeMachine(ctx context.Context, machine *v1alpha1.Machine, machineClass *v1alpha1.MachineClass, secret *corev1.Secret) (machineutils.RetryPeriod, error) {
 	req := &driver.InitializeMachineRequest{
 		Machine:      machine,
@@ -612,9 +611,6 @@ func (c *controller) initializeMachine(ctx context.Context, machine *v1alpha1.Ma
 		switch errStatus.Code() {
 		case codes.Unimplemented:
 			klog.V(3).Infof("Provider does not support Driver.InitializeMachine - skipping VM instance initialization for %q.", machine.Name)
-			return 0, nil
-		case codes.NotFound:
-			klog.Warningf("No VM instance found for machine %q. Skipping VM instance initialization.", machine.Name)
 			return 0, nil
 		}
 		updateRetryPeriod, updateErr := c.machineStatusUpdate(
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -377,7 +377,6 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 			Secret:       createMachineRequest.Secret,
 		},
 	)
-
 	if err != nil {
 		// VM with required name is not found.
 		machineErr, ok := status.FromError(err)
@@ -387,7 +386,6 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 			return machineutils.MediumRetry, err
 		}
 		klog.Warningf("For machine %q, obtained VM error status as: %s", machineName, machineErr)
-
 		// Decoding machine error code
 		switch machineErr.Code() {
 		case codes.NotFound, codes.Unimplemented:
@@ -404,13 +402,10 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 					klog.Errorf("Error while creating machine %s: %s", machine.Name, err.Error())
 					return c.machineCreateErrorHandler(ctx, machine, createMachineResponse, err)
 				}
-
 				nodeName = createMachineResponse.NodeName
 				providerID = createMachineResponse.ProviderID
-
 				// Creation was successful
 				klog.V(2).Infof("Created new VM for machine: %q with ProviderID: %q and backing node: %q", machine.Name, providerID, getNodeName(machine))
-
 				// If a node obj already exists by the same nodeName, treat it as a stale node and trigger machine deletion.
 				// TODO: there is a case with Azure where the VM may join the cluster before the CreateMachine call is completed,
 				// and that would make an otherwise healthy node, be marked as stale.
@@ -607,12 +602,11 @@ func (c *controller) initializeMachine(ctx context.Context, machine *v1alpha1.Ma
 			klog.Errorf("Cannot decode Driver error for machine %q: %s. Unexpected behaviour as Driver errors are expected to be of type status.Status", machine.Name, err)
 			return machineutils.LongRetry, err
 		}
-		klog.Errorf("Error occurred while initializing VM instance for machine %q: %s", machine.Name, err)
-		switch errStatus.Code() {
-		case codes.Unimplemented:
+		if errStatus.Code() == codes.Unimplemented {
 			klog.V(3).Infof("Provider does not support Driver.InitializeMachine - skipping VM instance initialization for %q.", machine.Name)
 			return 0, nil
 		}
+		klog.Errorf("Error occurred while initializing VM instance for machine %q: %s", machine.Name, err)
 		updateRetryPeriod, updateErr := c.machineStatusUpdate(
 			ctx,
 			machine,
@@ -629,14 +623,11 @@ func (c *controller) initializeMachine(ctx context.Context, machine *v1alpha1.Ma
 			},
 			machine.Status.LastKnownState,
 		)
-
 		if updateErr != nil {
 			return updateRetryPeriod, updateErr
 		}
-
 		return machineutils.ShortRetry, err
 	}
-
 	klog.V(3).Infof("VM instance %q for machine %q was initialized", resp.ProviderID, machine.Name)
 	return 0, nil
 }
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -539,20 +539,23 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 		if clone.Annotations[machineutils.MachinePriority] == "" {
 			clone.Annotations[machineutils.MachinePriority] = "3"
 		}
-
 		clone.Spec.ProviderID = providerID
 		_, err := c.controlMachineClient.Machines(clone.Namespace).Update(ctx, clone, metav1.UpdateOptions{})
 		if err != nil {
 			klog.Warningf("Machine UPDATE failed for %q. Retrying, error: %s", machine.Name, err)
 		} else {
 			klog.V(2).Infof("Machine labels/annotations UPDATE for %q", machine.Name)
-
 			// Return error even when machine object is updated
 			err = fmt.Errorf("Machine creation in process. Machine UPDATE successful")
 		}
 		return machineutils.ShortRetry, err
 	}
-
+	if uninitializedMachine {
+		retryPeriod, err := c.initializeMachine(ctx, createMachineRequest.Machine, createMachineRequest.MachineClass, createMachineRequest.Secret)
+		if err != nil {
+			return retryPeriod, err
+		}
+	}
 	if machine.Status.CurrentStatus.Phase == "" || machine.Status.CurrentStatus.Phase == v1alpha1.MachineCrashLoopBackOff {
 		clone := machine.DeepCopy()
 		clone.Status.LastOperation = v1alpha1.LastOperation{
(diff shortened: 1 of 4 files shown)


gardener-robot-ci-2:
Thank you @thiyyakat for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -944,77 +944,94 @@ func (c *controller) setMachineTerminationStatus(ctx context.Context, deleteMach
...
+	// Node label is required for drain of node, therefore we try to update machine object before proceeding to drain.
+	isNodeLabelUpdated := false
+	//check if node name label is already present in machine object
+	nodeName := getMachineStatusRequest.Machine.Labels[v1alpha1.NodeLabelKey]
+	if isValidNodeName(nodeName) {
+		isNodeLabelUpdated = true
+	} else {
+		// Figure out node label either by checking all nodes for label matching machine name or retrieving it using GetMachineStatus
+		// get all nodes and check if any node has the machine name as label
+		var nodes []*v1.Node
```
unmarshall, line 964: extract the logic to look for nodeName matching the machine name into a separate function. That function should just find the matching nodeName and the update should happen in the caller.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -944,77 +944,94 @@ func (c *controller) setMachineTerminationStatus(ctx context.Context, deleteMach
 	return machineutils.ShortRetry, err
 }
 
-// getVMStatus tries to retrive VM status backed by machine
+// getVMStatus tries to retrieve VM status backed by machine
 func (c *controller) getVMStatus(ctx context.Context, getMachineStatusRequest *driver.GetMachineStatusRequest) (machineutils.RetryPeriod, error) {
```
unmarshall, line 948: the function is named `getVMStatus` but it does not return any status. This is a bit weird. Shouldn't it be called `updateMachineStatusAndNodeLabel` instead?

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -944,77 +944,94 @@ func (c *controller) setMachineTerminationStatus(ctx context.Context, deleteMach
...
-		nodeName := getMachineStatusRequest.Machine.Labels[v1alpha1.NodeLabelKey]
-		if nodeName == "" {
-			err = c.updateMachineNodeLabel(ctx, getMachineStatusRequest.Machine, statusResp.NodeName)
-			if err != nil {
-				return machineutils.ShortRetry, err
+	// Node label is required for drain of node, therefore we try to update machine object before proceeding to drain.
+	isNodeLabelUpdated := false
+	//check if node name label is already present in machine object
+	nodeName := getMachineStatusRequest.Machine.Labels[v1alpha1.NodeLabelKey]
+	if isValidNodeName(nodeName) {
```
unmarshall, line 959: this function is actually quite useless. It only checks if the node Name is empty. I would just remove this function.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -944,77 +944,94 @@ func (c *controller) setMachineTerminationStatus(ctx context.Context, deleteMach
...
+					if err != nil {
+						return machineutils.ShortRetry, err
+					}
+					isNodeLabelUpdated = true
+					break
+				}
 			}
 		}
-
+		if !isNodeLabelUpdated {
```
unmarshall, line 979: It would be better to rewrite this function.
An idea:
```go
func (c *controller) updateMachineNodeLabelAndStatus(ctx context.Context, machine *v1alpha1.Machine) {
   // calls getNodeName
   // if there is no error then updates the node label
   // if there is an error either in calling getNodeName or updating the node label then updates the status accordingly

}

func (c *controller) getNodeName(ctx context.Context, machine *v1alpha1.Machine) (string, error) {
	nodeName := machine.Labels[v1alpha1.NodeLabelKey]
	if len(strings.TrimSpace(nodeName)) > 0 {
		return nodeName, nil
	}
	matchingNode, err := c.fetchMatchingNode(ctx, machine.Name)	
	if err == nil && matchingNode != nil {
		return matchingNode.Name, nil
	}
	if err != nil {
		klog.Errorf("Error trying to get node matching machine %s: %v. Will try to get the node name by calling driver.GetMachineStatus instead.", machine.Name, err)
	}
	// call GetMachineStatus to get the node name	
}

```

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -550,20 +539,23 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
...
-			err = fmt.Errorf("Machine creation in process. Machine UPDATE successful")
 		}
+	}
+	if uninitializedMachine {
+		retryPeriod, err := c.initializeMachine(ctx, createMachineRequest.Machine, createMachineRequest.MachineClass, createMachineRequest.Secret)
+		if err != nil {
+			return retryPeriod, err
+		}
+		// Return error even when machine object is updated
+		err = fmt.Errorf("Machine creation in process. Machine UPDATE successful")
```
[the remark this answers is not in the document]
thiyyakat: If we look at the call stack, this error message is propagated to `reconcileClusterMachineKey` which calls `enqueueMachineAfter` where this message is used to print a log specifying the reason for the enqueue operation. 

This is not something new that has been introduced. We are preserving the original way. Changing the message to `Machine creation in process. Machine initialization (if required) and label update successful`

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -577,23 +542,57 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
...
 		}
-
 		return machineutils.ShortRetry, err
 	}
-
 	return machineutils.LongRetry, nil
 }
 
+func (c *controller) updateLabelsAndInitializeMachine(ctx context.Context, createMachineRequest *driver.CreateMachineRequest, nodeName, providerID string, shouldInitializeMachine bool) (retryPeriod machineutils.RetryPeriod, err error) {
+	_, machineNodeLabelPresent := createMachineRequest.Machine.Labels[v1alpha1.NodeLabelKey]
```
unmarshall, line 559: you could have used:
```go
metav1.HasAnnotation(createMachineRequest.Machine.ObjectMeta, v1alpha1.NodeLabelKey)
```

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -577,23 +542,57 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
...
+		clone.Labels[v1alpha1.NodeLabelKey] = nodeName
+		if clone.Annotations == nil {
+			clone.Annotations = make(map[string]string)
+		}
+		if clone.Annotations[machineutils.MachinePriority] == "" {
+			clone.Annotations[machineutils.MachinePriority] = "3"
+		}
+		clone.Spec.ProviderID = providerID
+		var updatedMachine *v1alpha1.Machine
+		updatedMachine, err = c.controlMachineClient.Machines(clone.Namespace).Update(ctx, clone, metav1.UpdateOptions{})
```
unmarshall, line 575: I am not too sure if the return value which you have now initialized `updatedMachine` to contains the revised version/generation. Can you double check this?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -577,23 +542,57 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
...
+			klog.Warningf("Machine labels/annotations UPDATE failed for %q. Will retry after VM initialization (if required), error: %s", createMachineRequest.Machine.Name, err)
+			clone = createMachineRequest.Machine.DeepCopy()
+		} else {
+			clone = updatedMachine
+			klog.V(2).Infof("Machine labels/annotations UPDATE for %q", clone.Name)
+			err = fmt.Errorf("Machine creation in process. Machine labels/annotations update is successful.")
+		}
+	}
+	if shouldInitializeMachine {
+		retryPeriod, err = c.initializeMachine(ctx, clone, createMachineRequest.MachineClass, createMachineRequest.Secret)
```
unmarshall, line 586: why have `initializeMachine` call from here? I would have just renamed this method to `updateMachineLabels` since we already have `initializeMachine`. The just call these functions from `triggerCreationFlow`. This also removes the need to pass `shouldInitializeMachine` as a method argument.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -577,23 +542,57 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
...
+		if err != nil {
+			return retryPeriod, err
+		}
+		// Return error even when machine object is updated
+		err = fmt.Errorf("Machine creation in process. Machine initialization (if required) is successful.")
```
unmarshall, line 591: messages given to `fmt.Errorf` should not start with a Capital letter and should not end with a dot. You will see warnings shown in the IDE. Please correct it. There are other places as well where i see this issue. Can you at least correct it in the functions that you are touching?

unmarshall (changes requested):

pkg/util/provider/machinecontroller/machine.go
thiyyakat: We returned a custom error from `InitializeMachine`, and saw that the status update went through.
```
E0912 09:40:01.094021   69341 machine.go:619] Error occurred while initializing VM instance for machine "shoot--i749592--it-worker-f73ns-z1-56778-ws5qx": machine codes error: code = [Uninitialized] message = [TESTING]
I0912 09:40:01.262395   69341 machine_util.go:565] Machine/status UPDATE for "shoot--i749592--it-worker-f73ns-z1-56778-ws5qx"
```

So we think that the update call is returning the updated Machine object.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -497,6 +492,8 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 		case codes.Uninitialized:
 			uninitializedMachine = true
 			klog.Infof("VM instance associated with machine %s was created but not initialized.", machine.Name)
+			nodeName = getMachineStatusResponse.NodeName
```
unmarshall, line 495: We discussed that we will go with this even though this is quite dirty. We add a comment to improve this further.
The easier way to do this would be to add a bool pointer indicating the status of initialization status. If it is nil, then no initialization needs to be done (ignored). If it is false/true then we act upon it. Essentially do not return an error as initialization becomes part of the VM status.

unmarshall (approved):
/lgtm

PR #947, opened 2024-10-21
Fix context declaration for vol detachment and reattachment

### LINKED ISSUE #945: MCM doesn't wait for volumes to reattach

/area robustness
/kind bug
/priority 2

**What happened**:

After waiting for volumes to detach from a terminating node, MCM waits for the volumes to reattach to another node.
However, it immediately runs into a "timeout" instead of waiting (it's rather a context cancellation, but the logs say "timeout"):

```
I1010 16:22:37.863207       1 drain.go:752] Pod + volume detachment from Node garden for Pod prometheus-aggregate-0/shoot--garden--s-eu01-000-worker-z1-6f5b7-48qxs and took 10.193009281s
I1010 16:22:37.863483       1 drain.go:908] Waiting for following volumes to reattach: [pv-shoot--garden--s-eu01-000-5ca2886d-85d9-46d6-8ef3-175d194aaa03]
I1010 16:22:37.863751       1 drain.go:931] VolumeAttachment event received for PV: pv-shoot--garden--s-eu01-000-5ca2886d-85d9-46d6-8ef3-175d194aaa03
W1010 16:22:37.865246       1 drain.go:926] Timeout occurred while waiting for PVs [pv-shoot--garden--s-eu01-000-5ca2886d-85d9-46d6-8ef3-175d194aaa03] to reattach to a different node
W1010 16:22:37.865326       1 drain.go:771] Timeout occurred for following volumes to reattach: [pv-shoot--garden--s-eu01-000-5ca2886d-85d9-46d6-8ef3-175d194aaa03]
```

The effect is that MCM no longer waits for volumes to reattach and continues with the next pod right away.

MCM should wait for volumes to reattach to other nodes before evicting the next pod.

Delete a machine with persistent volumes attached to it and observe the logs.

The cause for this bug is that the context created for `waitForReattach` inherits from `ctx` ([ref](https://github.com/gardener/machine-controller-manager/blob/97cd75245dacc4e40327ded2f28a37a4a3f42cf1/pkg/util/provider/drain/drain.go#L760-L762)), which is the context created for `waitForDetach`. However, `ctx` is cancelled immediately after `waitForDetach` returns ([ref](https://github.com/gardener/machine-controller-manager/blob/97cd75245dacc4e40327ded2f28a37a4a3f42cf1/pkg/util/provider/drain/drain.go#L737-L739)), so the `ctx` for `waitForReattach` is cancelled right from the start.

It seems this was broken by https://github.com/gardener/machine-controller-manager/pull/920 (cc @sssash18).

**Environment**:

- Cloud provider or hardware configuration: all providers are affected, but reproduced with `europe-docker.pkg.dev/gardener-project/releases/gardener/machine-controller-manager-provider-openstack:v0.19.0`

rishabh-11:
Fixes a bug where volume reattachment context is declared using volume detachment context as the parent context. This PR also removes the support for `v1beta1` PDB as it was moved to `v1` from k8s `v1.21`

Fixes #945 

**Release note**:

```bugfix operator
Fixed a bug in drain logic where volume reattachment context declaration used incorrect parent context
```
```other operator
Removed support for v1beta1 PDBs
```


unmarshall (changes requested):
Looks good. Maybe you can also remove the support for v1beta1 PDB. Just check if g/g also does not support this version and only supports v1. This was introduced for a short period of time as we had to support both versions allowing consumers to transition from v1beta1 to v1 for PDB.

unmarshall (approved):
/lgtm

PR #956, opened 2024-11-21
Fix edge-case where scale-from-zero of MCD is mistaken for a rolling update

### LINKED ISSUE #952: Edge case where CA scale-down is  rejected by MCM since it is mistakenly considered as rolling update

/area  auto-scaling
/kind bug
/priority 3

**What happened**:

(Canary Issue  5862)

CA runs into a circumstance sometimes where it is unable to delete a node because the MCM mistakenly assumes that a rolling update is ongoing as shown by the log below. 

```
{"log":"Error while trying to delete nodes from shoot--hc-dev--dl-fc576-orc-default-z1: MachineDeployment shoot--hc-dev--dl-fc576-orc-default-z1 is under rolling update , cannot reduce replica count","pid":"1","severity":"WARN","source":"static_autoscaler.go:898"}Show context
```
This edge case can occur when CA attempts to scale up a Node Group, gets `ResourceExhausted` for a Node Group, attempts to reduce replica count for the Node Group and runs into the error above. ( backoffs becomes very slow until time-outs are reached)

CA should be able to scale-down the Node Group and deletes nodes without issues.

Create Node Groups with limited capacity and deploy large workload.  Attempt to get `ResourceExhausted` for a Node Group and then  when clearing nodes for the bad Node Group, get the `cannot reduce replica count` error.

Looking at the CA code, we can see that if the machine deployment has a `Progressing` condition set and if its `{status, reason}` value is not `{True, NewMachineSetAvailable}` then it considers that the machine deployment is under rolling update. (Refer https://github.com/gardener/autoscaler/blob/2f7cf881a55092e0448c66feca2ab706e821ffab/cluster-autoscaler/cloudprovider/mcm/mcm_manager.go#L873). 

From the MCM side, the scale-up of machine deployment by CA should be considered a scaling and not a rolling update event. Looking at the MCM code we see the following:-

1. It only considers active machine sets, (See https://github.com/gardener/machine-controller-manager/blob/0a6589d500be784c09c9d8e16159bc5f2f1882db/pkg/controller/deployment_sync.go#L672)
2. For a machine set to be considered active, its spec.Replicas >= 0.  (See https://github.com/gardener/machine-controller-manager/blob/31834e2a9beee5bdc3663cf2cce0d6f63d0c20a4/pkg/controller/controller_utils.go#L863)

In this case, there was only one machine set for the machine deployment scaled by CA and it had zero replicas. Hence, the scaling event by CA was not considered as a scaling event by MCM but rather a rolling update event. The status of the machine deployment was updated by MCM considering the event as a rolling update, and this was reflected in the CA logs as well. We all know that scale down is disabled during a rolling update and hence we see this deadlock happening. 

**Environment**:
NA

thiyyakat:

The PR adds a check to distinguish between a rolling update and a scale-from-zero scenario. Additionally, it also updates the integration tests to verify MCM's behaviour for the scenario.

Fixes #952 

Integration tests with mcm-provider-gcp passed successfully.

**Release note**:

```bugfix user
MCM recognizes scaling-up of `machineDeployment` from 0 to more replicas as a scaling event rather than a rolling update.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -176,10 +176,8 @@ add-license-headers: $(GO_ADD_LICENSE)
 
 .PHONY: sast
 sast: $(GOSEC)
-	@chmod +xw hack/sast.sh
 	@./hack/sast.sh
 
 .PHONY: sast-report
 sast-report:$(GOSEC)
-	@chmod +xw hack/sast.sh
 	@./hack/sast.sh --gosec-report true
--- a/pkg/apis/machine/v1alpha1/machinedeployment_types.go
+++ b/pkg/apis/machine/v1alpha1/machinedeployment_types.go
@@ -132,10 +132,9 @@ type RollingUpdateMachineDeployment struct {
 	// Value can be an absolute number (ex: 5) or a percentage of desired machines (ex: 10%).
 	// Absolute number is calculated from percentage by rounding down.
 	// This can not be 0 if MaxSurge is 0.
-	// By default, a fixed value of 1 is used.
-	// Example: when this is set to 30%, the old MC can be scaled down to 70% of desired machines
-	// immediately when the rolling update starts. Once new machines are ready, old MC
-	// can be scaled down further, followed by scaling up the new MC, ensuring
+	// Example: when this is set to 30%, the old machine set can be scaled down to 70% of desired machines
+	// immediately when the rolling update starts. Once new machines are ready, old machine set
+	// can be scaled down further, followed by scaling up the new machine set, ensuring
 	// that the total number of machines available at all times during the update is at
 	// least 70% of desired machines.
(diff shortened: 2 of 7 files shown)


gardener-robot-ci-1:
Thank you @thiyyakat for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

elankath (commented):
I would request to add a small unit test around `FilterActiveMachineSets` which tests this edge-case

unmarshall (commented):
A nit else it looks good

unmarshall (approved):
/lgtm

Issue #960, opened 2024-12-27
MCM integration tests failing due to missing label on Machine object


thiyyakat:
**How to categorize this issue?**

/area testing
/kind enhancement
/priority 3

**What happened**:
The `node-agent-authorizer` authenticating webhook in gardener-resource-manager introduced by PR https://github.com/gardener/gardener/pull/10535, specifically https://github.com/gardener/gardener/blob/3d9c6bbc9557f36ad346efae7073b7184745420f/pkg/resourcemanager/webhook/nodeagentauthorizer/authorizer.go#L204,  requires label `worker.gardener.cloud/gardener-node-agent-secret-name` to be set correctly on the `machine` object. However, the `machine` objects created as a part of the integration tests for MCM providers, do not have this label, preventing the node from being created and causing tests to fail across all providers.

PRs as a part of the fix:

- [x] MCM  > https://github.com/gardener/machine-controller-manager/pull/963
- [x] mcm-provider-aws > https://github.com/gardener/machine-controller-manager-provider-aws/pull/184
- [x] mcm-provider-azure > https://github.com/gardener/machine-controller-manager-provider-azure/pull/171
- [x] mcm-provider-gcp > https://github.com/gardener/machine-controller-manager-provider-gcp/pull/139
- [x] mcm-provider-openstack > NA
- [x] mcm-provider-alicloud > https://github.com/gardener/machine-controller-manager-provider-alicloud/pull/86

aaronfern:
Closing as completed for all providers
/close

PR #963, opened 2025-01-22
Change integration test to add `gardener-node-agent` `secret` name to `machine`s and `MCD`s.

aaronfern:
This PR updated the IT framework to add the `gardener-node-agent` `secret` name to the deployed machine object

Fixes partially #960

**Release note**:

```other operator
Update IT framework to add gardener-node-agent secret to deployed machines
```

### DIFF
--- a/pkg/test/integration/common/helpers/machine_resources.go
+++ b/pkg/test/integration/common/helpers/machine_resources.go
@@ -15,8 +15,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+const gnaSecretNameLabelKey = "worker.gardener.cloud/gardener-node-agent-secret-name"
+
 // CreateMachine creates a test-machine using machineclass "test-mc"
-func (c *Cluster) CreateMachine(namespace string) error {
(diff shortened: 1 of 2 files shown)


pkg/test/integration/common/helpers/machine_resources.go
```diff
@@ -74,6 +83,13 @@ func (c *Cluster) CreateMachineDeployment(namespace string, replicas int32) erro
 								Kind: "MachineClass",
 								Name: "test-mc-v1",
 							},
+							NodeTemplateSpec: v1alpha1.NodeTemplateSpec{
+								ObjectMeta: metav1.ObjectMeta{
+									Labels: map[string]string{
+										gnaSecretNameLabelKey: gnaSecretName,
```
elankath, line 89: hopefully, empty `gnaSecretName` causes no issues in the general case.

elankath (approved):
thanks for fix.

pkg/test/integration/common/helpers/machine_resources.go
aaronfern: No, it won't. The general case will ignore this label altogether

PR #964, opened 2025-01-29
Introduce machine termination queue to handle deletions

### LINKED ISSUE #943: Scalability issues with MCM

/area control-plane
/area scalability
/kind enhancement
/priority 1

**What happened**:
In the recent live update, we saw that the worker pools for our seeds were replaced with new ones concurrently. This caused the deletion of old machine-level objects (machine deployments, machine sets, machine classes and machines) and subsequent creation of new machine objects. All the old machines across worker pools went to `Terminating` state simultaneously. 

Since we have a shared queue and around 50 workers picking up items from this queue, this caused massive throttling due to certain potentially long-running operations like the draining of nodes. Because the workers were blocked in the drain operation, the create requests were getting stuck in the queue with no worker available to process these requests. 

The drain timeout was 2 hrs but it took more than 4 hrs because of https://github.com/gardener/machine-controller-manager/issues/785 which is part of `0.54.0` version of MCM which had not reached live landscape with the corresponding mcm-provider release. 

In the recent live update:-
1. 5-6 new worker pools were introduced to replace the existing 2 worker pools.
2. We observed around >100 or so machines in the `Terminating` state. 
3. Around 50 or so create machine requests were stuck in the queue.
4. For around 4 hrs, due to long drain times and throttling, the machines were stuck in drain.

MCM should scale much beyond handling 100 concurrent deletion/creation requests.

takoverflow:
Adds a new queue to handle machine objects scheduled for
deletion to unblock machine creation/update requests.

Fixes #943

**Release note**:

```other operator

```


takoverflow:
# Tested scenario
## Scale-up/Scale-down of a Machine Deployment
By causing a scale-down in a deployment having a large replica count, numerous machines simultaneously go to the terminating state where they clog up the machine queue preventing any new machine creations (which is simulated by causing a scale-up in a different Machine Deployment)

```
kubectl get mcd
NAME                              READY   DESIRED   UP-TO-DATE   AVAILABLE   AGE
shoot--i752152--scale-test-a-z1                                              0s
shoot--i752152--scale-test-b-z1           1                                  0s
```

1. Increase the replica count of one of the machine deployments (`a`).
```
kubectl scale mcd shoot--i752152--scale-test-a-z1 --replicas=100
machinedeployment.machine.sapcloud.io/shoot--i752152--scale-test-a-z1 scaled
```

2. Check that all the machines are running, then scale-down the Machine Deployment (`a`) to simulate deletion of numerous machines at once.
```
kubectl get machine | grep Running | wc -l
101
kubectl scale mcd shoot--i752152--scale-test-a-z1 --replicas=0
machinedeployment.machine.sapcloud.io/shoot--i752152--scale-test-a-z1 scaled
```

3. Scale up the other deployment (`b`) to have pending machine creation requests
```
kubectl scale mcd shoot--i752152--scale-test-b-z1 --replicas=100
machinedeployment.machine.sapcloud.io/shoot--i752152--scale-test-b-z1 scaled
```

4. Wait for some time and then check status of the number of machines that are running. The deleted machine requests were artifically delayed in the virtual provider to simulate real time-consuming deletion process.
```
sleep 180
kubectl get machine | grep 'a-z1.*Terminating' | wc -l
100
kubectl get machine | grep 'b-z1.*Running' | wc -l
1
```

Due to the terminating machines hogging all the workers in the machine queue, creation requests can't go through. The same scenario when replicated with the new `machineTerminationQueue` allows for separating the time-consuming deletion flow into a different queue thus allowing for creating requests to not have to wait for longer durations.
```
sleep 180
kubectl get machine | grep 'a-z1.*Terminating' | wc -l
100
kubectl get machine | grep 'b-z1.*Running' | wc -l
100
```

takoverflow:
# Stress Test
## Delete 4 machine deployments and Scale-up a new one
By causing a scale-down in a deployment having a large replica count, numerous machines simultaneously go to the terminating state where they clog up the machine queue preventing any new machine creations (which is simulated by causing a scale-up in a different Machine Deployment)

```
+ kubectl get mcd
NAME                              READY   DESIRED   UP-TO-DATE   AVAILABLE   AGE
shoot--i752152--scale-test-a-z1                                              1s
shoot--i752152--scale-test-b-z1                                              1s
shoot--i752152--scale-test-c-z1                                              1s
shoot--i752152--scale-test-d-z1           1                                  1s
shoot--i752152--scale-test-e-z1                                              0s
```

1. Increase the replica count of four of the machine deployments (`a`, `b`, `c` and `d`).
```
+ kubectl scale mcd shoot--i752152--scale-test-a-z1 --replicas=100
machinedeployment.machine.sapcloud.io/shoot--i752152--scale-test-a-z1 scaled
+ kubectl scale mcd shoot--i752152--scale-test-b-z1 --replicas=100
machinedeployment.machine.sapcloud.io/shoot--i752152--scale-test-b-z1 scaled
+ kubectl scale mcd shoot--i752152--scale-test-c-z1 --replicas=100
machinedeployment.machine.sapcloud.io/shoot--i752152--scale-test-c-z1 scaled
+ kubectl scale mcd shoot--i752152--scale-test-d-z1 --replicas=100
machinedeployment.machine.sapcloud.io/shoot--i752152--scale-test-d-z1 scaled
```

2. Schedule a pod on each machine, wait for some time and then delete all four Machine Deployments.
```
+ kubectl delete mcd shoot--i752152--scale-test-a-z1
machinedeployment.machine.sapcloud.io "shoot--i752152--scale-test-a-z1" deleted
+ kubectl delete mcd shoot--i752152--scale-test-b-z1
machinedeployment.machine.sapcloud.io "shoot--i752152--scale-test-b-z1" deleted
+ kubectl delete mcd shoot--i752152--scale-test-c-z1
machinedeployment.machine.sapcloud.io "shoot--i752152--scale-test-c-z1" deleted
+ kubectl delete mcd shoot--i752152--scale-test-d-z1
machinedeployment.machine.sapcloud.io "shoot--i752152--scale-test-d-z1" deleted
```

3. Scale up the remaining deployment (`e`) to have pending machine creation requests with a bit of a wait to allow the deletion requests to all process so that the creation requests can't seep in between.
```
+ sleep 120
+ kubectl scale mcd shoot--i752152--scale-test-e-z1 --replicas=100
machinedeployment.machine.sapcloud.io/shoot--i752152--scale-test-e-z1 scaled
```

4. Wait for some time and then check status of the number of machines that are running. The deleted machine requests were artificially delayed (by `10m` each) in the virtual provider to simulate real time-consuming deletion process.
```
+ sleep 180
+ kubectl get machine | grep 'Terminating' | wc -l
400
+ kubectl get machine | grep 'e-z1.*Running' | wc -l
0
```

Looking at the logs for the timestamps when the machines were scheduled for deletion and when the first machine from the newly scaled deployment(`e`) started running:
```
+ grep -Pi 'machine "shoot--i752152--scale-test-a-z1-6f964-2z6hs" status updated to terminating' old.log
I0204 21:45:19.924412   35563 machine_util.go:936] Machine "shoot--i752152--scale-test-a-z1-6f964-2z6hs" status updated to terminating

+ grep -Pi 'Start for "shoot--i752152--scale-test-e-z1-d5fc5-crrfj" with phase:"Running"' old.log | head -n 1
I0204 23:23:44.720681   35563 machine.go:116] reconcileClusterMachine: Start for "shoot--i752152--scale-test-e-z1-d5fc5-crrfj" with phase:"Running", description:"Machine shoot--i752152--scale-test-e-z1-d5fc5-crrfj successfully joined the cluster"
```
So the first machine that was scheduled for deletion was at `21:45:19` and the first machine that went to "Running" state for the newly scaled deployment was at `23:23:44`. Due to the terminating machines hogging all the workers in the machine queue, creation requests can't go through and it takes almost `1h40m` for successful creation of new machines.

### After the change
The same scenario when replicated with the new `machineTerminationQueue` allows for separating the time-consuming deletion flow into a different queue thus allowing for creation requests to not have to wait for longer duration.
```
+ sleep 180
+ kubectl get machine | grep 'Terminating' | wc -l
400
+ kubectl get machine | grep 'e-z1.*Running' | wc -l
100
```

Again, looking at the timestamps
```
+ grep -Pi 'shoot--i752152--scale-test-b-z1-64874-b6hjs.*reason: handling terminating machine object UPDATE event' new.log
I0204 23:46:33.287374   49438 machine.go:92] Adding machine object to termination queue "shoot--i752152--scale-test/shoot--i752152--scale-test-b-z1-64874-b6hjs", reason: handling terminating machine object UPDATE event

+ grep -Pi 'Start for "shoot--i752152--scale-test-e-z1-d5fc5-v8dhj" with phase:"Running"' new.log | head -n 1
I0204 23:51:08.974343   49438 machine.go:140] reconcileClusterMachine: Start for "shoot--i752152--scale-test-e-z1-d5fc5-v8dhj" with phase:"Running", description:"Machine shoot--i752152--scale-test-e-z1-d5fc5-v8dhj successfully joined the cluster"
```

The first machine scheduled for deletion was at `23:46:33` and the first "Running" machine from the new deployment is at `23:51:08` that's a difference of `~6m` and  removing `3m/180s` from the `sleep 180` invocation earlier, it takes close to `3m` for the new creation requests to go through.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -101,12 +119,18 @@ func (c *controller) reconcileClusterMachineKey(key string) error {
 		return err
 	}
 
+	if machine.DeletionTimestamp != nil {
+		klog.Errorf("should be in machine termination queue %s", machine.Name)
+		return nil
```
aaronfern, line 124: Why not return an error here?
A terminating machine in the machineQueue should be eventually removed and added to the proper queue

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -193,6 +205,57 @@ func (c *controller) reconcileClusterMachine(ctx context.Context, machine *v1alp
...
+			},
+		)
+
+		if err != nil {
+			c.enqueueMachineTerminationAfter(machine, time.Duration(retryPeriod), err.Error())
+			return err
+		}
+	}
+
+	c.enqueueMachineTerminationAfter(machine, time.Duration(machineutils.LongRetry), "periodic reconcile")
```
aaronfern, line 254: Since you have to have a condition above to check for the `DeletionTimestamp`, we end up here if `DeletionTimestamp == nil`.
If this is the case then we should not be requeueing the machine back into the `machineTerminationQueue`

pkg/util/provider/machinecontroller/machine.go
takoverflow: In case the reconcile returns an `err`, the worker/machine is re-queued again for `DefaultMaxRetries` for reconciliation, in this case however since its a machine that shouldn't even be in this queue, further re-queues should be prevented. Hence the error is logged and `nil` is returned.

pkg/util/provider/machinecontroller/machine.go
takoverflow: Removed the unnecessary re-queue.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -193,6 +210,55 @@ func (c *controller) reconcileClusterMachine(ctx context.Context, machine *v1alp
...
+	defer klog.V(2).Infof("reconcileClusterMachineTermination: Stop for %q", machine.Name)
+
+	machineClass, secretData, retry, err := c.ValidateMachineClass(ctx, &machine.Spec.Class)
+	if err != nil {
+		klog.Errorf("cannot reconcile machine %s: %s", machine.Name, err)
+		c.enqueueMachineTerminationAfter(machine, time.Duration(retry), err.Error())
+		return err
+	}
+
+	if machine.DeletionTimestamp != nil {
```
aaronfern, line 242: This check is probably redundant now. Consider removing it

takoverflow: Removed the unneeded check

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -35,7 +35,12 @@ SECTION
 Machine controller - Machine add, update, delete watches
 */
 func (c *controller) addMachine(obj interface{}) {
-	c.enqueueMachine(obj, "handling machine obj ADD event")
+	machine := obj.(*v1alpha1.Machine)
```
lenkite, line 38: This can unfortunately panic in those cases where obj is tomb-stoned. See `deleteMachineDeployment`. This is a common mistake in the delete callback. (I also made the same in the past)

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -83,6 +92,20 @@ func (c *controller) enqueueMachineAfter(obj interface{}, after time.Duration, r
 	}
 }
 
+func (c *controller) enqueueMachineTermination(obj interface{}, reason string) {
+	if key, ok := c.getKeyForObj(obj); ok {
```
lenkite, line 96: callers should ensure that tombstone handling is done correctly otherwise this will fail.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -193,6 +210,53 @@ func (c *controller) reconcileClusterMachine(ctx context.Context, machine *v1alp
...
+	retryPeriod, err := c.triggerDeletionFlow(
+		ctx,
+		&driver.DeleteMachineRequest{
+			Machine:      machine,
+			MachineClass: machineClass,
+			Secret:       &corev1.Secret{Data: secretData},
+		},
+	)
+
+	if err != nil {
```
unmarshall, line 252: Is it possible that `triggerDeletionFlow` returns a nil error but a non-zero retryPeriod? The reason is that you re-queue only when error is non-nil. In master code line no such check is done (https://github.com/gardener/machine-controller-manager/blob/master/pkg/util/provider/machinecontroller/machine.go#L149-L156) which  means that if there is a non-nil error and a non-zero retry period then an re-queue will happen in the master code line (see https://github.com/gardener/machine-controller-manager/blob/master/pkg/util/provider/machinecontroller/machine.go#L104-L110)

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -35,7 +35,12 @@ SECTION
 Machine controller - Machine add, update, delete watches
 */
 func (c *controller) addMachine(obj interface{}) {
-	c.enqueueMachine(obj, "handling machine obj ADD event")
+	machine := obj.(*v1alpha1.Machine)
+	if machine.DeletionTimestamp != nil {
```
unmarshall, line 39: A bit confused, why should we check for deletion timestamp in `addMachine`?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -35,7 +35,12 @@ SECTION
...
+		c.enqueueMachineTermination(obj, "handling terminating machine object ADD event")
+	} else {
+		c.enqueueMachine(obj, "handling machine obj ADD event")
+	}
 }
 
 func (c *controller) updateMachine(oldObj, newObj interface{}) {
```
unmarshall, line 46: A bit confused, why should we check for deletion timestamp in `updateMachine`?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -468,26 +532,7 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 		case codes.Unknown, codes.DeadlineExceeded, codes.Aborted, codes.Unavailable:
 			// GetMachineStatus() returned with one of the above error codes.
 			// Retry operation.
-			updateRetryPeriod, updateErr := c.machineStatusUpdate(
```
unmarshall, line 471: if you wish to use use `machineCreateErrorHandler` for `codes.Unknown`, `codes.DeadlineExceeded`, `codes.Aborted`, `codes.Unavailable` and `default` then is it not possible to remove the case for `codes.Unknown, codes.DeadlineExceeded, codes.Aborted, codes.Unavailable:` and just have `case default` instead?

unmarshall (changes requested):

pkg/util/provider/machinecontroller/machine.go
takoverflow: Yes, when machine deletion in `triggerDeletionFlow` succeds it returns a non-zero `retryPeriod` with `nil` error. [Ref](https://github.com/gardener/machine-controller-manager/blob/73405d0cb2ed327c919e11d6f9cb247b24b55715/pkg/util/provider/machinecontroller/machine.go#L707).
```go
	klog.V(2).Infof("Machine %q with providerID %q and nodeName %q deleted successfully", machine.Name, getProviderID(machine), getNodeName(machine))
	return machineutils.LongRetry, nil
```
Now since deletion is finished, `deleteMachine` eventHandler will be triggered anyways and one final reconciliation will happen so there's no need for an additional delayed re-queue. This is something that I observed in the logs when running the IT on master. Relevant snippet from the same:
```
I0206 17:05:56.868798   23474 machine.go:706] Machine "test-machine" with providerID "aws:///eu-west-1/i-03bc8c20876d09c88" and nodeName "ip-10-180-1-149.eu-west-1.compute.internal" deleted successfully

I0206 17:05:56.868804   23474 machine.go:149] reconcileClusterMachine: Stop for "test-machine"

I0206 17:05:56.868810   23474 machine.go:81] Adding machine object to queue "shoot--mcm-ci--aws-ci-it/test-machine" after 10m0s, reason: periodic reconcile

I0206 17:05:56.868833   23474 machine.go:74] Adding machine object to queue "shoot--mcm-ci--aws-ci-it/test-machine", reason: handling machine object DELETE event
```

pkg/util/provider/machinecontroller/machine.go
takoverflow: Please check [above](https://github.com/gardener/machine-controller-manager/pull/964/files#r1971105876).

pkg/util/provider/machinecontroller/machine.go
unmarshall: Ok this is fair. If the `triggerDeletionFlow` does not have an error i do not know why today there is a requeue with a `LongRetry`. Seems kind of useless and only occupies queue capacity.
@lenkite do you see any side effect of removing this requeue?

pkg/util/provider/machinecontroller/machine.go
unmarshall: Ok if a deletion timestamp is placed on a resource, then it is effectively treated as an `Update event` and not a `Delete event` which is called after the resource in question has been deleted.  This was also confirmed [here](https://github.com/kubernetes/client-go/issues/606#issuecomment-495046412).

pkg/util/provider/machinecontroller/machine.go
unmarshall: Can you double check why an `Add` event will be generated when `DeletionTimestamp` is added to a resource. It should ideally be an Update event, right?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -35,7 +36,12 @@ SECTION
 Machine controller - Machine add, update, delete watches
 */
 func (c *controller) addMachine(obj interface{}) {
-	c.enqueueMachine(obj, "handling machine obj ADD event")
+	machine := obj.(*v1alpha1.Machine)
```
unmarshall, line 39: When casting one should always check if the casting was done successfully.

pkg/util/provider/machinecontroller/machine.go
takoverflow: As discussed offline, `Add` event would be generated when re-populating the queue after process restart. [Refer](https://github.com/kubernetes/kubernetes/blob/60d0d67fefd62c05275b75bbaa5391754f1bca5e/pkg/controller/job/job_controller.go#L301).

elankath: @takoverflow  thanks for explanation. Please add this as a comment over the code in the deletion timestmap checking, so that it is clear to later maintainer.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -83,6 +109,27 @@ func (c *controller) enqueueMachineAfter(obj interface{}, after time.Duration, r
 	}
 }
 
+func (c *controller) enqueueMachineTermination(machine *v1alpha1.Machine, reason string) {
+
+	if key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(machine); err != nil {
+		utilruntime.HandleError(fmt.Errorf("Couldn't get key for object %#v: %v", machine, err))
+		return
+	} else {
```
elankath, line 117: No need for else block as it is the last option in the method.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -101,12 +148,18 @@ func (c *controller) reconcileClusterMachineKey(key string) error {
 		return err
 	}
 
+	if machine.DeletionTimestamp != nil {
+		klog.Errorf("should be in machine termination queue %s", machine.Name)
```
elankath, line 152: Could you change the error log to "Machine %q should be in machine termination queue." ?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -193,6 +234,53 @@ func (c *controller) reconcileClusterMachine(ctx context.Context, machine *v1alp
...
+		return err
+	}
+
+	klog.V(2).Infof("reconcileClusterMachineTermination: Start for %q with phase:%q, description:%q",
+		machine.Name, machine.Status.CurrentStatus.Phase, machine.Status.LastOperation.Description)
+	defer klog.V(2).Infof("reconcileClusterMachineTermination: Stop for %q", machine.Name)
+
+	machineClass, secretData, retry, err := c.ValidateMachineClass(ctx, &machine.Spec.Class)
+	if err != nil {
+		klog.Errorf("cannot reconcile machine %s: %s", machine.Name, err)
```
elankath, line 261: Use %q for machine names to maintain consistency with the new logs you have introduced

pkg/util/provider/machinecontroller/machine.go
takoverflow: The key value is computed as part of the `if` statement, which is then utilized in the `else` clause. Either I can extract the key and err out of the conditional to remove the else clause or keep it as is, whichever you feel is more idiomatic.

elankath: My bad, ignore - I gave the comment for wrong existing method.

pkg/util/provider/machinecontroller/machine.go
elankath: The only issue is that Go cache event callback is not 100% reliable so deleteMachine callback may be missed rarely.

elankath (changes requested):
looks good with minor safety re-queue

pkg/util/provider/machinecontroller/machine.go
takoverflow: We don't actually do anything post machine-deletion that might require another re-queue but just to be safe I've added it back.

unmarshall (approved):

PR #966, opened 2025-02-17
Add a `nil` check before trying to adding mc annotation

aaronfern:
This PR fixes a bug where a `nil` check was not performed before adding an annotation to a machine. 
Machines are not always guaranteed to have annotations, hence a `nil` check is required here

Fixes #

**Release note**:

```other operator

```


elankath (approved):
thanks.

PR #967, opened 2025-02-17
[rel-v0.56] Add a nil check before trying to adding mc annotation (#966)

aaronfern:
This PR cherry picks the following PR: https://github.com/gardener/machine-controller-manager/pull/966

Fixes #

**Release note**:

```bugfix operator
A bug was fixed where MCM panics when trying to add an annotation to a nil map
```


elankath (approved):
thanks /lgtm

Issue #944, opened 2024-10-07
[GEP-31] Support `InPlaceUpdate` deployment strategy


acumino:
**How to categorize this issue?**

/area open-source
/kind enhancement
/priority 2

**What would you like to be added**:
As of now, machine deployment provides two strategies called `Recreate` and `RollingUpdate`, we would like to have new strategy called InPlaceUpdate, with two flavors `Auto` and `Manual` which should work as follows - 

## InPlaceUpdate: 
The following step will be performed for in-place update on the machine:
  - Mark the machine as unschedulable.
  - Drain the node.
  - Mark the machine as ready for an update using a set of labels/annotations.
  - Wait for an annotation/label indicating that the machine has been updated.
  -  Once the update is detected move the machine to the new machine set.
  -  Uncordon the node successfully updated and moved to the new machine set.

### Auto:
Machines will be selected based on maxUnavailable and will go through the steps mentioned above.

### Manual:
Machines/nodes will be selected for updates based on certain labels, which the user has control over. This strategy gives users the flexibility to specify which machines/nodes should be updated by applying predefined labels. Once a machine is labeled, it will follow the same steps as the InPlace update. 

### Tasks
- **Machine Controller Manager** 
  - [x] API changes
    - [x] `MachineDeployment` https://github.com/gardener/machine-controller-manager/pull/962
    - [x] Introduce Label Constants and `MachineConfiguration` options https://github.com/gardener/machine-controller-manager/pull/970
  - [x] Introduce a new rolling strategy
    - [x] InPlaceUpdate
      - [x] Auto https://github.com/gardener/machine-controller-manager/pull/973
      - [x] Manual https://github.com/gardener/machine-controller-manager/pull/987
- **MCM-providers**: Vendor the new MCM and adapt changes to support the InPlace update.
  - [x] https://github.com/gardener/machine-controller-manager-provider-alicloud/pull/100. **Released with** [v0.13.0](https://github.com/gardener/machine-controller-manager-provider-alicloud/releases/tag/v0.13.0)
  - [x] https://github.com/gardener/machine-controller-manager-provider-aws/pull/192 **Released with** [v0.24.0](https://github.com/gardener/machine-controller-manager-provider-aws/releases/tag/v0.24.0)
  - [x] https://github.com/gardener/machine-controller-manager-provider-gcp/pull/153. **Released with** [v0.24.0](https://github.com/gardener/machine-controller-manager-provider-gcp/releases/tag/v0.24.0)
[...]

PR #973, opened 2025-03-11
[GEP-31] Controller changes to Support `InPlaceUpdate` `Auto` strategy

acumino:
This PR introduces the following changes -
- Introduces new node condition, '`LastOperations` state, and `CurrentStatus` `Phase` for in-place update. 
- Controller change for support `Auto` InPlaceUpdate strategy.
- Drain logic for in-place updates.
- Machine health check adaptation for in-place update.
- New predicate for machine reconciliation so if a node is labeled with any of the in-place update labels, the machine should get reconciled.

Part of https://github.com/gardener/gardener/issues/10219

/invite @aaronfern @elankath @unmarshall

**Release note**:

```other operator
NONE
```


pkg/controller/deployment_util.go
```diff
@@ -731,6 +747,31 @@ func GetNewMachineSet(ctx context.Context, deployment *v1alpha1.MachineDeploymen
 	return FindNewMachineSet(deployment, rsList), nil
 }
 
+func filterMachinesWithUpdateSuccessfulLabel(machines []*v1alpha1.Machine) []*v1alpha1.Machine {
+	machinesWithUpdateSuccessfulLabel := make([]*v1alpha1.Machine, 0)
+	for _, machine := range machines {
+		if labelValue, ok := machine.Labels[v1alpha1.LabelKeyNodeUpdateResult]; ok && labelValue == v1alpha1.LabelValueNodeUpdateSuccessful {
+			cond := getMachineCondition(machine, v1alpha1.NodeInPlaceUpdate)
+			// only consider machines with the update successful condition
+			if cond != nil && cond.Reason == v1alpha1.UpdateSuccessful {
```
aaronfern, line 756: Please check only labels as they should represent the definite case that machines are successfully updated

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,559 @@
...
+}
+
+func (dc *controller) syncMachineSets(ctx context.Context, oldISs []*v1alpha1.MachineSet, newIS *v1alpha1.MachineSet, deployment *v1alpha1.MachineDeployment) error {
+	machines, err := dc.machineLister.List(labels.SelectorFromSet(newIS.Spec.Selector.MatchLabels))
+	if err != nil {
+		return err
+	}
+
+	machinesWithUpdateSuccessfulLabel := filterMachinesWithUpdateSuccessfulLabel(machines)
+	klog.V(3).Infof("machine with update successful label in new machine set %v", len(machinesWithUpdateSuccessfulLabel))
```
aaronfern, line 140: If your intention is to log number of machines then you should use `%d`
Else if you wanted to log machine names, then `%v` is fine

Also, please supply the exact label name with `%q`

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,559 @@
...
+
+	if len(machines) > int(newIS.Spec.Replicas) && len(machinesWithUpdateSuccessfulLabel) > 0 {
+		// scale up the new machine set to the number of machines with the update successful label.
+		scaleUpBy := min(len(machinesWithUpdateSuccessfulLabel), len(machines)-int(newIS.Spec.Replicas))
```
aaronfern, line 144: Why do you need to consider the min here? Would `len(machines)-int(newIS.Spec.Replicas)` not suffice?

aaronfern (commented):
Just a few comments from out initial look. PTAL
-- MCM team

pkg/util/labels/labels.go
```diff
@@ -127,3 +130,41 @@ func AddLabelToSelector(selector *metav1.LabelSelector, labelKey, labelValue str
 func SelectorHasLabel(selector *metav1.LabelSelector, labelKey string) bool {
 	return len(selector.MatchLabels[labelKey]) > 0
 }
+
+// GetFormatedLabels retun labels in `json` format.
+func GetFormatedLabels(labels map[string]string) string {
```
aaronfern, line 135: Please use the json standard library for this

aaronfern (commented):
Some more comments. PTAL
Thanks, MCM Team

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,559 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
+	"k8s.io/utils/integer"
+)
+
+// rolloutAutoInPlace implements the logic for rolling  a machine set without replacing it.
+func (dc *controller) rolloutAutoInPlace(ctx context.Context, d *v1alpha1.MachineDeployment, isList []*v1alpha1.MachineSet, machineMap map[types.UID]*v1alpha1.MachineList) error {
```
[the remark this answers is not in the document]
ary1992: These namings are taken from the current MCM code base. Let's not change the namings all of a sudden, it will create unnecessary confusion.

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,559 @@
...
+			_, err = dc.targetCoreClient.CoreV1().Nodes().Update(ctx, node, metav1.UpdateOptions{})
+			if err != nil {
+				return false, fmt.Errorf("failed to uncordon the node %s: %w", node.Name, err)
+			}
+
+			transferredMachineCount++ // scale down the old machine set.
+			addedNewReplicasCount++   // scale up the new machine set.
+		}
+
+		klog.V(3).Infof("scale down machine set %s, transffered replicas to new machine set %d", is.Name, transferredMachineCount)
```
takoverflow, line 306:
```suggestion
		klog.V(3).Infof("scale down machine set %s, transferred %d replicas to new machine set", is.Name, transferredMachineCount)
```

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,559 @@
...
+	}
+
+	klog.V(3).Infof("allMachinesCount:%d,  minAvailable:%d,  newISUnavailableMachineCount:%d,  oldISsMachineInUpdateProcess:%d", allMachinesCount, minAvailable, newISUnavailableMachineCount, oldISsMachinesUndergoingUpdate)
+
+	maxUpdatePossible := allMachinesCount - minAvailable - newISUnavailableMachineCount - oldISsMachinesUndergoingUpdate
+	if maxUpdatePossible <= 0 {
+		return false, nil
+	}
+
+	// preapre machines from old machine sets to get updated, need to check maxUnavailable to ensure we can select machines for update.
```
takoverflow, line 358:
```suggestion
	// prepare machines from old machine sets for update, need to check maxUnavailable to ensure we can select machines for update.
```

takoverflow (commented):
Some minor cosmetic changes and typo fixes, PTAL.

pkg/controller/deployment_util.go
```diff
@@ -618,15 +619,30 @@ func SetReplicasAnnotations(is *v1alpha1.MachineSet, desiredReplicas, maxReplica
 
 // MaxUnavailable returns the maximum unavailable machines a rolling deployment can take.
 func MaxUnavailable(deployment v1alpha1.MachineDeployment) int32 {
-	if !IsRollingUpdate(&deployment) || (deployment.Spec.Replicas) == 0 {
+	if !IsRollingUpdate(&deployment) && !IsInPlaceUpdate(&deployment) || (deployment.Spec.Replicas) == 0 {
```
ary1992, line 622: Check this condition properly. Condition `!IsRollingUpdate(&deployment)` will be true for `inplace update` while `!IsInPlaceUpdate(&deployment)` will be true for `rolling update`. Both the condition cannot occur at once.

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,559 @@
...
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					klog.Warningf("Node %s not found for machine %s", nodeName, machine.Name)
+					continue
+				}
+				return addedNewReplicasCount > 0, fmt.Errorf("failed to get node %s for machine %s: %w", nodeName, machine.Name, err)
+			}
+
+			cond := getMachineCondition(machine, v1alpha1.NodeInPlaceUpdate)
+			if cond == nil || cond.Reason != v1alpha1.UpdateSuccessful || node.Labels[v1alpha1.LabelKeyNodeUpdateResult] != v1alpha1.LabelValueNodeUpdateSuccessful {
```
aaronfern, line 259: Consider creating a func for this which accepts a condition and a label, and returns a bool

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,559 @@
...
+				newIS.GetName(), newIS.GetUID(), labelsutil.GetFormatedLabels(machineNewLabels), machine.UID)
+
+			err = dc.machineControl.PatchMachine(ctx, machine.Namespace, machine.Name, []byte(addControllerPatch))
+			if err != nil {
+				klog.V(3).Infof("failed to transfer the ownership of machine %s to new machine set %s", machine.Name, err)
+				// scale up the new machine set to the already added replicas.
+				scaled, _, err2 := dc.scaleMachineSetAndRecordEvent(ctx, newIS, newIS.Spec.Replicas+addedNewReplicasCount, deployment)
+				klog.V(3).Infof("scale up after failure failed %s", err2)
+				if err2 != nil {
+					return addedNewReplicasCount > 0, err2
```
aaronfern, line 281: when err != nil, the caller does not consider the first boolean return parameter. You can just return `false, err2` here

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,559 @@
...
+			machineNewLabels := MergeStringMaps(MergeWithOverwriteAndFilter(machine.Labels, is.Spec.Selector.MatchLabels, newIS.Spec.Selector.MatchLabels), map[string]string{v1alpha1.LabelKeyNodeUpdateResult: v1alpha1.LabelValueNodeUpdateSuccessful})
+
+			// update the owner reference of the machine to the new machine set and update the labels
+			addControllerPatch := fmt.Sprintf(
+				`{"metadata":{"ownerReferences":[{"apiVersion":"machine.sapcloud.io/v1alpha1","kind":"%s","name":"%s","uid":"%s","controller":true,"blockOwnerDeletion":true}],"labels":{%s},"uid":"%s"}}`,
+				v1alpha1.SchemeGroupVersion.WithKind("MachineSet"),
+				newIS.GetName(), newIS.GetUID(), labelsutil.GetFormatedLabels(machineNewLabels), machine.UID)
+
+			err = dc.machineControl.PatchMachine(ctx, machine.Namespace, machine.Name, []byte(addControllerPatch))
+			if err != nil {
```
aaronfern, line 275: In the case of an error, such complicated error handling is not needed. Simple return, and in the next reconciliation run `SyncMachineSets()` can handle this error

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,559 @@
...
+func (dc *controller) reconcileOldMachineSetsInPlace(ctx context.Context, allISs []*v1alpha1.MachineSet, oldISs []*v1alpha1.MachineSet, newIS *v1alpha1.MachineSet, deployment *v1alpha1.MachineDeployment) (bool, error) {
+	oldMachinesCount := GetReplicaCountForMachineSets(oldISs)
+	if oldMachinesCount == 0 {
+		// Can't scale down further
+		return false, nil
+	}
+
+	// If maxSurge is defined, there will be machines left in the old machine set that do not require an update
+	// because the new machine set already has the required replicas due to the additional machines added as per maxSurge.
+	// In that case we simply scale down the old machine set to zero.
```
aaronfern, line 328:
```suggestion
	// In that case we simply scale down the old machinesets to zero.
```

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,559 @@
...
+	minAvailable := (deployment.Spec.Replicas) - maxUnavailable
+	newISUnavailableMachineCount := (newIS.Spec.Replicas) - newIS.Status.AvailableReplicas
+	oldISsMachinesUndergoingUpdate, err := dc.getMachinesUndergoingUpdate(oldISs)
+	if err != nil {
+		return false, err
+	}
+
+	klog.V(3).Infof("allMachinesCount:%d,  minAvailable:%d,  newISUnavailableMachineCount:%d,  oldISsMachineInUpdateProcess:%d", allMachinesCount, minAvailable, newISUnavailableMachineCount, oldISsMachinesUndergoingUpdate)
+
+	maxUpdatePossible := allMachinesCount - minAvailable - newISUnavailableMachineCount - oldISsMachinesUndergoingUpdate
```
aaronfern, line 353: Please move this computation to `selectMachineForUpdate()` as that is the most semantic place for this computation

aaronfern (commented):
Have some more review comments, please have a look
Thanks, MCM Team

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1045,6 +1299,179 @@ func printLogInitError(s string, err *error, description *string, machine *v1alp
...
+		if node, err = c.nodeLister.Get(nodeName); err == nil {
+			klog.V(3).Infof("(drainNode) For node %q, machine %q, nodeReadyCondition: %s, readOnlyFileSystemCondition: %s", nodeName, machine.Name, nodeReadyCondition, readOnlyFileSystemCondition)
+		} else if apierrors.IsNotFound(err) {
+			klog.Warningf("(drainNode) Node %q for machine %q doesn't exist, so drain will finish instantly", nodeName, machine.Name)
+		}
+
+		if !isConditionEmpty(nodeReadyCondition) && (nodeReadyCondition.Status != v1.ConditionTrue) && (time.Since(nodeReadyCondition.LastTransitionTime.Time) > nodeNotReadyDuration) {
+			message := "Setting forceDeletePods to true for drain as machine is NotReady for over 5min"
+			forceDeletePods = true
+			printLogInitErrorInPlace(message, &err, &description, machine)
```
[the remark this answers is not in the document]
acumino: I will keep it the way, the usage is drainNode and drainNodeForInPlace is quite different.

pkg/controller/machine_safety.go
```diff
@@ -229,8 +229,14 @@ func (c *controller) checkAndFreezeORUnfreezeMachineSets(ctx context.Context) er
 		if len(machineDeployments) >= 1 {
 			machineDeployment := machineDeployments[0]
 			if machineDeployment != nil {
+				var maxSurge *intstrutil.IntOrString
+				if machineDeployment.Spec.Strategy.RollingUpdate != nil {
+					maxSurge = machineDeployment.Spec.Strategy.RollingUpdate.MaxSurge
```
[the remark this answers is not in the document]
acumino: I would do it now, as there is only one use.

pkg/controller/machineset.go
```diff
@@ -377,6 +378,11 @@ func (c *controller) manageReplicas(ctx context.Context, allMachines []*v1alpha1
 			return nil
 		}
 
+		if _, ok := machineSet.Labels[machineutils.LabelKeyMachineSetScaleUpDisabled]; ok {
+			klog.V(2).Infof("MachineSet %s has label %s, and hence not creating new machines for it", machineSet.Name, machineutils.LabelKeyMachineSetScaleUpDisabled)
+			return nil
+		}
```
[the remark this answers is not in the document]
acumino: Only scale-up is disabled i.e. no new machines should be created but if there are extra machines, it should be deleted there before this condition is kept here and not at the top.

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,590 @@
...
+
+// syncMachineSets syncs the machine sets by scaling up the new machine set and scaling down the old machine sets to the required replicas.
+func (dc *controller) syncMachineSets(ctx context.Context, oldMachineSets []*v1alpha1.MachineSet, newMachineSet *v1alpha1.MachineSet, deployment *v1alpha1.MachineDeployment) error {
+	machines, err := dc.machineLister.List(labels.SelectorFromSet(newMachineSet.Spec.Selector.MatchLabels))
+	if err != nil {
+		return err
+	}
+
+	machinesWithUpdateSuccessfulLabel := filterMachinesWithUpdateSuccessfulLabel(machines)
+	klog.V(3).Infof("machine with update successful label in new machine set %d", len(machinesWithUpdateSuccessfulLabel))
```
elankath, line 134: Kindly change to: "Num machinse with %q label in new machine set: %d"

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,590 @@
...
+		}
+	}
+
+	// Sync deployment status
+	return dc.syncRolloutStatus(ctx, allMachineSets, newMachineSet, d)
+}
+
+// syncMachineSets syncs the machine sets by scaling up the new machine set and scaling down the old machine sets to the required replicas.
+func (dc *controller) syncMachineSets(ctx context.Context, oldMachineSets []*v1alpha1.MachineSet, newMachineSet *v1alpha1.MachineSet, deployment *v1alpha1.MachineDeployment) error {
+	machines, err := dc.machineLister.List(labels.SelectorFromSet(newMachineSet.Spec.Selector.MatchLabels))
```
elankath, line 128: `machines`->`newMachines`

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,590 @@
...
+		return err
+	}
+
+	machinesWithUpdateSuccessfulLabel := filterMachinesWithUpdateSuccessfulLabel(machines)
+	klog.V(3).Infof("machine with update successful label in new machine set %d", len(machinesWithUpdateSuccessfulLabel))
+
+	if len(machines) > int(newMachineSet.Spec.Replicas) && len(machinesWithUpdateSuccessfulLabel) > 0 {
+		// scale up the new machine set to the number of machines with the update successful label.
+		// This is to ensure that the machines with moved to the new machine set when ownership is transferred is accounted.
+		scaleUpBy := min(len(machinesWithUpdateSuccessfulLabel), len(machines)-int(newMachineSet.Spec.Replicas))
```
elankath, line 139: min here appears useless since `machinesWithUpdateSuccessfulLabel` is already subset of `newMachines` and you already do checks in the if. I can't think of any reason why it is needed.

aaronfern:
@acumino There are merge conflicts with this PR. Could you please resolve them?

pkg/controller/deployment_inplace.go
acumino: Min will handle the following cases - 
- If `machinesWithUpdateSuccessfulLabel > len(machines)-int(newMachineSet.Spec.Replicas)`, some one created the MCS before an old machine is moved to new MCS
- If `machinesWithUpdateSuccessfulLabel < len(machines)-int(newMachineSet.Spec.Replicas)`  some machine would have got deleted for some another reason, not related to inplace update.

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,590 @@
...
+	}
+
+	klog.V(3).Infof("reconcile new machine set %s", newMachineSet.Name)
+
+	oldMachinesCount := GetReplicaCountForMachineSets(oldMachineSets)
+	if oldMachinesCount == 0 {
+		// if there are no machines in the old machine sets, then there is no need to transfer the ownership of the machines.
+		// scale up the new machine set to the required replicas.
+		scaleBy := deployment.Spec.Replicas - newMachineSet.Spec.Replicas // here scaleBy is always positive
+		scaled, _, err := dc.scaleMachineSetAndRecordEvent(ctx, newMachineSet, newMachineSet.Spec.Replicas+scaleBy, deployment)
```
takoverflow, line 237:
```suggestion
		scaled, _, err := dc.scaleMachineSetAndRecordEvent(ctx, newMachineSet, deployment.Spec.Replicas, deployment)
```
`scaleBy` isn't required here right, consider removing the variable

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,590 @@
...
+		if err != nil {
+			return false, fmt.Errorf("failed to scale up machine set %s: %w", newMachineSet.Name, err)
+		}
+		klog.V(3).Infof("scaled up machine set %s to %d", newMachineSet.Name, newMachineSet.Spec.Replicas+scaleBy)
```
takoverflow, line 241:
```suggestion
		klog.V(3).Infof("scaled up machine set %s to %d", newMachineSet.Name, deployment.Spec.Replicas)
```

pkg/controller/deployment_inplace_test.go
```diff
@@ -0,0 +1,1305 @@
...
+	"k8s.io/utils/ptr"
+
+	machinev1 "github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
+)
+
+var _ = Describe("deployment_inplace", func() {
+	Describe("syncMachineSets", func() {
+		type setup struct {
+			oldMachineSetReplicas                  int32
+			oldMSMachinesMoveToNewMS               int32
```
takoverflow, line 26:
```suggestion
			oldMSMachinesMovedToNewMS              int32
```
Please consider this rename

pkg/controller/deployment_inplace_test.go
```diff
@@ -0,0 +1,1305 @@
...
+
+		machineSets := newMachineSets(
+			2,
+			&machinev1.MachineTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "machineset-0",
+				},
+				Spec: machinev1.MachineSpec{
+					Class: machinev1.ClassSpec{
+						Kind: "OpenStackMachineClass",
```
takoverflow, line 47:
```suggestion
						Kind: "MachineClass",
```
Class `Kind` isn't provider specific anymore afaik, the sample yamls/rolling tests might be specifying like this as a consequence of legacy code, consider renaming all of them.

pkg/controller/deployment_inplace_test.go
```diff
@@ -0,0 +1,1305 @@
...
+				for i := range actualNodes.Items {
+					node := actualNodes.Items[i]
+					Expect(node.Spec.Unschedulable).To(Equal(false))
+					Expect(node.Labels).ToNot(HaveKey(machinev1.LabelKeyNodeCandidateForUpdate))
+					Expect(node.Labels).ToNot(HaveKey(machinev1.LabelKeyNodeSelectedForUpdate))
+					Expect(node.Labels).ToNot(HaveKey(machinev1.LabelKeyNodeUpdateResult))
+				}
+			},
+
+			Entry("no scaling required as there are no extra machines affiliated to new machine set with update successfule label", &data{
```
takoverflow, line 173:
```suggestion
			Entry("no scaling required as there are no extra machines affiliated to new machine set with update successful label", &data{
```

takoverflow (commented):
Some minor nits, PTAL thanks.

pkg/controller/machineset_test.go
```diff
@@ -871,7 +871,180 @@ var _ = Describe("machineset", func() {
 			// replica count is still maintained.
 			Expect(len(afterMachines.Items)).To(Equal(int(testMachineSet.Spec.Replicas)))
 			Expect(err).Should(BeNil())
+		})
+
+		Describe("machine with update-result label", func() {
+			// Testcase: ActiveMachines + MachinesWithUpdateSuccessfulLabel < DesiredMachines
+			// It should create new machines and should not return erros.
```
takoverflow, line 878: Please remove this redundant comment line since its already described below

pkg/controller/machineset_test.go
```diff
@@ -871,7 +871,180 @@ var _ = Describe("machineset", func() {
...
+				activeMachines := []*machinev1.Machine{testActiveMachine1, testActiveMachine3}
+				Err := c.manageReplicas(context.TODO(), activeMachines, testMachineSet)
+				waitForCacheSync(stop, c)
+				machines, _ = c.controlMachineClient.Machines(testNamespace).List(context.TODO(), metav1.ListOptions{})
+				Expect(len(machines.Items)).To(Equal(int(testMachineSet.Spec.Replicas)))
+				Expect(Err).Should(BeNil())
+			})
+
+			// Testcase: ActiveMachines + MachinesWithUpdateSuccessfulLabel = DesiredMachines
+			// It should not return error.
```
takoverflow, line 922: Same as above, please remove this comment line

pkg/controller/machineset_test.go
```diff
@@ -871,7 +871,180 @@ var _ = Describe("machineset", func() {
...
+				Err := c.manageReplicas(context.TODO(), activeMachines, testMachineSet)
+				waitForCacheSync(stop, c)
+				machines, _ = c.controlMachineClient.Machines(testNamespace).List(context.TODO(), metav1.ListOptions{})
+				Expect(len(machines.Items)).To(Equal(int(testMachineSet.Spec.Replicas)))
+				Expect(Err).Should(BeNil())
+			})
+
+			// TestCase: ActiveMachines + MachinesWithUpdateSuccessfulLabel > DesiredMachines
+			// Testcase: It should not return error and delete extra running machine.
+			It("should not create or delete machines and should not return error", func() {
```
takoverflow, line 967: Please update the subject node description in accordance with the expectation and remove the above comment line
```suggestion
			It("should not return error and delete extra running machine", func() {
```

pkg/controller/machineset.go
```diff
@@ -355,6 +360,25 @@ func (c *controller) manageReplicas(ctx context.Context, allMachines []*v1alpha1
...
+	// might incorrectly assume that more machines need to be deleted, as machines without the label will be counted as active.
+	// To address this, we first check if the difference between the current active replicas and the desired replicas is negative.
+	// If it is, we then check if the difference plus the number of machines with the "update successful" label is greater than or equal to 0.
+	// In such cases, we set the difference to 0 because it is unclear whether the machine deployment controller still needs to increase
+	// the replicas of the MachineSet. Once the machine deployment controller removes the "update successful" label, the ambiguity will
+	// be resolved, and normal operations can proceed. This approach ensures that the controller does not create additional machines
+	// if the MachineSet is scaled up before the "update successful" label is removed.
+	if diff < 0 {
+		if diff+len(machinesWithUpdateSuccessfulLabel) >= 0 {
+			diff = 0
```
takoverflow, line 376: Rather than setting the diff to 0, consider early `return nil` here since no work is done for `diff = 0` and we'd be performing redundant checks below. Also helps when looking at the flow.

pkg/controller/machineset.go
```diff
@@ -355,6 +360,25 @@ func (c *controller) manageReplicas(ctx context.Context, allMachines []*v1alpha1
...
+		} else if diff+len(machinesWithUpdateSuccessfulLabel) < 0 {
```
takoverflow, line 377:
```suggestion
		} else {
```

pkg/util/labels/labels_test.go
```diff
@@ -214,3 +214,58 @@ func TestAddLabelToSelector(t *testing.T) {
 		}
 	}
 }
+
+func TestGetFormatedLabels(t *testing.T) {
```
takoverflow, line 218:
```suggestion
func TestGetFormattedLabels(t *testing.T) {
```

pkg/util/labels/labels_test.go
```diff
@@ -214,3 +214,58 @@ func TestAddLabelToSelector(t *testing.T) {
...
+		{
+			labels: map[string]string{},
+			want:   "{}", // Expected empty string for no labels
+		},
+	}
+
+	for _, tc := range cases {
+		got, _ := GetFormattedLabels(tc.labels)
+		if string(got) != tc.want {
+			t.Errorf("GetFormatedLabels(%v) = %q, want %q", tc.labels, got, tc.want)
```
takoverflow, line 239:
```suggestion
			t.Errorf("GetFormattedLabels(%v) = %q, want %q", tc.labels, got, tc.want)
```

pkg/util/labels/labels_test.go
```diff
@@ -214,3 +214,58 @@ func TestAddLabelToSelector(t *testing.T) {
...
+	cases := []struct {
+		labels map[string]string
+		want   string
+	}{
+		{
+			labels: map[string]string{
+				"foo1": "bar1",
+				"foo2": "bar2",
+			},
+			want: "{\"foo1\":\"bar1\",\"foo2\":\"bar2\"}",
```
takoverflow, line 228:
```suggestion
			want: `{"foo1":"bar1","foo2":"bar2"}`,
```

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -415,6 +431,28 @@ func (c *controller) getMachineFromNode(nodeName string) (*v1alpha1.Machine, err
...
+	labelKeys := []string{
+		v1alpha1.LabelKeyNodeCandidateForUpdate,
+		v1alpha1.LabelKeyNodeSelectedForUpdate,
+		v1alpha1.LabelKeyNodeUpdateResult,
+	}
+
+	for _, key := range labelKeys {
+		oldVal, oldOk := oldNode.Labels[key]
+		newVal, newOk := node.Labels[key]
+		if (!oldOk && newOk) || (key == v1alpha1.LabelKeyNodeUpdateResult && oldVal != newVal) {
```
takoverflow, line 448: For manual in-place update, `selected-for-update` label value on the node can potentially be changed by the user right? In that case shouldn't this check for the change in value for that label as well?
```suggestion
		if (!oldOk && newOk) || (key != v1alpha1.LabelKeyNodeCandidateForUpdate && oldVal != newVal) {
```

takoverflow (commented):
Some more changes requested, PTAL thanks

pkg/controller/deployment_util_test.go
```diff
@@ -80,6 +81,191 @@ var _ = Describe("deployment_util", func() {
...
+				if testData.shouldError {
+					Expect(err).To(HaveOccurred())
+				} else {
+					Expect(err).NotTo(HaveOccurred())
+					Expect(result).To(Equal(int32(testData.expected)))
+				}
+			},
+
+			// Case: RollingUpdate strategy, scaling up within max surge
+			Entry("RollingUpdate, scale up within max surge",
```
aaronfern, line 194: Do you want to keep these comments after a description is already present?

pkg/controller/deployment_util_test.go
```diff
@@ -80,6 +81,191 @@ var _ = Describe("deployment_util", func() {
...
+				machinev1.MachineDeploymentStrategy{
+					Type: machinev1.InPlaceUpdateMachineDeploymentStrategyType,
+					InPlaceUpdate: &machinev1.InPlaceUpdateMachineDeployment{
+						OrchestrationType: machinev1.OrchestrationTypeManual,
+					},
+				}, data{replicas: 5, oldISReplicas: 4, newISReplicas: 2, expected: 0, shouldError: false},
+			),
+
+			// Case: InPlaceUpdate strategy with ManualUpdate=false, scaling up within max surge
+			Entry("RollingUpdate, scale up within max surge",
```
aaronfern, line 235: I think you meant `InPlaceUpdate`

pkg/controller/machineset.go
elankath: I believe that remembering the  `transferredMachineNames` and setting them as an annotation or within a new field in the new machineSet when we do the below in `deployment_inplace.go` would solve the data-race here and remove the need for all the extensive diff edge-case handling. Then you don't need `machinesWithUpdateSuccessfulLabel` slice - since it is stored with the machine set atomically when replicas were last adjusted.

```
klog.V(3).Infof("scale down machine set %s, transffered replicas to new machine set %d", machineSet.Name, transferredMachineCount)
// Before this also set transferredMachineNames as annotation or field in machineSet
		_, _, err = dc.scaleMachineSetAndRecordEvent(ctx, machineSet, machineSet.Spec.Replicas-transferredMachineCount, deployment)
	
```

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,587 @@
...
+				return err
+			}
+		}
+		if err := dc.cleanupMachineDeployment(ctx, oldMachineSets, d); err != nil {
+			return err
+		}
+	}
+
+	// Sync deployment status
+	return dc.syncRolloutStatus(ctx, allMachineSets, newMachineSet, d)
```
elankath, line 123: Could you add a "Inplace Update Testing" section in  `docs/development/testing_and_dependencies.md` documenting how to test in-place update as a developer manually ?

pkg/util/provider/machinecontroller/machine.go
takoverflow: The `selected-for-update` label when present denotes selection, its value in inconsequential. So this comment is invalid.

pkg/controller/deployment_inplace.go
acumino: Will do it in a sperate PR as discussed.

pkg/controller/machineset.go
acumino: As discussed, it can be done later, if done.

pkg/controller/deployment_util_test.go
acumino: removed it.

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,586 @@
...
+
+func (dc *controller) reconcileNewMachineSetInPlace(ctx context.Context, oldMachineSets []*v1alpha1.MachineSet, newMachineSet *v1alpha1.MachineSet, deployment *v1alpha1.MachineDeployment) (bool, error) {
+	if newMachineSet.Spec.Replicas == deployment.Spec.Replicas {
+		// Scaling not required.
+		return false, nil
+	}
+
+	if newMachineSet.Spec.Replicas > deployment.Spec.Replicas {
+		// Scale down.
+		scaled, _, err := dc.scaleMachineSetAndRecordEvent(ctx, newMachineSet, (deployment.Spec.Replicas), deployment)
```
aaronfern, line 225:
```suggestion
		scaled, _, err := dc.scaleMachineSetAndRecordEvent(ctx, newMachineSet, deployment.Spec.Replicas, deployment)
```

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,586 @@
...
+
+	for _, machineSet := range oldMachineSets {
+		transferredMachineCount := int32(0)
+		// get the machines for the machine set
+		oldMachines, err := dc.machineLister.List(labels.SelectorFromSet(machineSet.Spec.Selector.MatchLabels))
+		if err != nil {
+			return false, err
+		}
+
+		klog.V(3).Infof("machine in old machine set %s: %d", machineSet.Name, len(oldMachines))
```
aaronfern, line 251:
```suggestion
		klog.V(3).Infof("Found %d machine(s) in old machine set %s", len(oldMachines), machineSet.Name)
```

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,586 @@
...
+				}
+				return false, fmt.Errorf("failed to get node %s for machine %s: %w", nodeName, oldMachine.Name, err)
+			}
+
+			cond := getMachineCondition(oldMachine, v1alpha1.NodeInPlaceUpdate)
+			if isUpdateNotSuccessful(cond, node.Labels) {
+				continue
+			}
+
+			klog.V(3).Infof("transferring machine %s to new machine set %s", oldMachine.Name, newMachineSet.Name)
```
aaronfern, line 273:
```suggestion
			klog.V(3).Infof("Attempting to transfer machine %s to new machine set %s", oldMachine.Name, newMachineSet.Name)
```

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,586 @@
...
+			}
+			// update the owner reference of the machine to the new machine set and update the labels
+			addControllerPatch := fmt.Sprintf(
+				`{"metadata":{"ownerReferences":[{"apiVersion":"machine.sapcloud.io/v1alpha1","kind":"%s","name":"%s","uid":"%s","controller":true,"blockOwnerDeletion":true}],"labels":%s,"uid":"%s"}}`,
+				v1alpha1.SchemeGroupVersion.WithKind("MachineSet"),
+				newMachineSet.GetName(), newMachineSet.GetUID(), string(formattedLabels), oldMachine.UID)
+
+			err = dc.machineControl.PatchMachine(ctx, oldMachine.Namespace, oldMachine.Name, []byte(addControllerPatch))
+			if err != nil {
+				klog.V(3).Infof("failed to transfer the ownership of machine %s to new machine set. Err: %v", oldMachine.Name, err)
```
aaronfern, line 292:
```suggestion
				klog.Errorf("failed to transfer the ownership of machine %s to new machine set. Err: %v", oldMachine.Name, err)
```

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,586 @@
...
+			transferredMachineCount++ // scale down the old machine set.
+			addedNewReplicasCount++   // scale up the new machine set.
+		}
+
+		if transferredMachineCount == 0 {
+			klog.V(3).Infof("no machines transferred from machine set %s", machineSet.Name)
+			continue
+		}
+
+		klog.V(3).Infof("scale down machine set %s, transffered replicas to new machine set %d", machineSet.Name, transferredMachineCount)
```
aaronfern, line 312:
```suggestion
		klog.V(3).Infof("%d replica(s) transferred to new machine set. scaling down machine set %s", transferredMachineCount, machineSet.Name)
```

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,586 @@
...
+		_, _, err = dc.scaleMachineSetAndRecordEvent(ctx, machineSet, machineSet.Spec.Replicas-transferredMachineCount, deployment)
+		if err != nil {
+			klog.V(3).Infof("scale down failed %s", err)
```
aaronfern, line 315:
```suggestion
			klog.Errorf("scale down failed %s", err)
```

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,586 @@
...
+		return err
+	}
+
+	// In this section, we will attempt to scale up the new machine set. Machines with the `node.machine.sapcloud.io/update-successful` label
+	// can transfer their ownership to the new machine set.
+	// It is crucial to ensure that during the ownership transfer, the machine is not deleted,
+	// and the old machine set is not scaled up to recreate the machine.
+	scaledUp, err := dc.reconcileNewMachineSetInPlace(ctx, oldMachineSets, newMachineSet, d)
+	if err != nil {
+		klog.V(3).Infof("failed to reconcile new machine set in place %s", err)
```
aaronfern, line 89:
```suggestion
		klog.Errorf("failed to reconcile new machine set in place %s", err)
```

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,586 @@
...
+	if err != nil {
+		return err
+	}
+	allMachineSets := append(oldMachineSets, newMachineSet)
+
+	if len(oldMachineSets) > 0 && !dc.machineSetsScaledToZero(oldMachineSets) {
+		// Label all the old machine sets to disable the scale up.
+		err := dc.labelMachineSets(ctx, oldMachineSets, map[string]string{machineutils.LabelKeyMachineSetScaleUpDisabled: "true"})
+		if err != nil {
+			klog.Errorf("Failed to add label %s on all machine sets. Error: %s", machineutils.LabelKeyMachineSetScaleUpDisabled, err)
```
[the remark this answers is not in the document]
acumino: The whole codebase has capitalized the first letter, that's why I chose it.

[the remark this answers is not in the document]
aaronfern: Yes, completely understand
This PR adds quite a few things and I don't want it to repeat small mistakes that were added before

pkg/controller/machineset.go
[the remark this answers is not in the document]
acumino: Added a comment for explanation.

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,584 @@
...
+				return fmt.Errorf("failed to scale down machine set %s: %w", machineSet.Name, err)
+			}
+			klog.V(3).Infof("scaled down machine set %s to %d", machineSet.Name, len(oldMachines))
+		}
+	}
+
+	return nil
+}
+
+func (dc *controller) reconcileNewMachineSetInPlace(ctx context.Context, oldMachineSets []*v1alpha1.MachineSet, newMachineSet *v1alpha1.MachineSet, deployment *v1alpha1.MachineDeployment) (bool, error) {
```
unmarshall, line 215: The name of the method suggests that you are going to reconcile only new machineSet. That is not true, as you are also transferring machines from old to new and therefore you are also reconciling old machine set as a consequence.

pkg/controller/deployment_util.go
```diff
@@ -1220,3 +1291,45 @@ func statusUpdateRequired(old v1alpha1.MachineDeploymentStatus, new v1alpha1.Mac
 
 	return true
 }
+
+// MergeWithOverwriteAndFilter returns a new map containing keys and values from the `currentMap` map
+// that are not present in `oldMap`, combined with all keys and values from `newMap`.
+// If a key exists in both `currentMap` and `newMap`, the value from `newMap` is used.
+// This effectively merges the maps while overwriting and filtering as described.
+func MergeWithOverwriteAndFilter[T any](currentMap, oldMap, newMap map[string]T) map[string]T {
```
unmarshall, line 1299: - This is a very generically named method which has specific comments about oldIS.
- You have made this look like a utility function when this function makes assumptions. It assumes that old and new Map have same keys.
- Then you are half way there w.r.t using generics. You assume that the key is always a string instead of using `comparable` and value can be any Type. Where as the usage suggests that both key and values are always going to be strings.

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,584 @@
...
+			cond := getMachineCondition(oldMachine, v1alpha1.NodeInPlaceUpdate)
+			if isUpdateNotSuccessful(cond, node.Labels) {
+				continue
+			}
+
+			klog.V(3).Infof("Attempting to transfer machine %s to new machine set %s", oldMachine.Name, newMachineSet.Name)
+
+			// removes labels not present in newMachineSet so that the machine is not selected by the old machine set
+			machineNewLabels := MergeStringMaps(
+				MergeWithOverwriteAndFilter(oldMachine.Labels, machineSet.Spec.Selector.MatchLabels, newMachineSet.Spec.Selector.MatchLabels),
```
unmarshall, line 275: this could have been simplified to be easily understandable.

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,584 @@
...
+				map[string]string{v1alpha1.LabelKeyNodeUpdateResult: v1alpha1.LabelValueNodeUpdateSuccessful})
+
+			formattedLabels, err := labelsutil.GetFormattedLabels(machineNewLabels)
```
unmarshall, line 278: these are not `formattedLables` - what you are asking for is JSON bytes for a map.

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,584 @@
...
+			if err != nil {
+				return false, err
+			}
+			// update the owner reference of the machine to the new machine set and update the labels
+			addControllerPatch := fmt.Sprintf(
+				`{"metadata":{"ownerReferences":[{"apiVersion":"machine.sapcloud.io/v1alpha1","kind":"%s","name":"%s","uid":"%s","controller":true,"blockOwnerDeletion":true}],"labels":%s,"uid":"%s"}}`,
+				v1alpha1.SchemeGroupVersion.WithKind("MachineSet"),
+				newMachineSet.GetName(), newMachineSet.GetUID(), string(formattedLabels), oldMachine.UID)
+
+			err = dc.machineControl.PatchMachine(ctx, oldMachine.Namespace, oldMachine.Name, []byte(addControllerPatch))
```
unmarshall, line 288: in some places typed client's `Update` is used and here `Patch` is used. Can you explain the reason?

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,584 @@
...
+			if err != nil {
+				klog.Errorf("failed to transfer the ownership of machine %s to new machine set. Err: %v", oldMachine.Name, err)
+				return false, err
+			}
+
+			// uncordon the node since the ownership of the machine has been transferred to the new machine set.
+			node.Spec.Unschedulable = false
+			_, err = dc.targetCoreClient.CoreV1().Nodes().Update(ctx, node, metav1.UpdateOptions{})
```
unmarshall, line 296: what happens when the machine is patched, thereby changing its owner reference, but the node update fails? Will this node remain unschedulable?

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,584 @@
...
+			if err != nil {
+				return false, fmt.Errorf("failed to uncordon the node %s: %w", node.Name, err)
+			}
+
+			transferredMachineCount++ // scale down the old machine set.
```
unmarshall, line 301: the increment of `transferredMachineCount` is done after node is un-cordoned. But if the uncordoning fails then while the owner reference has been changed to the new MS, the transferredMachineCount will not reflect that. What is the consequence?

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,584 @@
...
+	if oldMachinesCount == 0 {
+		scaled, _, err := dc.scaleMachineSetAndRecordEvent(ctx, newMachineSet, deployment.Spec.Replicas, deployment)
+		if err != nil {
+			return false, fmt.Errorf("failed to scale up machine set %s: %w", newMachineSet.Name, err)
+		}
+		klog.V(3).Infof("scaled up machine set %s to %d", newMachineSet.Name, deployment.Spec.Replicas)
+		return scaled, err
+	}
+
+	addedNewReplicasCount := int32(0)
```
unmarshall, line 239: you can just defined this as `	var addedNewReplicasCount int32` instead of using casting to tell golang compiler the type of the variable. Default value for int32 is anyways 0.

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,584 @@
...
+
+			// uncordon the node since the ownership of the machine has been transferred to the new machine set.
+			node.Spec.Unschedulable = false
+			_, err = dc.targetCoreClient.CoreV1().Nodes().Update(ctx, node, metav1.UpdateOptions{})
+			if err != nil {
+				return false, fmt.Errorf("failed to uncordon the node %s: %w", node.Name, err)
+			}
+
+			transferredMachineCount++ // scale down the old machine set.
+			addedNewReplicasCount++   // scale up the new machine set.
```
unmarshall, line 302: both `transferredMachineCount` and `addedNewReplicasCount` are incremented together. In which situation will they have different values. If they will always have identical values then why not just use one variable?

unmarshall (changes requested):
Partially reviewed

pkg/controller/deployment_inplace.go
ary1992: Patch is used for machines and Update call is used for nodes.

pkg/controller/deployment_inplace.go
ary1992: > Will this node remain unschedulable?

No, `syncMachineSets` function will take care of these scenarios.

pkg/controller/deployment_inplace.go
ary1992: `syncMachineSets` function will make sure that the old machine set is scaled down and new machinesets are scaled up as per the machines owned by them.

pkg/controller/deployment_inplace.go
ary1992: `transferredMachineCount ` is defined per old machineset. For every old machine set `transferredMachineCount` will count the number of machines which are transferred to the new machineset and by this number the corresponding old machinset needs to be scaled down. We are assuming there can be multiple machine sets, hence the loop runs for all the old machinsets. `addedNewReplicasCount` replica count hold the value of total number of machines transferred to the new machineset (sum of machines transferred from all the old machine set) and new machineset needs to be scaled up by this amount.

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,584 @@
...
+	}
+
+	return totalSelectedForUpdate, nil
+}
+
+// labelNodesBackingMachineSets annotates all nodes backing the machineSets
+func (dc *controller) labelNodesBackingMachineSets(ctx context.Context, machineSets []*v1alpha1.MachineSet, labelKey, labelValue string) error {
+	for _, machineSet := range machineSets {
+
+		if machineSet == nil {
```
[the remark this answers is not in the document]
ary1992: This is copy-pasted code from an already existing function `annotateNodesBackingMachineSets` in `deployment_rolling.go` file.

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,584 @@
...
+		if isUpdateNotSuccessful(cond, node.Labels) {
+			continue
+		}
+
+		// remove labels related to the inplace update.
+		delete(node.Labels, v1alpha1.LabelKeyNodeCandidateForUpdate)
+		delete(node.Labels, v1alpha1.LabelKeyNodeSelectedForUpdate)
+		delete(node.Labels, v1alpha1.LabelKeyNodeUpdateResult)
+		// remove annotations related to the inplace update.
+		delete(node.Annotations, v1alpha1.AnnotationKeyMachineUpdateFailedReason)
```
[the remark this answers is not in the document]
ary1992: yes

pkg/apis/machine/v1alpha1/machine_types.go
```diff
@@ -201,6 +216,29 @@ const (
 	ConditionUnknown ConditionStatus = "Unknown"
 )
 
+const (
+	// NodeInPlaceUpdate is a node condition type for in-place update
+	NodeInPlaceUpdate corev1.NodeConditionType = "InPlaceUpdate"
```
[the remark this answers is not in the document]
ary1992: In the upstream https://github.com/kubernetes/kubernetes/blob/ef7d51e5ba77004cb95dc2127ed90f8bbfdabc72/pkg/apis/core/types.go#L5468 , this is how the node conditions are defined.

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,584 @@
...
+	if availableMachineCount <= minAvailable {
+		// Cannot pick for updating.
+		return 0, nil
+	}
+
+	sort.Sort(MachineSetsByCreationTimestamp(oldMachineSets))
+
+	totalSelectedForUpdate := int32(0)
+	maxSelectableForUpdate := min(availableMachineCount-minAvailable, max(deployment.Spec.Replicas-newMachineSet.Spec.Replicas, 0))
+	for _, targetIS := range oldMachineSets {
```
unmarshall, line 400: can you rename `targetIS` with `targetMachineSet`? Same in other places in this function.

unmarshall (changes requested):

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,584 @@
...
+
+			formattedLabels, err := labelsutil.GetLabelsAsJSONBytes(machineNewLabels)
+			if err != nil {
+				return false, err
+			}
+			// update the owner reference of the machine to the new machine set and update the labels
+			addControllerPatch := fmt.Sprintf(
+				`{"metadata":{"ownerReferences":[{"apiVersion":"machine.sapcloud.io/v1alpha1","kind":"%s","name":"%s","uid":"%s","controller":true,"blockOwnerDeletion":true}],"labels":%s,"uid":"%s"}}`,
+				v1alpha1.SchemeGroupVersion.WithKind("MachineSet"),
+				newMachineSet.GetName(), newMachineSet.GetUID(), string(formattedLabels), oldMachine.UID)
```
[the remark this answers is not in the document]
acumino: UID should change since the new machine set will have a different UID.

pkg/controller/deployment_inplace.go
```diff
@@ -0,0 +1,584 @@
...
+	return nil
+}
+
+func (dc *controller) reconcileNewMachineSetInPlace(ctx context.Context, oldMachineSets []*v1alpha1.MachineSet, newMachineSet *v1alpha1.MachineSet, deployment *v1alpha1.MachineDeployment) (bool, error) {
+	if newMachineSet.Spec.Replicas == deployment.Spec.Replicas {
+		// Scaling not required.
+		return false, nil
+	}
+
+	if newMachineSet.Spec.Replicas > deployment.Spec.Replicas {
```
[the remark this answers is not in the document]
acumino: Yes, it will not happen in the usual case, but what if manual intervention is there, so at the end of the day, reconciliation should stabilize any such inconsistencies.

unmarshall (approved):

PR #979, opened 2025-04-03
Check if pod still exists before further eviction is attempted

### LINKED ISSUE #974: MCM continues sending eviction requests although pods are gone from node

/area control-plane
/kind bug
/priority 3

**What happened**:
We have a set of pods for which we want to avoid eviction for (from a statefulset, say xyz) and our own operator takes care of proper rolling and deleting the pod ultimately.
So the idea was that the we have a webhook intercepting eviction requests (from MCM). It will for our xyz pods always deny the eviction and annotate the pod so that our operator takes over (it will delete it eventually).
Now what I see in the log of the webhook is that eviction requests are sent for example for xyz-2 even after xyz-2  was deleted by the operator and started on a new node (every 20 seconds).
I would have expected those requests to cease some time after the respective pods are moved out of the node?
Note that this happened when we changed the machine type in gardener UI, but I assume this makes no difference?
If I manually drain a node with kubectl drain  our logic works allright: There will be some errors because the eviction is denied, but eventuelly they are deleted by our operator. and kubectl drain  succeeds as theere are no more pods on the node.
Our assumption was that it would work the same way with gardener. 

Ultimately, the gardener cluster should reconcile successfully.

This is difficult without custom webhooks.

**Environment**:

- Kubernetes version (use `kubectl version`): 1.29.2
- Cloud provider or hardware configuration: aws 
- Others:

aaronfern:
We recently came across a bug where MCM kept on trying to evict a pod even though the pod was already evicted.
What we found was that MCM constructs a list of pods it wants to evict and keeps on trying to evict them until the eviction call succeeds.
This logic today fails when the pod eviction is done by another actor. It was seen that MCM does not fetch the pod again from the cache and hence does not see that the pod is already evicted

This PR adds a check to see if the pod is already evicted before it continues eviction attempts

Fixes #974 

**Release note**:

```bugfix operator
Fix a bug where MCM does not check if a pod has already been evicted before reattempting eviction blindly
```

### DIFF
--- a/pkg/util/provider/drain/drain.go
+++ b/pkg/util/provider/drain/drain.go
@@ -416,7 +416,7 @@ func (o *Options) evictPod(ctx context.Context, pod *corev1.Pod, policyGroupVers
 		DeleteOptions: deleteOptions,
 	}
 	klog.V(3).Infof("Attempting to evict the pod:%q from node %q", pod.Name, o.nodeName)
-	// TODO: Remember to change the URL manipulation func when Evction's version change
+	// TODO: Remember to change the URL manipulation func when Eviction's version change
 	return o.client.PolicyV1beta1().Evictions(eviction.Namespace).Evict(ctx, eviction)
 }
 
@@ -682,7 +682,7 @@ func (o *Options) evictPodsWithPv(ctx context.Context, attemptEvict bool, pods [
 }
 
 // checkAndDeleteWorker is a helper method that check if volumeAttachmentHandler
-// is supported and delete's the worker from the list of event handlers
+// is supported and deletes the worker from the list of event handlers
 func (o *Options) checkAndDeleteWorker(volumeAttachmentEventCh chan *storagev1.VolumeAttachment) {
 	if o.volumeAttachmentHandler != nil {
 		o.volumeAttachmentHandler.DeleteWorker(volumeAttachmentEventCh)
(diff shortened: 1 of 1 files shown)


aaronfern:
## Testing This PR
---

#### Replicating this bug

1. Apply the following validating webhook on a cluster. This webhook blocks all pod eviction calls
<details>
<summary>ValidatingWebhookConfiguration</summary>

```
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingWebhookConfiguration
metadata:
  name: admissionwebhooks.org.com
webhooks:
- admissionReviewVersions:
  - v1
  - v1beta1
  clientConfig:
    service:
      name: admissionwebhooks
      namespace: default
      path: /intercept-evictions
      port: 443
  failurePolicy: Fail
  matchPolicy: Equivalent
  name: admissionwebhooks.org.com
  namespaceSelector:
    matchExpressions:
    - key: gardener.cloud/purpose
      operator: NotIn
      values:
      - kube-system
  timeoutSeconds: 1
  rules:
  - apiGroups:
    - ""
    apiVersions:
    - v1
    operations:
    - CREATE
    - UPDATE
    - DELETE
    - CONNECT
    resources:
    - pods/eviction
    scope: Namespaced
  sideEffects: None
```

</details>

2. Create some sample pods on the cluster
3. Drain the node where this pod is running

4. MCM will try to evict this pod, but the webhook will block the eviction
5. Now delete this pod manually (using `kubectl delete po <pod_name>`). MCM will still see this pod and continue trying to evict this pod even though this pod no longer exists

---
#### PR 
With the changes that are part of this PR, run the same test above
After deleting the pod manually, MCM should notice that the pod no longer exists and go ahead with the pod drain

takoverflow (approved):

Issue #961, opened 2024-12-30
`MachineDeploymentFrozen` has the same doc string as `MachineDeploymentReplicaFailure `


Kostov6:
**How to categorize this issue?**

/area documentation
/kind bug
/priority 5

**What happened**:

https://github.com/gardener/machine-controller-manager/blob/f73366907e5c7a6c7b6fe2dad846ad6b646986db/pkg/apis/machine/v1alpha1/machinedeployment_types.go#L216-L222

**What you expected to happen**:

`MachineDeploymentFrozen` and `MachineDeploymentReplicaFailure` to not have the same doc string as this implies that the conditions are practically the same

**Environment**:

- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- Others:

PR #981, opened 2025-04-09
Minor fixes

### LINKED ISSUE #977: Machine creation retry too frequently for machines with `ResourceExhausted`
/area robustness
/kind enhancement
/priority 3

- Adding a check for `codes.ResourceExhausted` with a higher retry period.

https://github.com/gardener/machine-controller-manager/blob/1e2563f85f5b6912aeb7ea42b5c5264ca67d667a/pkg/util/provider/machinecontroller/machine_util.go#L503-L515

**Why is this needed**:
Currently, machines that fail due to `codes.ResourceExhaused`, are retried using `machineutils.MediumRetry` which is every 3 minutes.
When a resource in the underlying infrastructure is exhausted, it's unlikely that this will change in that short period of time.

The `provider-openstack` first creates the `volume` and then the `machine`.
However, depending on the size of the `nodePool`, this can lead to a large number of unnecessary create/delete API calls.

takoverflow:

Fixes #977 and #961

**Release note**:

```other operator
Resource exhaustion on machine creation results in a longer retry period
```

### DIFF
--- a/pkg/apis/machine/types.go
+++ b/pkg/apis/machine/types.go
@@ -615,8 +615,8 @@ const (
 	// or deleted.
 	MachineDeploymentReplicaFailure MachineDeploymentConditionType = "ReplicaFailure"
 
-	// MachineDeploymentFrozen is added in a MachineDeployment when one of its machines fails to be created
-	// or deleted.
+	// MachineDeploymentFrozen is added in a MachineDeployment when one of its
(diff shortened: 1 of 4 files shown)


aaronfern (approved):
Thanks for the PR!

Can you please add a release note for long retry on `ResourceExhausted` errors?

takoverflow:
> Can you please add a release note for long retry on `ResourceExhausted` errors?

Done

PR #987, opened 2025-04-23
[GEP-31] Controller changes to Support Manual InPlaceUpdate strategy

ary1992:
This PR adds support for manual In-Place update.

Part of https://github.com/gardener/gardener/issues/10219

/invite @aaronfern @elankath @unmarshall

**Release note**:

```other operator
NONE
```


pkg/controller/deployment_inplace.go
```diff
@@ -284,6 +289,13 @@ func (dc *controller) reconcileOldMachineSetsInPlace(ctx context.Context, allMac
 		return false, err
 	}
 
+	// Machines from old machine sets which are undergoing update will eventually move to new machine set.
+	// So once the current new machine set replcas + old machine set replicas undergoing update reaches the desired replicas,
+	// we can stop selecting machines from old machine sets for update.
+	if oldISsMachinesUndergoingUpdate+newMachineSet.Spec.Replicas >= deployment.Spec.Replicas {
+		return false, nil
```
aaronfern, line 296: Should this not return `true, nil`? I assume we would want `syncRolloutStatus()` to run

acumino: The last reconciliation would have already called the `syncRolloutStatus`, so we don't need it.

pkg/controller/deployment_inplace.go
```diff
@@ -25,8 +25,8 @@ import (
 	"k8s.io/utils/integer"
 )
 
-// rolloutAutoInPlace implements the logic for rolling  a machine set without replacing it.
-func (dc *controller) rolloutAutoInPlace(ctx context.Context, d *v1alpha1.MachineDeployment, isList []*v1alpha1.MachineSet, machineMap map[types.UID]*v1alpha1.MachineList) error {
+// rolloutInPlace implements the logic for rolling a machine set without replacing it.
```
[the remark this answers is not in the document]
acumino: Adapted PTAL

pkg/controller/deployment_inplace.go
```diff
@@ -273,6 +273,11 @@ func (dc *controller) reconcileOldMachineSetsInPlace(ctx context.Context, allMac
 		return true, nil
 	}
 
+	// Manual inplace update
+	if deployment.Spec.Strategy.InPlaceUpdate.OrchestrationType == v1alpha1.OrchestrationTypeManual {
+		return true, nil
```
aaronfern, line 278: Since we return `true, nil` here, this would mean that in the caller the `if MachineDeploymentComplete(d, &d.Status) {...` block will not be called and ig eventually we would like that to be called, no?
I do not see a way that this block will eventually be called as well except [here](https://github.com/gardener/machine-controller-manager/blob/master/pkg/controller/deployment_inplace.go#L259)
Am I missing anything?

aaronfern (commented):
Thanks for addressing my doubt @acumino!
I have one more doubt, please address that
Thanks!

pkg/controller/deployment_inplace.go
acumino: That's correct, I changed it to return false. Thanks for pointing it out.

aaronfern (approved):
Thanks for the changes!
/lgtm

pkg/controller/deployment.go
```diff
@@ -222,6 +222,32 @@ func (dc *controller) deleteMachineSetToDeployment(obj interface{}) {
 	dc.enqueueMachineDeployment(d)
 }
 
+// updateMachineToMachineDeployment will enqueue the machine deployment if the machine InPlaceUpdate node condition changes to UpdateSuccessful.
+func (dc *controller) updateMachineToMachineDeployment(old, cur any) {
```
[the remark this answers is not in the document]
acumino: Yes, also we can process faster, as soon as machine get update, it should move to new machine set and new machine should get selected to get updated.

pkg/controller/deployment_inplace.go
```diff
@@ -273,6 +274,11 @@ func (dc *controller) reconcileOldMachineSetsInPlace(ctx context.Context, allMac
 		return true, nil
 	}
 
+	// Manual inplace update
+	if deployment.Spec.Strategy.InPlaceUpdate.OrchestrationType == v1alpha1.OrchestrationTypeManual {
```
unmarshall, line 278: For manual orchestration type, our understanding was that `MaxSurge` cannot be set as it would no longer be manual. In addition, MCM does not orchestrate in manual mode which means even if `MaxSurge` is set to any value > 0 then MCM will not be able to honor it. 

So the question is can we move this check above 266 line number where you check:
```go
if newMachineSet.Spec.Replicas == deployment.Spec.Replicas {
}
```

pkg/controller/deployment_inplace.go
```diff
@@ -284,6 +290,13 @@ func (dc *controller) reconcileOldMachineSetsInPlace(ctx context.Context, allMac
 		return false, err
 	}
 
+	// Machines from old machine sets which are undergoing update will eventually move to new machine set.
+	// So once the current new machine set replcas + old machine set replicas undergoing update reaches the desired replicas,
+	// we can stop selecting machines from old machine sets for update.
+	if oldISsMachinesUndergoingUpdate+newMachineSet.Spec.Replicas >= deployment.Spec.Replicas {
```
unmarshall, line 296: Can you please refrain from using `IS` and instead use `MachineSet`?

unmarshall (changes requested):

pkg/controller/deployment_inplace.go
acumino: I would prefer to keep it here, just to save any manual intervention issue.

pkg/controller/deployment_inplace.go
acumino: Adapted it, PTAL

pkg/controller/deployment.go
```diff
@@ -243,7 +269,7 @@ func (dc *controller) deleteMachineToMachineDeployment(ctx context.Context, obj
 		}
 	}
 	klog.V(4).Infof("Machine %s deleted.", machine.Name)
-	if d := dc.getMachineDeploymentForMachine(ctx, machine); d != nil && d.Spec.Strategy.Type == v1alpha1.RecreateMachineDeploymentStrategyType {
+	if d := dc.getMachineDeploymentForMachine(machine); d != nil && d.Spec.Strategy.Type == v1alpha1.RecreateMachineDeploymentStrategyType {
```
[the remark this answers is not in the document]
elankath: Can you please re-introduce `ctx` ? It should not be removed.

elankath (changes requested):
Please address feedback and will approve asap.

pkg/controller/deployment.go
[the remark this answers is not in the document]
acumino: its's not needed any more.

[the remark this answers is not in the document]
acumino: cc @elankath PTAL.

elankath (approved):

pkg/controller/deployment.go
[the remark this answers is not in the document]
acumino: Nothing will change in the old behavior, `deleteMachineToMachineDeployment` is dead code.

Issue #994, opened 2025-05-16
[GEP-28] Support managing machines without target cluster


timebertt:
**How to categorize this issue?**

/area ipcei
/kind enhancement

**What would you like to be added**:

If machine-controller-manager or the machine-controller-manager-provider is started with `--target-kubeconfig=none`, it should disable all interactions with the target cluster.

This should allow the creation of machines in the infrastructure without an existing target cluster. Notable differences from the standard machine creation flow are:

- mcm doesn't create bootstrap token secrets
- mcm doesn't set the `node` label on the `Machine` object
- mcm doesn't wait for the `Node` object to exist in the target cluster
- mcm sets the `Machine` object to `Available` after creating it in the infrastructure (target state)

When deleting machines without a target cluster:

- mcm doesn't drain or delete the `Node` in the target cluster
- mcm doesn't delete or wait for `VolumeAttachments` in the target cluster

**Why is this needed**:

[GEP-28](https://github.com/gardener/gardener/tree/master/docs/proposals/28-autonomous-shoot-clusters.md) (https://github.com/gardener/gardener/issues/2906) includes the [medium-touch scenario](https://github.com/gardener/gardener/blob/master/docs/proposals/28-autonomous-shoot-clusters.md#medium-touch), where the control plane machines of the autonomous shoot cluster are managed using the provider extension and machine-controller-manager (`gardenadm bootstrap`). The main idea is to reuse existing components for managing the infrastructure resources (network resources and machines), and then `gardenadm init` will bootstrap the cluster's control plane when executed on the prepared machines.

In contrast to normal shoot clusters (with a control plane hosted on a seed cluster), there is no control plane running when creating machines (with `gardenadm bootstrap`) that will be used as control plane nodes. To support this scenario, machine-controller-manager should allow disabling all interactions with the target cluster.

For successfully running `gardenadm bootstrap`, supporting the deletion of machines without a target cluster might not be needed. However, the automatic deletion of machines will be helpful in error cases, e.g., when misconfiguring the machines and during development.

timebertt:
Starting to implement this :)

PR #1004, opened 2025-06-12
[GEP-28] Support managing machines without target cluster for `gardenadm bootstrap`

timebertt:
/kind enhancement

This PR implements changes to support provisioning control plane machines of autonomous shoot clusters in `gardenadm bootstrap` ([GEP-28](https://github.com/gardener/gardener/tree/master/docs/proposals/28-autonomous-shoot-clusters.md) medium-touch scenario).
If `--target-kubeconfig=none` is configured, MCM skips all interaction with the target cluster as described in the referenced issue.

Fixes https://github.com/gardener/machine-controller-manager/issues/994

I built `ghcr.io/timebertt/machine-controller-manager:gardenadm-bootstrap` for testing this PR in a gardener local setup.
You can check out (https://github.com/gardener/gardener/pull/12307) to test this PR with `gardenadm bootstrap`.

**Release note**:

```feature operator
machine-controller-manager can manage machines without a target cluster by specifying `--target-kubeconfig=none`. See the [documentation](https://github.com/gardener/machine-controller-manager/blob/master/docs/FAQ.md#how-to-run-mcm-in-different-cluster-setups) for more details.
```

### DIFF
--- a/cmd/machine-controller-manager/app/controllermanager.go
+++ b/cmd/machine-controller-manager/app/controllermanager.go
@@ -105,9 +105,6 @@ func Run(s *options.MCMServer) error {
 			return err
 		}
 	} else {
-		if s.TargetKubeconfig == constants.TargetKubeconfigDisabledValue {
-			return fmt.Errorf("--control-kubeconfig cannot be empty if --target-kubeconfig=%s is specified", constants.TargetKubeconfigDisabledValue)
-		}
 		controlkubeconfig = targetkubeconfig
 	}
 
--- a/cmd/machine-controller-manager/app/options/options.go
+++ b/cmd/machine-controller-manager/app/options/options.go
@@ -36,7 +36,6 @@ import (
 	"github.com/gardener/machine-controller-manager/pkg/util/client/leaderelectionconfig"
 
 	// add the machine feature gates
-	"fmt"
 	"github.com/gardener/machine-controller-manager/pkg/apis/constants"
 	_ "github.com/gardener/machine-controller-manager/pkg/features"
 )
@@ -147,6 +146,9 @@ func (s *MCMServer) Validate() error {
 	}
 	if s.SafetyOptions.MachineSafetyOvershootingPeriod.Duration < 0 {
 		errs = append(errs, fmt.Errorf("machine safety overshooting period should be a non negative number: got: %v", s.SafetyOptions.MachineSafetyOvershootingPeriod.Duration))
+	}
+	if s.ControlKubeconfig == "" && s.TargetKubeconfig == constants.TargetKubeconfigDisabledValue {
+		errs = append(errs, fmt.Errorf("--control-kubeconfig cannot be empty if --target-kubeconfig=%s is specified", constants.TargetKubeconfigDisabledValue))
 	}
 
 	return utilerrors.NewAggregate(errs)
--- a/pkg/util/provider/app/app.go
+++ b/pkg/util/provider/app/app.go
@@ -103,9 +103,6 @@ func Run(s *options.MCServer, driver driver.Driver) error {
 			return err
 		}
 	} else {
-		if s.TargetKubeconfig == constants.TargetKubeconfigDisabledValue {
-			return fmt.Errorf("--control-kubeconfig cannot be empty if --target-kubeconfig=%s is specified", constants.TargetKubeconfigDisabledValue)
-		}
 		controlkubeconfig = targetkubeconfig
 	}
 
--- a/pkg/util/provider/app/options/options.go
+++ b/pkg/util/provider/app/options/options.go
@@ -37,7 +37,6 @@ import (
 	"github.com/gardener/machine-controller-manager/pkg/util/client/leaderelectionconfig"
 
 	// add the machine feature gates
-	"fmt"
 	"github.com/gardener/machine-controller-manager/pkg/apis/constants"
 	_ "github.com/gardener/machine-controller-manager/pkg/features"
 )
@@ -187,6 +186,9 @@ func (s *MCServer) Validate() error {
 	}
 	if s.SafetyOptions.MachineSafetyAPIServerStatusCheckPeriod.Duration < s.SafetyOptions.MachineSafetyAPIServerStatusCheckTimeout.Duration {
 		errs = append(errs, fmt.Errorf("machine safety APIServer status check period should not be less than APIServer status check timeout"))
+	}
+	if s.ControlKubeconfig == "" && s.TargetKubeconfig == constants.TargetKubeconfigDisabledValue {
+		errs = append(errs, fmt.Errorf("--control-kubeconfig cannot be empty if --target-kubeconfig=%s is specified", constants.TargetKubeconfigDisabledValue))
 	}
 
 	return utilerrors.NewAggregate(errs)


gardener-robot-ci-1:
Thank you @timebertt for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

cmd/machine-controller-manager/app/controllermanager.go
```diff
@@ -64,6 +64,8 @@ const (
 	controllerManagerAgentName = "machine-controller-manager"
 	targetkubeconfigTimeout    = 1 * time.Minute
 	controlkubeconfigTimeout   = 1 * time.Minute
+
+	targetKubeconfigValueDisabled = "none"
```
elankath: `targetKubeconfigValueDisabled` -> `TargetKubeconfigDisabledValue` and move it to new file  `pkg/apis/constants/constants.go` and reuse this instead of repeating constant.

cmd/machine-controller-manager/app/controllermanager.go
```diff
@@ -240,17 +257,20 @@ func StartControllers(s *options.MCMServer,
...
 		s.MinResyncPeriod.Duration,
 		s.Namespace,
 		nil,
 	)
 
-	targetCoreInformerFactory := coreinformers.NewSharedInformerFactory(
-		targetCoreClientBuilder.ClientOrDie("target-core-shared-informers"),
-		s.MinResyncPeriod.Duration,
-	)
+	var targetCoreInformerFactory kubernetesinformers.SharedInformerFactory
```
unmarshall, line 267: instead of passing the entire SharedInformerFactory you can just initialize `NodeInformer`. We used to pass `targetCoreInformerFactory.Core().V1().Nodes()` before. Now this will either be nil in case targetCoreClientBuilder is nil or it will `targetCoreInformerFactory.Core().V1().Nodes()`

Any reason to pass the entire `SharedInformerFactory`?

cmd/machine-controller-manager/app/controllermanager.go
```diff
@@ -261,7 +281,7 @@ func StartControllers(s *options.MCMServer,
 		controlMachineClient,
 		controlCoreClient,
 		targetCoreClient,
-		targetCoreInformerFactory.Core().V1().Nodes(),
+		targetCoreInformerFactory,
```
unmarshall, line 280: Instead of passing the entire SharedInformerFactory you can just create a `NodeInformer`. 

We used to pass `targetCoreInformerFactory.Core().V1().Nodes()` before. We can retain the same, this will either be nil in case targetCoreClientBuilder is nil or it will `targetCoreInformerFactory.Core().V1().Nodes()`

Any reason to pass the entire `SharedInformerFactory`?

cmd/machine-controller-manager/app/options/options.go
```diff
@@ -82,7 +82,7 @@ func (s *MCMServer) AddFlags(fs *pflag.FlagSet) {
 	fs.DurationVar(&s.MinResyncPeriod.Duration, "min-resync-period", s.MinResyncPeriod.Duration, "The resync period in reflectors will be random between MinResyncPeriod and 2*MinResyncPeriod")
 	fs.BoolVar(&s.EnableProfiling, "profiling", true, "Enable profiling via web interface host:port/debug/pprof/")
 	fs.BoolVar(&s.EnableContentionProfiling, "contention-profiling", false, "Enable lock contention profiling, if profiling is enabled")
-	fs.StringVar(&s.TargetKubeconfig, "target-kubeconfig", s.TargetKubeconfig, "Filepath to the target cluster's kubeconfig where node objects are expected to join")
+	fs.StringVar(&s.TargetKubeconfig, "target-kubeconfig", s.TargetKubeconfig, "Filepath to the target cluster's kubeconfig where node objects are expected to join or 'none' if there is no target cluster")
```
elankath:
```go
fs.StringVar(&s.TargetKubeconfig, "target-kubeconfig", s.TargetKubeconfig, fmt.Sprintf("Filepath to the target cluster's kubeconfig where node objects are expected to join or %q if there is no target cluster", constants.TargetKubeconfigDisabledValue))
```

pkg/util/provider/app/app.go
```diff
@@ -64,6 +64,8 @@ import (
 
 const (
 	controllerManagerAgentName = "machine-controller"
+
+	targetKubeconfigValueDisabled = "none"
```
elankath: Please use the suggested`constants` package instead of repeating this.

pkg/util/provider/app/options/options.go
```diff
@@ -90,7 +90,7 @@ func (s *MCServer) AddFlags(fs *pflag.FlagSet) {
 	fs.DurationVar(&s.MinResyncPeriod.Duration, "min-resync-period", s.MinResyncPeriod.Duration, "The resync period in reflectors will be random between MinResyncPeriod and 2*MinResyncPeriod")
 	fs.BoolVar(&s.EnableProfiling, "profiling", true, "Enable profiling via web interface host:port/debug/pprof/")
 	fs.BoolVar(&s.EnableContentionProfiling, "contention-profiling", false, "Enable lock contention profiling, if profiling is enabled")
-	fs.StringVar(&s.TargetKubeconfig, "target-kubeconfig", s.TargetKubeconfig, "Filepath to the target cluster's kubeconfig where node objects are expected to join")
+	fs.StringVar(&s.TargetKubeconfig, "target-kubeconfig", s.TargetKubeconfig, "Filepath to the target cluster's kubeconfig where node objects are expected to join or 'none' if there is no target cluster")
```
elankath: fs.StringVar(&s.TargetKubeconfig, "target-kubeconfig", s.TargetKubeconfig, fmt.Sprintf("Filepath to the target cluster's kubeconfig where node objects are expected to join or %q if there is no target cluster", constants.TargetKubeconfigDisabledValue))

cmd/machine-controller-manager/app/controllermanager.go
timebertt: Hmm, your suggestion will cause panics when dereferencing a nil interface in `targetCoreInformerFactory.Core().V1().Nodes()`. What am I missing?

Passing the entire `SharedInformerFactory` was simpler when multiple informers are needed (see this PR's changes to `pkg/util/provider/app/app.go`). I performed a similar change in `pkg/controller/controller.go` to handle it consistently.

timebertt:
@elankath @unmarshall can you kindly take another look at this PR? :)

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -661,14 +674,15 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
...
 	clone = machine.DeepCopy()
-	if !machineNodeLabelPresent || !machinePriorityAnnotationPresent || machine.Spec.ProviderID == "" {
-		if clone.Labels == nil {
-			clone.Labels = make(map[string]string)
+	if machineNodeLabelMissing || !machinePriorityAnnotationPresent || machine.Spec.ProviderID == "" {
+		if c.targetCoreClient != nil {
+			// If running without a target cluster, don't add the node label. This disables all interaction with the
+			// Node object and related objects in the target cluster.
+			metav1.SetMetaDataLabel(&clone.ObjectMeta, v1alpha1.NodeLabelKey, nodeName)
 		}
```
timebertt, line 685: I just figured that, for `gardenadm bootstrap` to work, we need to know the control plane machines' IPs or hostnames so that it knows where to jump to from the `Bastion`. This was not considered in the original proposal for this PR: https://github.com/gardener/machine-controller-manager/issues/994

My change relies on the fact that if the `node` label is not set, MCM skips all of the target cluster interaction, because it considers that there is no `Node` object attached to the `Machine` – which is correct. That's why my change skips setting the `Node` label in the first place.
With this, however, we also lose the information returned in `CreateMachineResponse.NodeName`.

I see two options:
1. Go back to setting the `node` label even if there is no target cluster. This will require many `targetClient != nil` checks.
2. Add the `CreateMachineResponse.NodeName` information to the `Machine` status (`status.hostname` or even `status.addresses[]`). This keeps the interaction with the `Node` object separate from retrieving the needed VM information. We could also add the private IP(s) to the `CreateMachineResponse` or `InitializeMachineResponse` and a corresponding `Machine.status` field later on if required (some infrastructures might not allow resolving other machine hostnames in the same network).

Option 2 can be implemented in a dedicated PR. Option 1 would require reworking this PR, though.

I prefer option 2. But I would like to get some feedback on this question from @gardener/mcm-maintainers.

aaronfern: From what I can see, I would prefer option 2.
Would not prefer setting the node label when there is no target cluster, and adding this info to the status seems fine to me

One question though. Would it be possible for mcm to distinguish between the `gardenadm bootstrap` case and any other case? I ask because if this is (easily) possible then it would be nice for the `status.hostname` (or whatever name is eventually chosen) field to be added only in this one case

timebertt: MCM can identify the `gardenadm bootstrap` use case by the missing target client, i.e., `--target-kubeconfig=none` and `targetClient == nil`.
Accordingly, we can populate the proposed status field(s) only when running without a target cluster.

timebertt: Thanks for your feedback. 
I opened https://github.com/gardener/machine-controller-manager/issues/1007 as a proposal. We will open a dedicated PR for implementing this feature.
In this case, we can continue with this PR :)

cmd/machine-controller-manager/app/controllermanager.go
```diff
@@ -82,35 +83,44 @@ func Run(s *options.MCMServer) error {
...
-			//kubeconfig for the seedcluster where MachineCRDs are supposed to be registered.
+			// kubeconfig for the seed cluster where MachineCRDs are supposed to be registered.
 			controlkubeconfig, err = clientcmd.BuildConfigFromFlags("", s.ControlKubeconfig)
 		}
 		if err != nil {
 			return err
 		}
+	} else {
+		if s.TargetKubeconfig == constants.TargetKubeconfigDisabledValue {
+			return fmt.Errorf("--control-kubeconfig cannot be empty if --target-kubeconfig=%s is specified", constants.TargetKubeconfigDisabledValue)
```
aaronfern: Not a hard opinion, but this ideally is a validation and should be part of `Validate()` in `/app/options/options.go`

pkg/util/provider/app/app.go
```diff
@@ -80,33 +81,42 @@ func Run(s *options.MCServer, driver driver.Driver) error {
...
-			//kubeconfig for the seedcluster where MachineCRDs are supposed to be registered.
+			// kubeconfig for the seed cluster where MachineCRDs are supposed to be registered.
 			controlkubeconfig, err = clientcmd.BuildConfigFromFlags("", s.ControlKubeconfig)
 		}
 		if err != nil {
 			return err
 		}
+	} else {
+		if s.TargetKubeconfig == constants.TargetKubeconfigDisabledValue {
+			return fmt.Errorf("--control-kubeconfig cannot be empty if --target-kubeconfig=%s is specified", constants.TargetKubeconfigDisabledValue)
```
aaronfern: Similar here as well. This can be a validation is `/pkg/util/provider/app/options/options.go`

cmd/machine-controller-manager/app/controllermanager.go
timebertt: Good point, done. PTAL :)

PR #1010, opened 2025-07-02
Added a new target `make check-generate` to Makefile

aaronfern:
This PR does the following
1. Adds a new target `make check-generate` to the `Makefile`. This target is intended to verify if apis/docs have been correctly generated
2. Added `make check-generate` as a check in the `.ci/check` script so that this can run as part of the `check` script in any pipeline

Fixes #

**Release note**:

```other operator
Added a `make check-generate` target to the Makefile
```

### DIFF
--- a/hack/check-generate.sh
+++ b/hack/check-generate.sh
@@ -68,7 +68,7 @@ if which git &>/dev/null; then
 
   old_status="$(git status -s)"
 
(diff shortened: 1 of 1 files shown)


aaronfern:
I've noticed too many anomalies with the check-generate script 
I will fix it and open another PR
/close

PR #1012, opened 2025-07-07
[GEP-28] Report `Machine.status.addresses` when running without target cluster

### LINKED ISSUE #1007: [GEP-28] Report `Machine.status.addresses` when running without target cluster

/area ipcei
/kind enhancement

If the MCM provider runs without a target cluster (`--target-kubeconfig=none`), it should report a new status field: `Machine.status.addresses`. The type is the same as `Node.status.addresses`.

If the driver returns `{CreateMachine,InitializeMachine,GetMachineStatus}Response.NodeName`, it should be added as an element with `type=Hostname` to the new status field, e.g.:
```yaml
status:
  addresses:
  - type: Hostname
    address: shoot--foo--bar-worker-z1-77fd4-7tl93
```

`{CreateMachine,InitializeMachine,GetMachineStatus}Response` is extended with a new `Addresses` field so that the driver can return additional addresses as reported by the infrastructure API. All `{CreateMachine,InitializeMachine,GetMachineStatus}Response.Addresses` are merged into `Machine.status.addresses`.
With this, a driver can report all machine addresses, e.g., `type=InternalIP` or even `type=ExternalIP` where applicable.

**Why is this needed**:

With https://github.com/gardener/machine-controller-manager/issues/994, MCM supports managing machines without a target cluster for the [GEP-28 medium-touch scenario](https://github.com/gardener/gardener/blob/master/docs/proposals/28-autonomous-shoot-clusters.md#medium-touch) (https://github.com/gardener/gardener/issues/2906).

To bootstrap the control plane machines, `gardenadm bootstrap` needs to connect to the machines via SSH to run commands and copy files. 
For most cases, this will be done by creating a [`Bastion`](https://github.com/gardener/gardener/blob/master/docs/extensions/resources/bastion.md) and proxying the node SSH connection over the bastion SSH connection.

Independent of whether a bastion connection or a direct connection to the machine (only possible with an external IP) is used, `gardenadm` needs to determine how to connect to the machine (i.e., the SSH target host). 
Typically, a client (e.g., `gardenctl ssh`) would consult the `Node.status.addresses` field for this ([ref](https://github.com/gardener/gardenctl-v2/blob/588e535a3614a08e307998511c820ab38ef2ba3a/pkg/cmd/ssh/connect_information.go#L123)).
However, there is no target cluster in `gardenadm bootstrap` – so no `Node` object.

By reporting `Machine.status.addresses` based on the infrastructure API, similar to what is reported in `Node.status.addresses`, MCM can support the `gardenadm bootstrap` use case.

maboehm:
Implements the suggestion from #1007 which is required for `gardenadm bootstrap` by introducing a new status field `status.addresses`, which is **only** populated when the MCM runs without a target cluster (as usually the `Node` status should be used).

The `Driver` interface Response types are also extended to allow providers to return additional addresses. It is assumed that the `nodeName` is always a valid address of `type=HostName`, so it is always added.

Fixes #1007

/cc @timebertt 

I also created a draft PR in gardener https://github.com/gardener/gardener/pull/12489, where the new fields are returned in the provider-local. This results in the following status on the machine object:

```yaml
status:
  addresses:
  - address: machine-shoot--garden--root-control-plane-695dc-tr9jl
    type: Hostname
  - address: 10.1.130.204
    type: InternalIP
```

**Release note**:

```feature operator
When running without a target cluster machine-controller-manager will report a new status field `addresses`
```

### DIFF
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -492,13 +492,13 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 		machine              = createMachineRequest.Machine
 		machineName          = createMachineRequest.Machine.Name
 		uninitializedMachine = false
-		addresses            = map[corev1.NodeAddress]struct{}{}
+		addresses            = sets.New[corev1.NodeAddress]()
 	)
 
 	// This field is only modified during the creation flow. We have to assume that every Address that was added to the
 	// status in the past remains valid, otherwise we have no way of keeping the addresses that might be only returned
 	// by the `CreateMachine` or `InitializeMachine` calls. Before persisting, the addresses are deduplicated.
-	appendAddresses(addresses, createMachineRequest.Machine.Status.Addresses)
+	addresses.Insert(createMachineRequest.Machine.Status.Addresses...)
 
 	// we should avoid mutating Secret, since it goes all the way into the Informer's store
 	secretCopy := createMachineRequest.Secret.DeepCopy()
@@ -546,7 +546,7 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 				}
 				nodeName = createMachineResponse.NodeName
 				providerID = createMachineResponse.ProviderID
-				appendAddresses(addresses, createMachineResponse.Addresses)
+				addresses.Insert(createMachineResponse.Addresses...)
 				// Creation was successful
 				klog.V(2).Infof("Created new VM for machine: %q with ProviderID: %q and backing node: %q", machine.Name, providerID, nodeName)
 
@@ -628,7 +628,7 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 		}
 		nodeName = getMachineStatusResponse.NodeName
 		providerID = getMachineStatusResponse.ProviderID
-		appendAddresses(addresses, getMachineStatusResponse.Addresses)
+		addresses.Insert(getMachineStatusResponse.Addresses...)
 	}
 
 	//Update labels, providerID
@@ -649,7 +649,7 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 		if c.targetCoreClient == nil {
 			// persist addresses from the InitializeMachine and CreateMachine responses
 			clone := clone.DeepCopy()
-			appendAddresses(addresses, initResponse.Addresses)
+			addresses.Insert(initResponse.Addresses...)
 			clone.Status.Addresses = buildAddressStatus(addresses, nodeName)
 			if _, err := c.controlMachineClient.Machines(clone.Namespace).UpdateStatus(ctx, clone, metav1.UpdateOptions{}); err != nil {
 				return machineutils.ShortRetry, fmt.Errorf("failed to persist status addresses after initialization was successful: %w", err)
@@ -840,20 +840,14 @@ func (c *controller) triggerDeletionFlow(ctx context.Context, deleteMachineReque
 	return machineutils.LongRetry, nil
 }
 
-func appendAddresses(addresses map[corev1.NodeAddress]struct{}, newAddresses []corev1.NodeAddress) {
-	for _, address := range newAddresses {
-		addresses[address] = struct{}{}
-	}
-}
-
 // buildAddressStatus adds the nodeName as a HostName address, if it is not empty, and returns a sorted and deduplicated
 // slice.
-func buildAddressStatus(addresses map[corev1.NodeAddress]struct{}, nodeName string) []corev1.NodeAddress {
+func buildAddressStatus(addresses sets.Set[corev1.NodeAddress], nodeName string) []corev1.NodeAddress {
 	if nodeName != "" {
-		addresses[corev1.NodeAddress{
+		addresses.Insert(corev1.NodeAddress{
 			Type:    corev1.NodeHostName,
 			Address: nodeName,
-		}] = struct{}{}
+		})
 	}
 	res := slices.Collect(maps.Keys(addresses))
 	slices.SortStableFunc(res, func(a, b corev1.NodeAddress) int {


gardener-robot-ci-1:
Thank you @maboehm for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -619,26 +627,44 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
...
 		var retryPeriod machineutils.RetryPeriod
-		retryPeriod, err = c.initializeMachine(ctx, clone, createMachineRequest.MachineClass, createMachineRequest.Secret)
+		var initAddresses []corev1.NodeAddress
+		retryPeriod, initAddresses, err = c.initializeMachine(ctx, clone, createMachineRequest.MachineClass, createMachineRequest.Secret)
 		if err != nil {
 			return retryPeriod, err
 		}
+
+		if c.targetCoreClient == nil {
+			// persist addresses from the InitializeMachine and CreateMachine responoses
```
timebertt:
```suggestion
			// persist addresses from the InitializeMachine and CreateMachine responses
```

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -619,26 +627,44 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
...
+			clone := clone.DeepCopy()
+			addresses = append(addresses, initAddresses...)
+			clone.Status.Addresses = buildAddressStatus(addresses, nodeName)
+			if _, err := c.controlMachineClient.Machines(clone.Namespace).UpdateStatus(ctx, clone, metav1.UpdateOptions{}); err != nil {
+				return 0, fmt.Errorf("failed to persist status addresse after initialization was successfull: %w", err)
```
timebertt:
```suggestion
				return 0, fmt.Errorf("failed to persist status addresses after initialization was successful: %w", err)
```

timebertt (commented):
Nice PR! I found only two typos in my review :)

pkg/apis/machine/types.go
```diff
@@ -162,6 +162,11 @@ type CurrentStatus struct {
 
 // MachineStatus holds the most recently observed status of Machine.
 type MachineStatus struct {
+	// Addresses of this machines. This field is only present if the MCM provider runs without a target cluster and may
+	// be used by clients to determine how to connect to the machine, instead of the `Node.status.addresses` field.
+	// +optional
+	Addresses []corev1.NodeAddress `json:"addresses,omitempty"`
```
aaronfern: Is the json tag really required here?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -619,26 +627,44 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
...
 		nodeName = getMachineStatusResponse.NodeName
 		providerID = getMachineStatusResponse.ProviderID
+		addresses = append(addresses, getMachineStatusResponse.Addresses...)
 	}
+
 	//Update labels, providerID
 	var clone *v1alpha1.Machine
 	clone, err = c.updateLabels(ctx, createMachineRequest.Machine, nodeName, providerID)
+	if err != nil {
+		klog.V(2).Infof("failed to update labels and providerID for machine %q. err=%q", machine.Name, err.Error())
```
aaronfern:
```suggestion
		klog.Errorf("failed to update labels and providerID for machine %q. err=%q", machine.Name, err.Error())
```

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -619,26 +627,44 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
...
 			return retryPeriod, err
 		}
+
+		if c.targetCoreClient == nil {
+			// persist addresses from the InitializeMachine and CreateMachine responses
+			clone := clone.DeepCopy()
+			addresses = append(addresses, initAddresses...)
+			clone.Status.Addresses = buildAddressStatus(addresses, nodeName)
+			if _, err := c.controlMachineClient.Machines(clone.Namespace).UpdateStatus(ctx, clone, metav1.UpdateOptions{}); err != nil {
+				return 0, fmt.Errorf("failed to persist status addresses after initialization was successful: %w", err)
```
aaronfern:
```suggestion
				return machineutils.ShortRetry, fmt.Errorf("failed to persist status addresses after initialization was successful: %w", err)
```

aaronfern (commented):
Thanks for the PR @maboehm!
Just a few comments from me

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -485,13 +486,19 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
...
 		// Initializations
 		machine              = createMachineRequest.Machine
 		machineName          = createMachineRequest.Machine.Name
 		uninitializedMachine = false
 	)
 
+	// This field is only modified during the creation flow. We have to assume that every Address that was added to the
+	// status in the past remains valid, otherwise we have no way of keeping the addresses that might be only returned
+	// by the `CreateMachine` or `InitializeMachine` calls. Before persisting, the addresses are deduplicated.
+	addresses = append(addresses, createMachineRequest.Machine.Status.Addresses...)
```
aaronfern: Why is this appended? this can be an simple assignment

pkg/util/provider/driver/driver.go
```diff
@@ -59,6 +59,10 @@ type CreateMachineResponse struct {
 
 	// LastKnownState represents the last state of the VM during an creation/deletion error
 	LastKnownState string
+
+	// Addresses to reach the VM. Returning this field is optional, and only used if the MCM provider runs without a
+	// target cluster.
+	Addresses []corev1.NodeAddress
```
aaronfern, line 65: Why is the response being changed in `CreateMachineResponse`, `InitializeMachineResponse`, and `GetMachineStatusResponse`? Can these hostnames be changed arbitrarily within a nodes lifetime?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -485,13 +486,19 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 	var (
 		// Declarations
 		nodeName, providerID string
+		addresses            []corev1.NodeAddress
```
aaronfern: Can you use a map of type map[corev1.NodeAddress]struct{} here? would help with avoiding having to explicitly deduplicate

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -811,3 +838,18 @@ func (c *controller) triggerDeletionFlow(ctx context.Context, deleteMachineReque
...
 }
+
+// buildAddressStatus adds the nodeName as a HostName address, if it is not empty, and returns a sorted and deduplicated
+// slice.
+func buildAddressStatus(addresses []corev1.NodeAddress, nodeName string) []corev1.NodeAddress {
+	if nodeName != "" {
+		addresses = append(addresses, corev1.NodeAddress{
+			Type:    corev1.NodeHostName,
+			Address: nodeName,
+		})
```
aaronfern: The assumption that node name is always the hostname need not always be true. This can be overridden by a kubelet startup parameter (--hostname-override)
```
--hostname-override string
```
If non-empty, will use this string as identification instead of the actual hostname.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -704,7 +731,7 @@ func (c *controller) updateLabels(ctx context.Context, machine *v1alpha1.Machine
 	return clone, err
 }
 
-func (c *controller) initializeMachine(ctx context.Context, machine *v1alpha1.Machine, machineClass *v1alpha1.MachineClass, secret *corev1.Secret) (machineutils.RetryPeriod, error) {
+func (c *controller) initializeMachine(ctx context.Context, machine *v1alpha1.Machine, machineClass *v1alpha1.MachineClass, secret *corev1.Secret) (retry machineutils.RetryPeriod, addresses []corev1.NodeAddress, err error) {
```
aaronfern: Why not return an `InitializeMachineResponse` here, now that there is a need to consume response data from the caller of this method?

pkg/util/provider/machinecontroller/machine.go
maboehm: mhm, in this case there is no kubelet involved. We are basically assuming that within the network the `nodeName` can be resolved to the nodeIP - which is likely true in all environments.

Since this code is only relevant in the context of GEP-28, I would say we can make that assumption for now. If other users end up having vastly different setups, they can use the IP addresses that are also returned, or we can revisit this topic.

pkg/util/provider/machinecontroller/machine.go
maboehm: not sure if this made it better, but implemented it.

pkg/util/provider/machinecontroller/machine.go
maboehm: makes no difference to me

pkg/util/provider/driver/driver.go
maboehm: I mean they all return mostly the same data already...
Providers can still chose not to implement these fields, so I dont think they will hurt, and we wanted to avoid making assumptions.
I guess `CreateMachineResponse` and `InitializeMachineResponse` MUST have it (at least I understood that the initialization step was introduced specifically for some IPV6 edge cases in provider-aws https://github.com/gardener/machine-controller-manager/issues/871) - and at this point adding this to `GetMachineStatusResponse` is not semantically wrong, even though it might not change - the `NodeName` also does not change / should not change, right?

pkg/util/provider/machinecontroller/machine.go
aaronfern: Okay, we can accept this for now and revisit this later on if this assumption is broken.

pkg/util/provider/machinecontroller/machine.go
aaronfern: Sorry about this, but we had a late realization that we can use the generic apimachinery.Sets package (new feature). This will allow removing the `appendAddresses()` func and make the code more easier to read.

If this is too much of a change for you, we can make this change and push it to the PR

maboehm: ahh I didnt know about this package, I only knew about k8s.io/utils/set, which supports fewer types...
changed it :)

pkg/util/provider/driver/driver.go
maboehm: @aaronfern Are you fine with this explanation?

aaronfern: @maboehm Yes, this is fine

PR #1017, opened 2025-07-18
Add taint for critical components not ready after update

acumino:
After a successful in-place update, the controller now adds the 'critical components not ready' taint to the node. This prevents pods from being scheduled until critical component pods are ready, improving node readiness handling.

**Release note**:

```other operator
`node.gardener.cloud/critical-components-not-ready` taint is added to the node after the successful in-place update to prevent scheduling any workload before critical component pods are ready.
```

### DIFF
--- a/pkg/controller/deployment_inplace.go
+++ b/pkg/controller/deployment_inplace.go
@@ -193,6 +193,13 @@ func (dc *controller) syncMachineSets(ctx context.Context, oldMachineSets []*v1a
 		// uncordon the node since the inplace update is successful.
 		node.Spec.Unschedulable = false
 
+		// add the critical components not ready taint to the node this is to ensure that
+		// the pods are not scheduled on the node until the critical components pods are ready.
+		node.Spec.Taints = append(node.Spec.Taints, v1.Taint{
+			Key:    machineutils.TaintNodeCriticalComponentsNotReady,
+			Effect: v1.TaintEffectNoSchedule,
+		})
+
 		_, err = dc.targetCoreClient.CoreV1().Nodes().Update(ctx, node, metav1.UpdateOptions{})
 		if err != nil {
 			return fmt.Errorf("failed to remove inplace labels/annotations and uncordon node %s: %w", node.Name, err)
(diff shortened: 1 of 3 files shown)


pkg/controller/deployment_inplace.go
```diff
@@ -193,6 +194,18 @@ func (dc *controller) syncMachineSets(ctx context.Context, oldMachineSets []*v1a
 		// uncordon the node since the inplace update is successful.
 		node.Spec.Unschedulable = false
 
+		// remove the PreferNoSchedule taint if it exists which was added during the inplace update.
+		node.Spec.Taints = slices.DeleteFunc(node.Spec.Taints, func(t v1.Taint) bool {
+			return t.Key == PreferNoScheduleKey && t.Value == "True" && t.Effect == v1.TaintEffectPreferNoSchedule
+		})
```
aaronfern, line 200: Why is this taint not required anymore?

pkg/controller/deployment_inplace.go
```diff
@@ -193,6 +194,18 @@ func (dc *controller) syncMachineSets(ctx context.Context, oldMachineSets []*v1a
...
+
+		// add the critical components not ready taint to the node this is to ensure that
```
aaronfern:
```suggestion
		// add the critical components not ready taint to the node. This is to ensure that
```

pkg/controller/deployment_inplace.go
```diff
@@ -193,6 +194,18 @@ func (dc *controller) syncMachineSets(ctx context.Context, oldMachineSets []*v1a
...
+		// the pods are not scheduled on the node until the critical components pods are ready.
```
aaronfern:
```suggestion
		// workload pods are not scheduled on the node until the critical components pods are ready.
```

pkg/controller/deployment_inplace.go
acumino: This is added at the start of the in-place update. The machine that will go through the update, pod should preferably not be scheduled on that node.
Later, after the update is successful we need to remove it from the node

PR #1025, opened 2025-08-07
Fix `mcm_machine_set_failed_machines` metric to more accurately reflect current state

### LINKED ISSUE #476: MCM does not reset the failed_machines gauge once the machine is deleted
**What happened**:
When a machine is no longer reachable, MCM correctly reports the machine as unhealthy and creates the corresponding `failed_machines` gauge metric. However when the machine is actually deleted that metric is not cleared and it keeps on being reported.
This makes it problematic for alerting rules based on the `failed_machines` metric as it never gets reset to 0.
I may be easier to change this metric to a counter, so you don't have to worry about resetting its value, as keeping track of the failed machines may not be trivial or wanted.

```
I0616 23:12:14.072380       1 event.go:255] Event(v1.ObjectReference{Kind:"MachineSet", Namespace:"machine-controller-manager", Name:"mcm-immutable-node-az-b-6cb7bb5d9", UID:"31b3fc43-a668-11ea-b0fa-0a1dc5f423b0", APIVersion:"machine.sapcloud.io/v1alpha1", ResourceVersion:"371959992", FieldPath:""}): type: 'Normal' reason: 'SuccessfulDelete' Deleted machine: mcm-immutable-node-az-b-6cb7bb5d9-7zgk9
E0616 23:12:12.968331       1 machine.go:931] Machine mcm-immutable-node-az-b-6cb7bb5d9-7zgk9 is not healthy since 10m0s minutes. Changing status to failed. Node Conditions: [...]  {Type:Ready Status:Unknown LastHeartbeatTime:2020-06-16 22:59:55 +0000 UTC LastTransitionTime:2020-06-16 23:01:18 +0000 UTC Reason:NodeStatusUnknown Message:Kubelet stopped posting node status.}
W0616 23:01:38.218468       1 machine.go:729] Machine mcm-immutable-node-az-b-6cb7bb5d9-7zgk9 is unhealthy - changing MachineState to Unknown
```

`failed_machines` gauge metric should be reset once the machines is deleted (or recovers)

- Try creating a machine that doesn't bootstrap and watch the number of failed machines keep going up: `count(mcm_machine_deployment_failed_machines)`
- Create a machine, then delete the underlying VM and observe the value of `failed_machines`

**Environment**:
- kubernetes 1.12
- image: `eu.gcr.io/gardener-project/gardener/machine-controller-manager:v0.26.3`

**Interpretation/Solutions**

- We need a new gauge metric by the name `num_failed_machines` which tells the number of machines which are in `Failed` Phase currently
- `mcm_machine_deployment_operation_failed` gauge metric -> for the machines last operation failed
- gauge here means that the metric won't repeat, slice needs to be cleared once the machine is deleted

gagan16k:
This PR adds changes such that a gauge metric scraped on prometheus gets cleared and re-populated on every poll, so that it reflects the latest state of the reference slice `*machineSet.Status.FailedMachines`.
The logic to set this variable has also been updated so that it contains the current set of machines in the `Failed` state.

Fixes #476 

IT for mcm-provider-aws passed

Testing was done by setting the `machineCreationTimeout` to a value lesser than the time it takes for a machine to be provisioned and move into the `Running` state (1m). The machine moves to the `Failed` state, when it can be checked against a prometheus graph to verify if it was recorded accurately

Some changes were required to ensure a `Failed` machine stays failed, like an artificial delay in the goroutine marking a machine for deletion

image
Stacked graph reflecting number of failed machines accurately matched number in the cluster

**Release note**:

``` bugfix operator
Fixed metric `mcm_machine_set_failed_machines` and underlying variable `*machineSet.Status.FailedMachines` so that they reflect the current state of machines

```

### DIFF
--- a/pkg/controller/deployment_machineset_util.go
+++ b/pkg/controller/deployment_machineset_util.go
@@ -127,7 +127,7 @@ func calculateMachineSetStatus(is *v1alpha1.MachineSet, filteredMachines []*v1al
 	}
 
 	// Update the FailedMachines field when we see new failures
-	// Set the FailedMachines field to nil if there are no failed machines.
+	// Clear FailedMachines if there are no failed machines.
 	if len(failedMachines) > 0 {
 		newStatus.FailedMachines = &failedMachines
 	} else {
--- a/pkg/controller/deployment_machineset_util.go
+++ b/pkg/controller/deployment_machineset_util.go
@@ -126,13 +126,11 @@ func calculateMachineSetStatus(is *v1alpha1.MachineSet, filteredMachines []*v1al
 		}
 	}
 
-	// Update the FailedMachines field only if we see new failures
-	// Clear FailedMachines if ready replicas equals total replicas,
-	// which means the machineset doesn't have any machine objects which are in any failed state
-	// #nosec G115 -- number of machines will not exceed MaxInt32
+	// Update the FailedMachines field when we see new failures
+	// Set the FailedMachines field to nil if there are no failed machines.
 	if len(failedMachines) > 0 {
 		newStatus.FailedMachines = &failedMachines
-	} else if int32(readyReplicasCount) == is.Status.Replicas {
+	} else {
 		newStatus.FailedMachines = nil
 	}
 
--- a/pkg/controller/metrics.go
+++ b/pkg/controller/metrics.go
@@ -119,6 +119,11 @@ func updateMachineSetStatusRelatedMetric(machineSet *v1alpha1.MachineSet, msMeta
 }
 
 func updateMachineSetStatusFailedMachinesMetric(machineSet *v1alpha1.MachineSet, msMeta metav1.ObjectMeta) {
+	metrics.MachineSetStatusFailedMachines.DeletePartialMatch(prometheus.Labels{
+		"name":      msMeta.Name,
+		"namespace": msMeta.Namespace,
+	})
+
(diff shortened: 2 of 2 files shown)


takoverflow (approved):
LGTM
Please update the release notes (you can check other PRs for inspiration)

pkg/controller/deployment_machineset_util.go
```diff
@@ -126,13 +126,11 @@ func calculateMachineSetStatus(is *v1alpha1.MachineSet, filteredMachines []*v1al
 		}
 	}
 
-	// Update the FailedMachines field only if we see new failures
-	// Clear FailedMachines if ready replicas equals total replicas,
-	// which means the machineset doesn't have any machine objects which are in any failed state
-	// #nosec G115 -- number of machines will not exceed MaxInt32
+	// Update the FailedMachines field when we see new failures
+	// Set the FailedMachines field to nil if there are no failed machines.
```
aaronfern: nit
```suggestion
	// Clear FailedMachines if there are no failed machines.
```

aaronfern (approved):
Looks good to me!
Just a small nit

pkg/controller/deployment_machineset_util.go
gagan16k: Committed changes

aaronfern (approved):

PR #1036, opened 2025-10-10
Safeguard to prevent termination of machines part of creation flow

### LINKED ISSUE #1035: Machine object allowed to be added to the deletion queue when it is being processed by the creation flow

/area control-plane
/kind bug
/priority 1

**What happened**:
We recently discovered a bug where a machine object can be added to the deletion queue while it is still being processed by the creation flow.

This leads to a race condition: the deletion flow proceeds to delete the machine and its corresponding node object even though the creation flow is still waiting for the CSP to provision the VM. Once the VM is eventually created, the machine creation flow completes successfully—but the resulting VM is left orphaned, since its associated machine and node objects have already been deleted.

A machine should never be enqueued in both the creation and deletion queues at the same time.

We already have a safeguard in place that prevents a machine from being added to the creation queue if it has a deletion timestamp.
Now, we need to introduce a similar check to ensure that a machine cannot be enqueued into the deletion queue if it is currently being processed by the creation flow.

**Environment**:

- Others:

takoverflow:
This PR adds a `sync.Map` that consists of list of all machines that are
being processed by the creation flow. This is used to decide whether a
machine that's scheduled for deletion (has its `DeletionTimestamp` set)
should be processed by the deletion flow. This helps avoid scenarios where
the machine object is removed while a `CreateMachine()` call is still being
processed by the Cloud provider, leading to orphan (leaked) VMs.

There's still an edge case where a machine is marked for deletion and is
processed by the creation flow but MCM crashes. Upon restart, since the map
would be empty, the machine would go to the termination queue and get removed
all the while the CSP creates a VM for the same removed machine leading to an orphan VM.

Fixes #1035

Gingko tests hamper readability quite a lot, I'd request focusing on
`action.testFunc` which contains the crux of the functionality being tested.
Open to suggestions for alternative names.

**Release note**:

```bugfix operator
Added a safeguard to delay deletion of machines that are undergoing a `Create` Request to prevent orphaning of VMs.
```

### DIFF
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -83,7 +83,7 @@ func (c *controller) deleteMachine(obj interface{}) {
 		}
 		machine, ok = tombstone.Obj.(*v1alpha1.Machine)
 		if !ok {
-			utilruntime.HandleError(fmt.Errorf("Tombstone contained object that is not a Machine %#v", obj))
+			utilruntime.HandleError(fmt.Errorf("tombstone contained object that is not a Machine %#v", obj))
 			return
 		}
 	}
@@ -794,7 +794,7 @@ func (c *controller) triggerDeletionFlow(ctx context.Context, deleteMachineReque
 
 	switch {
 	case isMachineInCreationFlow:
-		err := fmt.Errorf("Machine %q is in creation flow. Deletion cannot proceed", machine.Name)
+		err := fmt.Errorf("machine %q is in creation flow. Deletion cannot proceed", machine.Name)
 		return machineutils.MediumRetry, err
 
 	case !finalizers.Has(MCMFinalizerName):
@@ -874,7 +874,7 @@ func (c *controller) shouldMachineBeMovedToTerminatingQueue(machine *v1alpha1.Ma
 	_, isMachineInCreationFlow := c.pendingMachineCreationMap.Load(machine.Name)
 
 	if machine.DeletionTimestamp != nil && isMachineInCreationFlow {
-		klog.Infof("Cannot delete machine %q, its deletionTimestamp is set but it is currently being processed by the creation flow\n", machine.Name)
+		klog.Warningf("Cannot delete machine %q, its deletionTimestamp is set but it is currently being processed by the creation flow\n", machine.Name)
 	}
 
 	return !isMachineInCreationFlow && machine.DeletionTimestamp != nil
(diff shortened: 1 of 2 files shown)


pkg/util/provider/machinecontroller/machine.go
```diff
@@ -83,7 +83,7 @@ func (c *controller) deleteMachine(obj interface{}) {
 		}
 		machine, ok = tombstone.Obj.(*v1alpha1.Machine)
 		if !ok {
-			utilruntime.HandleError(fmt.Errorf("Tombstone contained object that is not a Machine Deployment %#v", obj))
+			utilruntime.HandleError(fmt.Errorf("Tombstone contained object that is not a Machine %#v", obj))
```
aaronfern: nit
```suggestion
			utilruntime.HandleError(fmt.Errorf("tombstone contained object that is not a Machine %#v", obj))
```

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -780,11 +787,16 @@ func (c *controller) initializeMachine(ctx context.Context, machine *v1alpha1.Ma
...
-		machine    = deleteMachineRequest.Machine
-		finalizers = sets.NewString(machine.Finalizers...)
+		machine                    = deleteMachineRequest.Machine
+		finalizers                 = sets.NewString(machine.Finalizers...)
+		_, isMachineInCreationFlow = c.pendingMachineCreationMap.Load(machine.Name)
 	)
 
 	switch {
+	case isMachineInCreationFlow:
+		err := fmt.Errorf("Machine %q is in creation flow. Deletion cannot proceed", machine.Name)
```
aaronfern: nit
```suggestion
		err := fmt.Errorf("machine %q is in creation flow. Deletion cannot proceed", machine.Name)
```

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -857,3 +869,13 @@ func buildAddressStatus(addresses sets.Set[corev1.NodeAddress], nodeName string)
 	})
 	return res
 }
+
+func (c *controller) shouldMachineBeMovedToTerminatingQueue(machine *v1alpha1.Machine) bool {
+	_, isMachineInCreationFlow := c.pendingMachineCreationMap.Load(machine.Name)
+
+	if machine.DeletionTimestamp != nil && isMachineInCreationFlow {
+		klog.Infof("Cannot delete machine %q, its deletionTimestamp is set but it is currently being processed by the creation flow\n", machine.Name)
```
aaronfern: Should this be a warning instead? WDYT?

takoverflow: Makes sense, turned it into a warning.

PR #1043, opened 2025-10-15
Use fully qualified machine name for pending create map

takoverflow:
- Use machine `NamespacedName` when populating/checking `pendingMachineCreationMap`
- Populate the map with the machine name inside `triggerCreationFlow`

Fixes #

cc: @aaronfern @elankath 

**Release note**:

```other operator

```

### DIFF
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -495,10 +495,9 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 		uninitializedMachine = false
 		addresses            = sets.New[corev1.NodeAddress]()
 	)
-	machineNamespacedName := getMachineKey(machine)
-	c.pendingMachineCreationMap.Store(machineNamespacedName, "")
+	c.markCreationProcessing(machine)
 	defer func() {
-		c.pendingMachineCreationMap.Delete(machineNamespacedName)
+		c.unmarkCreationProcessing(machine)
 	}()
 
 	// This field is only modified during the creation flow. We have to assume that every Address that was added to the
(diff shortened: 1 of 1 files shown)


elankath:
Whyy can't we see the log for `err := fmt.Errorf("machine %q is in creation flow. Deletion cannot proceed", machine.Name)` ? Is it possible to add the log for this case ? (So that it is clear from logs that deletion is postponed until creation is done)

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -501,6 +495,11 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 		uninitializedMachine = false
 		addresses            = sets.New[corev1.NodeAddress]()
 	)
+	machineNamespacedName := getMachineKey(machine)
```
aaronfern: Can you move this up to the var block?

aaronfern: Is it not possible to use `c.getKeyForObj()` here instead?

takoverflow: In case `MetaNamespaceKeyFunc` fails for `getKeyForObj`, it returns `""` name which can't be used to differentiate different machines stored in the map, hence in the `getMachineKey()` utility function, I manually create it using `Namespace/Name` upon failure (as a fallback).

aaronfern: Yes, but the `getKeyForObj` would return an error when the object does not have a metadata
In that case also, manually creating `Namespace/Name` should also fail, shouldn't it?

takoverflow:
> Whyy can't we see the log for err := fmt.Errorf("machine %q is in creation flow. Deletion cannot proceed", machine.Name) ? Is it possible to add the log for this case ? (So that it is clear from logs that deletion is postponed until creation is done)

The error log would only show if the machine is actually being processed by `machineTerminationQueue` and the check in `updateMachine` would prevent enqueuing a machine there if it's undergoing creation as discussed.

pkg/util/provider/machinecontroller/machine.go
takoverflow: That's valid but avoiding adding a generic identifier such as `""` didn't make much sense if fetching the key errors out.

aaronfern (approved):

PR #1049, opened 2025-11-07
Changing MCM queues to rate-limiting queues

### LINKED ISSUE #1030: Implement exponential backoff and do not update MCD, MS on same error

/area control-plane
/kind bug
/priority 1

**What happened**:

We had a live, scalability issue where due to invalid credentials, the `etcd` database was filled up.
The `machine-controller-manager` was continuing updating `MachineDeployments` and `MachineSets`. The `MachineDeployment` status contains entry for each `Machine` and its `lastError`. 

(issues-canary/issues/7190 internally)

- `machine-controller-manager` should adhere to controller best practices such as exponential backoff and skipping no-op (status) updates if there is no change in the status.

- Use the virtual mcm provider and local api-server and etcd to simulate credential failure for large number of machines (> 1000)
- Check size of etcd db.

r4mek:
This PR is part of the solution for the issue #1030. In this PR we implement a rate-limiting queue for error scenarios where we want to requeue the machine.

Fixes #1030 (partly)

**Release note**:

```other operator

```

### DIFF
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -183,14 +183,14 @@ func (c *controller) reconcileClusterMachineKey(key string) error {
 	}
 
 	err = c.reconcileClusterMachine(ctx, machine)
-	if err == nil {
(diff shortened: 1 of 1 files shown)


r4mek:
No longer required. Raised a different PR #1056

PR #1050, opened 2025-11-11
Check `providerID` instead of node label in creation flow

maboehm:
When running without a target cluster, the mcm intentionally never sets the `node` label on the `Machine` object. So this check makes no sense. It is better to check if the providerID was set in the spec. Both fields are set at the same time in `r.updateLabels()`, but only `providerID` is set in all scenarios.

When running without a target cluster and with a provider that does not implement `GetMachineStatus()`, the controller would be in an infinite loop, because this label is never set - and without the label always calls `CreateMachine()` and returns.

Essentially a follow up to https://github.com/gardener/machine-controller-manager/pull/1004

**Release note**:

```other operator

```

### DIFF
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -612,7 +612,7 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 				}
 				uninitializedMachine = true
 			} else {
-				// if node label present that means there must be a backing VM ,without need of GetMachineStatus() call
+				// if .spec.ProviderID is present that means there must be a backing VM, without need of GetMachineStatus() call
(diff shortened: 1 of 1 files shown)


maboehm:
@aaronfern could you please take a look at this mini change?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -614,6 +614,7 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 			} else {
 				// if node label present that means there must be a backing VM ,without need of GetMachineStatus() call
```
aaronfern: Nit, but can you please update this comment to be more accurate gives this change?

aaronfern (commented):
Thanks for this PR!
Change looks fine to me. Just one minor nit. Please address that

PR #1052, opened 2025-11-17
Fix scaling logic in reconcileNewMachineSetInPlace

acumino:
Adds a check to ensure the new machine set is scaled up when the combined replicas of old and new machine sets are less than the desired deployment replicas. This basically ensures that if a machine is deleted manually from the old machine sets, a new machine should be created in the new machine set with the updated configuration.

Fixes #

**Release note**:

```bugfix operator
Fixed an issue where deleting a machine from an old MachineSet could prevent proper scaling. The new MachineSet now scales up when total replicas (old + new) drop below the desired count, ensuring replacements are created with the updated configuration.
```

### DIFF
--- a/pkg/controller/deployment_inplace.go
+++ b/pkg/controller/deployment_inplace.go
@@ -262,7 +262,7 @@ func (dc *controller) reconcileNewMachineSetInPlace(ctx context.Context, oldMach
 	if totalReplicas < deployment.Spec.Replicas {
 		// Scale up the new machine set to reach the desired replica count, considering old machines.
 		klog.V(3).Infof("scale up the new machine set %s from %d to %d replicas (delta: %d)", newMachineSet.Name, newMachineSet.Spec.Replicas, deployment.Spec.Replicas-oldMachinesCount, deployment.Spec.Replicas-totalReplicas)
-		scaled, _, err := dc.scaleMachineSetAndRecordEvent(ctx, newMachineSet, totalReplicas, deployment)
+		scaled, _, err := dc.scaleMachineSetAndRecordEvent(ctx, newMachineSet, deployment.Spec.Replicas-oldMachinesCount, deployment)
 		return scaled, err
 	}
 


pkg/controller/deployment_inplace.go
```diff
@@ -258,6 +258,13 @@ func (dc *controller) reconcileNewMachineSetInPlace(ctx context.Context, oldMach
 		return scaled, err
 	}
 
+	if oldMachinesCount+newMachineSet.Spec.Replicas < deployment.Spec.Replicas {
+		// Scale up the new machine set to the maximum possible replicas.
```
aaronfern: This term "maximum possible replicas" could be misleading. Maybe change it to say "deployment replicas"

pkg/controller/deployment_inplace.go
```diff
@@ -258,6 +258,13 @@ func (dc *controller) reconcileNewMachineSetInPlace(ctx context.Context, oldMach
...
+		klog.V(3).Infof("scale up the new machine set %s by %d to %d replicas", newMachineSet.Name, (deployment.Spec.Replicas - (oldMachinesCount + newMachineSet.Spec.Replicas)), deployment.Spec.Replicas-oldMachinesCount)
```
aaronfern: This log can be improved a bit. Can you update this to be like the following (feel free to instead add a log line you feel is better suited)
```
scaling up new machine set <name> from y to z (delta: x)
```

elankath: Please also assign the expression `oldMachinesCount + newMachineSet.Spec.Replicas` to variable `desiredReplicas` to avoid excessive repeat of this expression.

pkg/controller/deployment_inplace.go
```diff
@@ -258,6 +258,14 @@ func (dc *controller) reconcileNewMachineSetInPlace(ctx context.Context, oldMach
 		return scaled, err
 	}
 
+	totalReplicas := oldMachinesCount + newMachineSet.Spec.Replicas
+	if totalReplicas < deployment.Spec.Replicas {
```
takoverflow, line 262: Just wanted to understand the need for the change, if `totalReplicas` are less than desired, when the transfer from old machineset to new machineset is done and `oldMachinesCount` above is 0, the `newMachineSet` would've been scaled to the desired replicas anyways right? Am I missing something

Is this just an optimization to speed up the process of scaling of the new machineset?

takoverflow: From offline discussion:

If machine from old set is explicitly removed, consider the scenario with `desired` being `6`, old machines now being `2` and machines in new set being `3` with `maxUnavailable` as `1`.

Then [this check](https://github.com/gardener/machine-controller-manager/blob/84dcfb5cf8c1c6790e2e5fe3edad97eeee064a15/pkg/controller/deployment_inplace.go#L328)

```go
    maxUpdatePossible := allMachinesCount - minAvailable - newMachineSetUnavailableMachineCount - oldMachineSetsMachinesUndergoingUpdate
    if maxUpdatePossible <= 0 {
        klog.V(3).Infof("no machines can be selected for update from old machine sets")
        return nil
    }
```
Now `allMachinesCount` is `5` (2 old + 3 new) and `minAvailable = desired(6) - maxUnavailable(1) = 5` so `maxUpdatePossible` is `0` and the check results in no machines from the old set being picked for in-place update and the process comes to a halt.

The fix above allows for a new machine to be brought up in the new machineSet thus making `maxUpdatePossible` as `1` and resulting in one more machine from the old set to be processed for in-place update.

takoverflow (approved):
Thanks for the fix! LGTM

PR #1054, opened 2025-11-25
Use `ToBeDeletedByClusterAutoscaler` Taint to improve load balancing during machine terminations

dergeberl:

If a machine is deleted the it is possible that the load balancer do not know yet that the node is not available anymore. This results eventually in unsuccessful new connections as the node where the load balancer is forwarding the traffic is not existing anymore.

There are 2 ways a load balancer gets informed: 1. Via a reconcile of the load balancer in the `cloud-provider` package or the `kube-proxy` health check. Unfortunately the "drain" or conditions set by the MCM dose not trigger a reconcile or the health check to fail. 

The `ToBeDeletedByClusterAutoscaler` Taint is used in both [`cloud-provider`](https://github.com/kubernetes/cloud-provider/blob/080e91c4b910bf92dfd43ca79eb74f5d39dcba75/controllers/service/controller.go#L1027) and[ `kube-proxy`](https://github.com/kubernetes/kubernetes/blob/5bcb7599736327cd8c6d23e398002354a6e40f68/pkg/proxy/healthcheck/proxy_health.go#L187), therefore we want to add this Taint to improve the load balancing during machine terminations.

This is new step is added between `Initiate node drain`/`Delete Volume Attachments` and `Initiate VM deletion` with one additional `ShortRetry` (5s) to give the  `cloud-provider` and `kube-proxy` time to react before removing the server.

CC: @kon-angelo 

Fixes #

**Release note**:

```other operator
Use `ToBeDeletedByClusterAutoscaler` Taint to improve load balancing during machine terminations.
```

### DIFF
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -641,6 +641,9 @@ func (c *controller) triggerDeletionFlow(ctx context.Context, deleteMachineReque
 	case strings.Contains(machine.Status.LastOperation.Description, machineutils.InitiateDrain):
 		return c.drainNode(ctx, deleteMachineRequest)
 
+	case strings.Contains(machine.Status.LastOperation.Description, machineutils.SetDeletionTaint):
+		return c.taintNode(ctx, deleteMachineRequest)
+
 	case strings.Contains(machine.Status.LastOperation.Description, machineutils.DelVolumesAttachments):
 		return c.deleteNodeVolAttachments(ctx, deleteMachineRequest)
 
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -641,11 +641,11 @@ func (c *controller) triggerDeletionFlow(ctx context.Context, deleteMachineReque
 	case strings.Contains(machine.Status.LastOperation.Description, machineutils.InitiateDrain):
 		return c.drainNode(ctx, deleteMachineRequest)
 
-	case strings.Contains(machine.Status.LastOperation.Description, machineutils.SetDeletionTaint):
-		return c.taintNode(ctx, deleteMachineRequest)
-
 	case strings.Contains(machine.Status.LastOperation.Description, machineutils.DelVolumesAttachments):
 		return c.deleteNodeVolAttachments(ctx, deleteMachineRequest)
+
+	case strings.Contains(machine.Status.LastOperation.Description, machineutils.SetDeletionTaint):
+		return c.taintNode(ctx, deleteMachineRequest)
 
 	case strings.Contains(machine.Status.LastOperation.Description, machineutils.InitiateVMDeletion):
 		return c.deleteVM(ctx, deleteMachineRequest)
--- a/pkg/util/provider/machinecontroller/machine_test.go
+++ b/pkg/util/provider/machinecontroller/machine_test.go
@@ -1949,7 +1949,7 @@ var _ = Describe("machine", func() {
 					},
 				},
 				expect: expect{
-					err:                           fmt.Errorf("Drain successful. %s", machineutils.InitiateVMDeletion),
+					err:                           fmt.Errorf("Drain successful. %s", machineutils.SetDeletionTaint),
 					retry:                         machineutils.ShortRetry,
 					nodeTerminationConditionIsSet: true,
 					machine: newMachine(
@@ -1969,7 +1969,7 @@ var _ = Describe("machine", func() {
 								LastUpdateTime: metav1.Now(),
 							},
 							LastOperation: v1alpha1.LastOperation{
-								Description:    fmt.Sprintf("Drain successful. %s", machineutils.InitiateVMDeletion),
+								Description:    fmt.Sprintf("Drain successful. %s", machineutils.SetDeletionTaint),
 								State:          v1alpha1.MachineStateProcessing,
 								Type:           v1alpha1.MachineOperationDelete,
 								LastUpdateTime: metav1.Now(),
@@ -2465,7 +2465,7 @@ var _ = Describe("machine", func() {
 					},
 				},
 				expect: expect{
-					err:   fmt.Errorf("Drain successful. %s", machineutils.InitiateVMDeletion),
+					err:   fmt.Errorf("Drain successful. %s", machineutils.SetDeletionTaint),
 					retry: machineutils.ShortRetry,
 					machine: newMachine(
 						&v1alpha1.MachineTemplateSpec{
@@ -2484,7 +2484,7 @@ var _ = Describe("machine", func() {
 								LastUpdateTime: metav1.Now(),
 							},
 							LastOperation: v1alpha1.LastOperation{
-								Description:    fmt.Sprintf("Drain successful. %s", machineutils.InitiateVMDeletion),
+								Description:    fmt.Sprintf("Drain successful. %s", machineutils.SetDeletionTaint),
 								State:          v1alpha1.MachineStateProcessing,
 								Type:           v1alpha1.MachineOperationDelete,
 								LastUpdateTime: metav1.Now(),
@@ -2567,7 +2567,7 @@ var _ = Describe("machine", func() {
 					},
 				},
 				expect: expect{
-					err:   fmt.Errorf("Drain successful. %s", machineutils.InitiateVMDeletion),
+					err:   fmt.Errorf("Drain successful. %s", machineutils.SetDeletionTaint),
 					retry: machineutils.ShortRetry,
 					machine: newMachine(
 						&v1alpha1.MachineTemplateSpec{
@@ -2586,7 +2586,7 @@ var _ = Describe("machine", func() {
 								LastUpdateTime: metav1.Now(),
 							},
 							LastOperation: v1alpha1.LastOperation{
-								Description:    fmt.Sprintf("Drain successful. %s", machineutils.InitiateVMDeletion),
+								Description:    fmt.Sprintf("Drain successful. %s", machineutils.SetDeletionTaint),
 								State:          v1alpha1.MachineStateProcessing,
 								Type:           v1alpha1.MachineOperationDelete,
 								LastUpdateTime: metav1.Now(),
@@ -3064,7 +3064,7 @@ var _ = Describe("machine", func() {
 								LastUpdateTime: metav1.Now(),
 							},
 							LastOperation: v1alpha1.LastOperation{
-								Description:    fmt.Sprintf("Drain successful. %s", machineutils.InitiateVMDeletion),
+								Description:    fmt.Sprintf("Node tainted. %s", machineutils.InitiateVMDeletion),
 								State:          v1alpha1.MachineStateProcessing,
 								Type:           v1alpha1.MachineOperationDelete,
 								LastUpdateTime: metav1.Now(),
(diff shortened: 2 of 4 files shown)


pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1742,6 +1742,67 @@ func (c *controller) drainNode(ctx context.Context, deleteMachineRequest *driver
...
+			Effect: v1.TaintEffectPreferNoSchedule,
+		}
+		description     = ""
+		taintAlreadySet = false
+		skipStep        = false
+	)
+	node, err := c.nodeLister.Get(getNodeName(machine))
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			klog.Errorf("Error occurred while trying to fetch node object - err: %s", err)
```
aaronfern: nit, since go convention is that errors should not be capitalised, also %v is a more appropriate formatter for errors.
I understand that the rest of the file contains many examples of exactly this, but we will correct this soon for the other places
```suggestion
			klog.Errorf("error occurred while trying to fetch node object - err: %v", err)
```

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1742,6 +1742,67 @@ func (c *controller) drainNode(ctx context.Context, deleteMachineRequest *driver
...
+			machine,
+			v1alpha1.LastOperation{
+				Description:    description,
+				State:          v1alpha1.MachineStateProcessing,
+				Type:           v1alpha1.MachineOperationDelete,
+				LastUpdateTime: metav1.Now(),
+			},
+			// Let the clone.Status.CurrentStatus (LastUpdateTime) be as it was before.
+			// This helps while computing when the drain timeout to determine if force deletion is to be triggered.
+			// Ref - https://github.com/gardener/machine-controller-manager/blob/rel-v0.34.0/pkg/util/provider/machinecontroller/machine_util.go#L872
```
aaronfern: This ref link does not exist. Could you please correct it?

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1742,6 +1742,67 @@ func (c *controller) drainNode(ctx context.Context, deleteMachineRequest *driver
...
+			machine.Status.CurrentStatus,
+			machine.Status.LastKnownState,
+		)
+	}
+
+	node.Spec.Taints = append(node.Spec.Taints, toBeDeletedTaint)
```
aaronfern: This could potentially cause duplicate taints on the node. Do you think this could cause an issue?
Maybe consider using [`AddOrUpdateTaint()`](https://github.com/gardener/machine-controller-manager/blob/master/pkg/util/taints/taints.go#L268) in that case

aaronfern (changes requested):

pkg/util/provider/machinecontroller/machine_util.go
dergeberl: TBH I just copied this link. I hope I found the correct reference. I updated all places in that file.

pkg/util/provider/machinecontroller/machine_util.go
dergeberl: I changed the check above if the taint already exists to only check for the `key`. This we would not change the taint if the cluster autoscaler already set the taint (the `MatchTaint()` also checks the `effect`)

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -814,6 +814,9 @@ func (c *controller) triggerDeletionFlow(ctx context.Context, deleteMachineReque
 	case strings.Contains(machine.Status.LastOperation.Description, machineutils.InitiateDrain):
 		return c.drainNode(ctx, deleteMachineRequest)
 
+	case strings.Contains(machine.Status.LastOperation.Description, machineutils.SetDeletionTaint):
+		return c.taintNode(ctx, deleteMachineRequest)
+
 	case strings.Contains(machine.Status.LastOperation.Description, machineutils.DelVolumesAttachments):
 		return c.deleteNodeVolAttachments(ctx, deleteMachineRequest)
```
gagan16k:
```suggestion
	case strings.Contains(machine.Status.LastOperation.Description, machineutils.DelVolumesAttachments):
		return c.deleteNodeVolAttachments(ctx, deleteMachineRequest)

	case strings.Contains(machine.Status.LastOperation.Description, machineutils.SetDeletionTaint):
		return c.taintNode(ctx, deleteMachineRequest)
```

Could the order of these be swapped, as `taintNode` logically happens after `deleteNodeVolAttachments`?

Also could the actual definitions in `machine_util.go` be swapped as well, for better readability?

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1758,11 +1819,11 @@ func (c *controller) deleteNodeVolAttachments(ctx context.Context, deleteMachine
 			return retryPeriod, err
 		}
 		// node not found move to vm deletion
-		description = fmt.Sprintf("Skipping deleteNodeVolAttachments due to - %s. Moving to VM Deletion. %s", err.Error(), machineutils.InitiateVMDeletion)
+		description = fmt.Sprintf("Skipping deleteNodeVolAttachments due to - %s. Moving to VM Deletion. %s", err.Error(), machineutils.SetDeletionTaint)
```
gagan16k: Description text needs to be updated as well, in all instances of this change

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1742,6 +1742,67 @@ func (c *controller) drainNode(ctx context.Context, deleteMachineRequest *driver
...
+		}
+		skipStep = true
+		description = fmt.Sprintf("Node dose not exist. %s", machineutils.InitiateVMDeletion)
+	}
+
+	if node != nil {
+		for _, taint := range node.Spec.Taints {
+			if taint.Key == toBeDeletedTaint.Key {
+				taintAlreadySet = true
+				description = fmt.Sprintf("Node tainted. %s", machineutils.InitiateVMDeletion)
```
gagan16k: Can break early from the loop when taint is found.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1742,6 +1742,67 @@ func (c *controller) drainNode(ctx context.Context, deleteMachineRequest *driver
...
+		if !apierrors.IsNotFound(err) {
+			klog.Errorf("error occurred while trying to fetch node object - err: %v", err)
+			return machineutils.ShortRetry, err
+		}
+		skipStep = true
+		description = fmt.Sprintf("Node dose not exist. %s", machineutils.InitiateVMDeletion)
+	}
+
+	if node != nil {
+		for _, taint := range node.Spec.Taints {
```
gagan16k: Alternatively, [AddOrUpdateTaint()](https://github.com/gardener/machine-controller-manager/blob/master/pkg/util/taints/taints.go#L268) can be used like:
``` go
if node != nil {
	updatedNode, taintUpdated, _ := taintutils.AddOrUpdateTaint(node, &toBeDeletedTaint)
    if taintUpdated {
		taintAlreadySet = true
		description = fmt.Sprintf("Node tainted. %s", machineutils.InitiateVMDeletion)
	}
}
```

And then `updatedNode` can be directly used with `Update()` later on.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1742,6 +1742,67 @@ func (c *controller) drainNode(ctx context.Context, deleteMachineRequest *driver
...
+			if taint.Key == toBeDeletedTaint.Key {
+				taintAlreadySet = true
+				description = fmt.Sprintf("Node tainted. %s", machineutils.InitiateVMDeletion)
+			}
+		}
+	}
+
+	if taintAlreadySet || skipStep {
+		return c.machineStatusUpdate(
```
gagan16k: Since `machineStatusUpdate` already returns `ShortRetry`, the current implementation results in a cumulative 10s delay. Is this intentional?

If we need the extended delay, the structure could potentially be changed like so, to double the delay and do it in a single step.
``` go
updateRetryPeriod, updateErr := c.machineStatusUpdate(............)
if updateErr != nil {
  return 2*updateRetryPeriod, nil
 }
return updateRetryPeriod, updateErr
```

This would also ensure that machine status is properly updated with the current descriptions at each step

pkg/util/provider/machinecontroller/machine_test.go
```diff
@@ -3128,6 +3128,318 @@ var _ = Describe("machine", func() {
 					),
 				},
 			}),
+			Entry("Set ToBedeletedByClusterAutoscaler Taint", &data{
```
gagan16k, line 3236: This might need to be rebased, potential merge conflicts on this block due to recent commits

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1742,6 +1742,67 @@ func (c *controller) drainNode(ctx context.Context, deleteMachineRequest *driver
...
+		skipStep        = false
+	)
+	node, err := c.nodeLister.Get(getNodeName(machine))
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			klog.Errorf("error occurred while trying to fetch node object - err: %v", err)
+			return machineutils.ShortRetry, err
+		}
+		skipStep = true
+		description = fmt.Sprintf("Node dose not exist. %s", machineutils.InitiateVMDeletion)
```
gagan16k:
```suggestion
		description = fmt.Sprintf("Node does not exist. %s", machineutils.InitiateVMDeletion)
```
typo, test also needs to be updated

gagan16k (changes requested):
Thanks for the PR! Have suggested some changes, PTAL

pkg/util/provider/machinecontroller/machine_util.go
dergeberl: You mean because this return is only executed during the second run as in the first run `taintAlreadySet` is `false`. 

We could move that to the last return and handle it like this (also only increase the delay if the step was not skipped).

I think 5 seconds are to slow as the load balancer solution need to run a healthcheck in the time between annotate and vm deletion.

pkg/util/provider/machinecontroller/machine_test.go
dergeberl: I rebased which worked without merge conflicts. Also the test are still working.

aaronfern:
Hi @dergeberl 
After discussing this with the the team, I am still a bit hesitant to add this.
Would a better solution not be to have a generic label to the cloud-provider that could be used to exclude nodes from node balancing?

dergeberl:
As said in the comments above, we also can use the `node.kubernetes.io/exclude-from-external-load-balancers` label. This would reconcile the load balancers (if the `cloud-provider` package is used). But this is "against" the change in the `cloud-provider` package to reduce the number of API calls ([KEP-3458](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/3458-remove-transient-node-predicates-from-service-controller/README.md)). To relay on health checks we would need a way to make a node as unready (currently only `ToBeDeletedByClusterAutoscaler` taint is used). 

I would be also fine with the  `node.kubernetes.io/exclude-from-external-load-balancers` label solution. We currently maintain a [fork of the `cloud-provider`](https://github.com/stackitcloud/cloud-provider/commit/9ca02b64fa18933490d315c702db529e05002ca7) to include the `Terminating` condition set by the MCM (which also triggers a reconcile for load balancers). We really would like to remove the fork.

I also agree that some more generic label/taint would make sense. I remember multiple issues where this was also discussed but I could not find them now. So I had the feeling this could end up in never ending dicussions.

WDYT?

aaronfern:
Yes, the label solution seems more appropriate and would be more agreeable.

> I also agree that some more generic label/taint would make sense. I remember multiple issues where this was also discussed but I could not find them now. So I had the feeling this could end up in never ending dicussions.

We should open up another issue with kube-proxy to support this generic label/taint. Some additional noise might hopefully help nudge things in the right direction

fischerman:
> We should open up another issue with kube-proxy to support this generic label/taint. Some additional noise might hopefully help nudge things in the right direction

I agree. Depending on the load balancer implementation and configuration they behave different between a node removal and failing a health check. The externalTrafficPolicy also matters here.

What matters is how the load balancers performs draining. Some load balancers sever connections during node removal, others drain them. Envoys default behavior is even more counterintuitive. Even if the node was removed, as long as the health check continues to succeed new connections are routed to that node ([Envoy docs](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/service_discovery#on-eventually-consistent-service-discovery)). (You can change that in Envoy but I haven't tested either behavior.) 

Failing health checks usually just cause new connections to not be routed to that node but gives old connections the chance to close naturally. Other implementations might also close existing connections.

externalTrafficPolicy influences how connections are closed naturally during a node shutdown. With eTP=Cluster, because Kubernetes cannot close the connections that it NATs (to a node where the pod is), those connections silently break. Closing the connections on the load balancer might be the best choice here. The question is based on what: node removal and / or HC failure? For eTP=Local the pod shuts down before the node shuts down. So connections are actively closed and the load balancer should just wait for that. IMHO This is preferred because it can happen on the application layer rather than on TCP.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -644,6 +644,9 @@ func (c *controller) triggerDeletionFlow(ctx context.Context, deleteMachineReque
...
 
+	case strings.Contains(machine.Status.LastOperation.Description, machineutils.SetDeletionTaint):
+		return c.taintNode(ctx, deleteMachineRequest)
```
elankath, line 648: No need for separate stage. We should follow CA, where this taint is immediately set before draining the node. This can be moved to `c.drainNode`. No need for this explicit case and implementation can be made simpler. (This should not be done for `c.drainPreservedNode` or `c.drainNodeForInPlace`  as those Nodes will come back)

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1808,6 +1809,63 @@ func (c *controller) deleteNodeVolAttachments(ctx context.Context, deleteMachine
 	return retryPeriod, err
 }
 
+func (c *controller) taintNode(ctx context.Context, deleteMachineRequest *driver.DeleteMachineRequest) (machineutils.RetryPeriod, error) {
```
elankath, line 1812: This impl is now un-necessarily complex. If it will be called from `c.drainNode` before actual drain commences, then it can simply attempt to taint and on failure, just return the error. No need for `machineStatusUpdate`, `skipSet`, etc

PR #1055, opened 2025-11-28
Remove `node.machine.sapcloud.io/trigger-deletion-by-mcm` annotation

gagan16k:
**What this PR does / why we need it**:
This PR introduces changes to the machine deletion workflow, specifically: Node finalizers. Added steps for adding/removing node finalizers, support when dealing with orphaned nodes, and updated related tests. 

* Added a new constant `NodeFinalizer` for tagging node objects and implemented a new step in the machine deletion flow to explicitly remove node finalizers before node deletion.
* Rename node event handler methods (`addNodeToMachine`, `updateNodeToMachine`, `deleteNodeToMachine`) to (`addNode`, `updateNode`, `deleteNode`) and moved them from `machine.go` to `node.go`
  * Added logic to enqueue nodes on the node queue on startup, to add finalizer if not already present.
  * Reconciliation adds finalizers to any nodes not marked as `NotManagedByMCM`, and trigger machine deletion if the node has a deletion timestamp.
  * Removed logic that enables annotation-based machine deletion trigger; machine deletion is now triggered when a node has a deletion timestamp.
  * Triggers machine deletion if node is force-deleted by manual removal of finalizers.
* Updated the machine deletion state machine (`triggerDeletionFlow`) to transition from VM deletion to node finalizer removal, and only then to node deletion. All related status messages and retry logic now reflect this new intermediate step.
* In the node safety controller, orphaned nodes (not managed by MCM) are annotated and now have their MCM finalizer removed to allow garbage collection. Managed nodes are queued for reconciliation for finalizer handling.
* Added and updated unit tests to cover the new node finalizer removal logic, a scenario for orphan node annotation and finalizer removal, and the new machine deletion workflow. 

Changes made in the PR will deprecate [this](https://github.com/gardener/machine-controller-manager/blob/master/docs/faq.md#how-to-delete-machine-object-immedietly-if-i-dont-have-access-to-it) feature.

**Which issue(s) this PR fixes**:
Fixes #1051 

**Special notes for your reviewer**:
<details>
<summary>IT logs</summary>

```
Random Seed: 1764335203

Will run 10 of 10 specs
------------------------------
[BeforeSuite]
/Users/I765230/go/src/github.com/gagan16k/machine-controller-manager-provider-aws/test/integration/controller/controller_test.go:47
  > Enter [BeforeSuite] TOP-LEVEL @ 11/28/25 18:36:53.711
  STEP: Checking for the clusters if provided are available @ 11/28/25 18:36:53.711
[...]

PR #1067, opened 2026-02-05
Update documentation regarding node deletion changes

gagan16k:
This PR updates documentation to account for changes made by #1055 , regarding the `node.machine.sapcloud.io/trigger-deletion-by-mcm` annotation. 

* Updated `docs/faq.md` to clarify that deleting a node now immediately triggers the deletion and replacement of the corresponding machine, and that the `trigger-deletion-by-mcm` annotation is no longer supported on node objects but is still supported on machine and machineDeployment objects for CA-MCM integration.
* Updated comment for `TriggerDeletionByMCM` in `pkg/util/provider/machineutils/utils.go` 

**Release note**:

```doc user
NONE
```


docs/faq.md
```diff
@@ -155,16 +155,14 @@ It can be unpaused again by removing the `Paused` field from the machine-deploym
 
 ### How to delete machine object immedietly if I don't have access to it?
```
aaronfern: I know this line was not added by you, nor is it in scope of this PR, but could you please correct the spelling error in "immediately"?

docs/faq.md
```diff
@@ -155,16 +155,14 @@ It can be unpaused again by removing the `Paused` field from the machine-deploym
...
 
-If the user doesn't have access to the machine objects (like in case of Gardener clusters) and they would like to replace a node immedietly then they can place the annotation `node.machine.sapcloud.io/trigger-deletion-by-mcm: "true"` on their node. This will start the replacement of the machine with a new node.
-
-On the other hand if the user deletes the node object immedietly then replacement will start only after `MachineHealthTimeout`.
-
-This annotation can also be used if the user wants to expedite the [replacement of unhealthy nodes](#how-does-rate-limiting-replacement-of-machine-work-in-mcm-how-is-it-related-to-meltdown-protection)
+If the user deletes the node object, the corresponding machine object is marked for deletion immediately, triggering a replacement. The replacement machine is created to maintain the `desired replicas` specified for the machineDeployment/machineSet.
```
aaronfern: imo, the answer here answers the question "what happens if I delete a node object" rather than what it's supposed to answer. I've given a suggestion but please feel free to improvise or change the direction
```suggestion
If the user doesn't have access to the machine object, they can directly delete the node object. This causes the corresponding machine object to be immediately marked for deletion, triggering a replacement. The replacement machine is created to maintain the `desired replicas` specified for the machineDeployment/machineSet.
```

docs/faq.md
```diff
@@ -155,16 +155,14 @@ It can be unpaused again by removing the `Paused` field from the machine-deploym
...
+This can also be used if the user wants to expedite the [replacement of unhealthy nodes](#how-does-rate-limiting-replacement-of-machine-work-in-mcm-how-is-it-related-to-meltdown-protection)
 
 `NOTE`:
 
-- `node.machine.sapcloud.io/trigger-deletion-by-mcm: "false"` annotation is NOT acted upon by MCM , neither does it mean that MCM will not replace this machine.
-- this annotation would delete the desired machine but another machine would be created to maintain `desired replicas` specified for the machineDeployment/machineSet. Currently if the user doesn't have access to machineDeployment/machineSet then they cannot remove a machine without replacement.
+- The `node.machine.sapcloud.io/trigger-deletion-by-mcm` annotation is no longer supported on node objects. It is however, still supported on machine and machineDeployment objects (used by the CA-MCM cloud provider integration).
+- Setting the annotation to `"false"` has never been acted upon by MCM and does not prevent machine replacement.
```
aaronfern: The old way is fair enough here
```suggestion
- Setting the annotation to `"false"` is NOT acted upon by MCM and does not prevent machine replacement.
```

pkg/util/provider/machineutils/utils.go
```diff
@@ -59,7 +59,7 @@ const (
 	// NotManagedByMCM annotation helps in identifying the nodes which are not handled by MCM
 	NotManagedByMCM = "node.machine.sapcloud.io/not-managed-by-mcm"
 
-	// TriggerDeletionByMCM annotation on the node or machine would trigger the deletion of the corresponding node and machine object in the control cluster
+	// TriggerDeletionByMCM annotation on the machine would trigger the deletion of the corresponding node and machine object in the control cluster
```
takoverflow: Can this line be removed entirely now, as it gives a false impression that adding the annotation on machine would trigger its deletion which is not the case.

Its only used on the `MachineDeployment` now.

docs/faq.md
```diff
@@ -155,16 +155,14 @@ It can be unpaused again by removing the `Paused` field from the machine-deploym
...
+If the user deletes the node object, the corresponding machine object is marked for deletion immediately, triggering a replacement. The replacement machine is created to maintain the `desired replicas` specified for the machineDeployment/machineSet.
+This can also be used if the user wants to expedite the [replacement of unhealthy nodes](#how-does-rate-limiting-replacement-of-machine-work-in-mcm-how-is-it-related-to-meltdown-protection)
 
 `NOTE`:
 
-- `node.machine.sapcloud.io/trigger-deletion-by-mcm: "false"` annotation is NOT acted upon by MCM , neither does it mean that MCM will not replace this machine.
-- this annotation would delete the desired machine but another machine would be created to maintain `desired replicas` specified for the machineDeployment/machineSet. Currently if the user doesn't have access to machineDeployment/machineSet then they cannot remove a machine without replacement.
+- The `node.machine.sapcloud.io/trigger-deletion-by-mcm` annotation is no longer supported on node objects. It is however, still supported on machine and machineDeployment objects (used by the CA-MCM cloud provider integration).
```
takoverflow:
```suggestion
- The `node.machine.sapcloud.io/trigger-deletion-by-mcm` annotation is no longer supported on node objects. It is however, still used on the machineDeployment objects (used by the CA-MCM cloud provider integration).
```

pkg/util/provider/machineutils/utils.go
thiyyakat: Hi. Should this change be reverted ? Someone recently pointed out that the annotation works on the machine too, right? @takoverflow , @gagan16k

takoverflow (dismissed):

PR #1070, opened 2026-02-16
New machine should not move to the pending state if node label is missing

aaronfern:
We recently ran into a bug where a machine was stuck in the `Pending` state while its corresponding node had already joined the cluster. 
The reason for this was that the node name label on the machine had not yet been populated.
This can occur due to the an oversight in the machine creation flow where a newly created machine is moved to the `Pending` state after it's underlying node (or VM) has been created and initialised. The machine creation flow makes a best effort attempt to populate they node label, but it does not requeue the machine if it could not.
The reason we do not want machines to move to the `Pending` state in this case is that when a machine is in this state, it is not allowed to reenter the machine creation flow since we are just waiting for the node to join the k8s cluster. This causes an issue as the node name label is only updated by the machine creation flow and the machine deletion flow, leading to machines being stuck in this state indefinitely

These kind of bugs can surface when a CSP is not able to provide a node name on time. The machine creation flow in turn moved the machine to the `Pending` state after initialising the node (VM) and this machine is stuck

This PR enhances the machine creation flow to not allow newly created machines to proceed to the `Pending` phase when the `NodeName` and `ProviderID` are not available.

Fixes #

**Release note**:

```other developer
newly created machines not moved to `Pending` state without valid node name and providerID from provider
```

### DIFF
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -460,7 +460,7 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 	if (c.targetCoreClient != nil && nodeName == "") || providerID == "" {
 		err := status.Error(codes.Internal, fmt.Sprintf("machine %q: nodeName (%q) or providerID (%q) is empty after creation flow", machine.Name, nodeName, providerID))
 		klog.Error(err)
-		return machineutils.ShortRetry, err
+		return machineutils.MediumRetry, err
 	}
 
 	//Update labels, providerID


pkg/util/provider/machinecontroller/machine.go
```diff
@@ -457,6 +457,12 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 		addresses.Insert(getMachineStatusResponse.Addresses...)
 	}
 
+	if (c.targetCoreClient != nil && nodeName == "") || providerID == "" {
```
r4mek, line 460: Can it be the case where `c.targetCoreClient == nil` and `nodeName == ""`? We won't be returning error in that case.

takoverflow: If `targetCoreClient` is `nil`, i.e. MCM is running for an autonomous cluster, in that case `nodeLabel` is never set. Check `updateLabels()` for ref:
```go
		if c.targetCoreClient != nil {
			// If running without a target cluster, don't add the node label. This disables all interaction with the
			// Node object and related objects in the target cluster.
			metav1.SetMetaDataLabel(&clone.ObjectMeta, v1alpha1.NodeLabelKey, nodeName)
		}
```

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -457,6 +457,12 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
...
+		err := status.Error(codes.Internal, fmt.Sprintf("machine %q: nodeName or providerID is empty after creation flow", machine.Name))
```
takoverflow: Would it make sense to print both values in the error string to know what value was actually empty? WDYT @aaronfern

pkg/util/provider/machinecontroller/machine.go
aaronfern: yes, when `c.targerCoreClient == nil`, the nodeLable is not set, so we do not have a condition where we need to return an error when `c.targetCoreClient == nil` and `nodeName == ""`

PR #1083, opened 2026-03-05
Make ResourceExhausted Retry Period configurable

### LINKED ISSUE #977: Machine creation retry too frequently for machines with `ResourceExhausted`

viragvoros:
This PR makes the retry period used  when machine creation fails with a `codes.ResourceExhausted` error configurable.

PR #981 introduced the `LongRetry` retry period (10 minutes) to handle situations where machine creation fails due to exhausted resources. However, in certain environments this retry duration may still be insufficient. 
To improve flexibility, this PR allows operators to adjust the retry period used specifically for the `ResourceExhausted` error case. The default behavior stays unchanged: if no value is provided as a a command-line flag (`--resource-exhausted-retry`) to the machine-controller-manager-provider, the controller continues to use  `machineutils.LongRetry`.

Fixes #977 
Extends logic introduced in PR #981 to make the retry duration configurable.

The implementation keeps the existing default behavior unchanged:
- The `ResourceExhausted` retry period defaults to `machineutils.LongRetry`.
- The value can be optionally overridden with `--resource-exhausted-retry` CLI flag in machine-controller-manager-provider.
- The override is applied once during startup before the controller begins running.

No changes are required for existing deployments unless operators want to configure the retry duration.

**Release note**:
```other operator
machine-controller-manager now allows configuring the retry duration for ResourceExhausted errors during machine creation with the `--resource-exhausted-retry` CLI flag in machine-controller-manager-provider. If not specified, the default `machineutils.LongRetry` stays unchanged.
``` 
@Kumm-Kai @hasit97

### DIFF
--- a/pkg/util/provider/app/options/options.go
+++ b/pkg/util/provider/app/options/options.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	drain "github.com/gardener/machine-controller-manager/pkg/util/provider/drain"
+	"github.com/gardener/machine-controller-manager/pkg/util/provider/machineutils"
 	machineconfig "github.com/gardener/machine-controller-manager/pkg/util/provider/options"
 	"github.com/spf13/pflag"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -45,9 +46,8 @@ import (
 type MCServer struct {
 	machineconfig.MachineControllerConfiguration
 
-	ControlKubeconfig      string
-	TargetKubeconfig       string
-	ResourceExhaustedRetry string
+	ControlKubeconfig string
+	TargetKubeconfig  string
 }
 
 // NewMCServer creates a new MCServer with a default config.
@@ -119,8 +119,7 @@ func (s *MCServer) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&s.NodeConditions, "node-conditions", s.NodeConditions, "List of comma-separated/case-sensitive node-conditions which when set to True will change machine to a failed state after MachineHealthTimeout duration. It may further be replaced with a new machine if the machine is backed by a machine-set object.")
 	fs.StringVar(&s.BootstrapTokenAuthExtraGroups, "bootstrap-token-auth-extra-groups", s.BootstrapTokenAuthExtraGroups, "Comma-separated list of groups to set bootstrap token's \"auth-extra-groups\" field to")
 
-	fs.StringVar(&s.ResourceExhaustedRetry, "resource-exhausted-retry", "", "Retry duration used when machine creation fails with ResourceExhausted. Defaults to LongRetry.")
-
+	fs.DurationVar(&s.ResourceExhaustedRetry.Duration, "resource-exhausted-retry", time.Duration(machineutils.LongRetry), "Retry duration used when machine creation fails with ResourceExhausted. Defaults to LongRetry.")
 	logs.AddFlags(fs) // adds --v flag for log level.
 
 	leaderelectionconfig.BindFlags(&s.LeaderElection, fs)
@@ -198,5 +197,9 @@ func (s *MCServer) Validate() error {
 	if s.ControlKubeconfig == "" && s.TargetKubeconfig == constants.TargetKubeconfigDisabledValue {
 		errs = append(errs, fmt.Errorf("--control-kubeconfig cannot be empty if --target-kubeconfig=%s is specified", constants.TargetKubeconfigDisabledValue))
 	}
+	if s.ResourceExhaustedRetry.Duration < 0 {
+		errs = append(errs, fmt.Errorf("resource exhausted retry duration should be a non negative value: got: %v", s.ResourceExhaustedRetry.Duration))
+	}
+
 	return utilerrors.NewAggregate(errs)
 }
--- a/pkg/util/provider/app/options/options.go
+++ b/pkg/util/provider/app/options/options.go
@@ -119,7 +119,7 @@ func (s *MCServer) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&s.NodeConditions, "node-conditions", s.NodeConditions, "List of comma-separated/case-sensitive node-conditions which when set to True will change machine to a failed state after MachineHealthTimeout duration. It may further be replaced with a new machine if the machine is backed by a machine-set object.")
 	fs.StringVar(&s.BootstrapTokenAuthExtraGroups, "bootstrap-token-auth-extra-groups", s.BootstrapTokenAuthExtraGroups, "Comma-separated list of groups to set bootstrap token's \"auth-extra-groups\" field to")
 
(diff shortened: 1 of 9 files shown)


viragvoros:
@Kumm-Kai  So an entirely new retry period variable would be better maybe?

pkg/util/provider/app/options/options.go
```diff
@@ -45,8 +45,9 @@ import (
 type MCServer struct {
 	machineconfig.MachineControllerConfiguration
 
-	ControlKubeconfig string
-	TargetKubeconfig  string
+	ControlKubeconfig      string
+	TargetKubeconfig       string
+	ResourceExhaustedRetry string
```
thiyyakat: This field can be added to `MachineControllerManagerConfiguration` in `pkg/options/types.go` instead.

pkg/util/provider/app/options/options.go
```diff
@@ -116,6 +117,8 @@ func (s *MCServer) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&s.NodeConditions, "node-conditions", s.NodeConditions, "List of comma-separated/case-sensitive node-conditions which when set to True will change machine to a failed state after MachineHealthTimeout duration. It may further be replaced with a new machine if the machine is backed by a machine-set object.")
 	fs.StringVar(&s.BootstrapTokenAuthExtraGroups, "bootstrap-token-auth-extra-groups", s.BootstrapTokenAuthExtraGroups, "Comma-separated list of groups to set bootstrap token's \"auth-extra-groups\" field to")
 
+	fs.StringVar(&s.ResourceExhaustedRetry, "resource-exhausted-retry", "", "Retry duration used when machine creation fails with ResourceExhausted. Defaults to LongRetry.")
```
thiyyakat: This can be of type `fs.DurationVar` like so:

`fs.DurationVar(&s.ResourceExhaustedRetryDuration.Duration, "resource-exhausted-retry", time.Duration(machineutils.LongRetry), "Retry duration used when machine creation fails with ResourceExhausted. Defaults to LongRetry.")`

This way the default is set to `machineutils.LongRetry` here itself.

Please also add validation for this option below in the `Validate()` method. 

`if s.ResourceExhaustedRetryDuration.Duration < 0 {
		errs = append(errs, fmt.Errorf("resource exhausted retry duration should be a non negative value: got: %v", s.ResourceExhaustedRetryDuration.Duration))
	}`

thiyyakat (commented):
Hi! Thanks for the PR. Just a few changes to maintain consistency with existing code. PTAL.

pkg/util/provider/app/options/options.go
viragvoros: Hi @thiyyakat, thank you for the review :) I added the changes. In this comment, did you mean the `pkg/util/provider/options/types.go` `MachineControllerConfiguration`? The variable was intended to be introduced only in the `MachineController` not in the outer main.

thiyyakat: Hi! Yes, you're absolutely right. Apologies.

pkg/util/provider/options/types.go
```diff
@@ -73,6 +73,8 @@ type MachineControllerConfiguration struct {
 
 	//BootstrapTokenAuthExtraGroups is a comma-separated string of groups to set bootstrap token's "auth-extra-groups" field to.
 	BootstrapTokenAuthExtraGroups string
+
```
thiyyakat, line 76: Please add a docstring for this field:

`// ResourceExhaustedRetry is the duration after which the controller retries reconciliation of machine object if the call to driver.CreateMachine failed with "Resource Exhausted" error.`

thiyyakat (changes requested):
Hi @viragvoros ! There is a docstring missing. Please add it. TIA.

pkg/util/provider/options/types.go
viragvoros: added it, thank you for the suggestion :)

pkg/util/provider/app/options/options.go
```diff
@@ -116,6 +117,7 @@ func (s *MCServer) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&s.NodeConditions, "node-conditions", s.NodeConditions, "List of comma-separated/case-sensitive node-conditions which when set to True will change machine to a failed state after MachineHealthTimeout duration. It may further be replaced with a new machine if the machine is backed by a machine-set object.")
 	fs.StringVar(&s.BootstrapTokenAuthExtraGroups, "bootstrap-token-auth-extra-groups", s.BootstrapTokenAuthExtraGroups, "Comma-separated list of groups to set bootstrap token's \"auth-extra-groups\" field to")
 
+	fs.DurationVar(&s.ResourceExhaustedRetry.Duration, "resource-exhausted-retry", time.Duration(machineutils.LongRetry), "Retry duration used when machine creation fails with ResourceExhausted. Defaults to LongRetry.")
```
gagan16k: This statement could be reworded to make it clearer that a value of `0` gets set to `LongRetry`

pkg/util/provider/app/options/options.go
```diff
@@ -190,6 +192,9 @@ func (s *MCServer) Validate() error {
 	if s.ControlKubeconfig == "" && s.TargetKubeconfig == constants.TargetKubeconfigDisabledValue {
 		errs = append(errs, fmt.Errorf("--control-kubeconfig cannot be empty if --target-kubeconfig=%s is specified", constants.TargetKubeconfigDisabledValue))
 	}
+	if s.ResourceExhaustedRetry.Duration < 0 {
+		errs = append(errs, fmt.Errorf("resource exhausted retry duration should be a non negative value: got: %v", s.ResourceExhaustedRetry.Duration))
```
gagan16k:
```suggestion
		errs = append(errs, fmt.Errorf("resource exhausted retry duration should be a non-negative value, got: %v", s.ResourceExhaustedRetry.Duration))
```
nit

gagan16k (commented):
Thanks for the PR! I’ve added a few small suggestions, PTAL

pkg/util/provider/app/options/options.go
viragvoros: Hi @gagan16k, thank you for the review. I added the changes for the 0 value input and the error as well.

thiyyakat:
@viragvoros, please rebase your changes.

gagan16k:
Thanks for the PR
/lgtm

thiyyakat:
Thanks!
/lgtm

thiyyakat (approved):

PR #1086, opened 2026-03-12
Exclude terminating machines when managing machineset replicas

takoverflow:
Excludes considering terminating machines in machineset controller when 
computing the necessary machines that need to be created/removed.

A machine will only move to terminating state once the machine set has been scaled-down
hence if a machine is already in terminating state, it should be excluded from the list
of machines that are being considered when computing the active replicas of a machine set.

The only other scenario where a machine moves to terminating is when its in `Failed` state 
and the MCS controller proceeds to consider it as a stale machine and marks it
for removal. In this case, by excluding the terminating machine from diff computation,
we ensure that a replacement machine is spawned for the failed machine.

**Release note**:

```bugfix operator
Exclude terminating machines when managing machineset replicas to allow for faster replacement of a failed machine.
```

### DIFF
--- a/pkg/controller/machineset.go
+++ b/pkg/controller/machineset.go
@@ -359,6 +359,9 @@ func (c *controller) manageReplicas(ctx context.Context, allMachines []*v1alpha1
 	// are machines that require removal.
 	machinesWithoutUpdateSuccessfulLabelDiff := len(machinesWithoutUpdateSuccessfulLabel) - int(machineSet.Spec.Replicas)
 
+	// During in-place updates, ScaleUps are disabled in the oldMachineSet and the newMachineSet's
+	// ReplicaCount would never increase before a machine from oldMachineSet is moved.
+	// So no need for any special case during in-place updates when checking if machines need to be created.
 	if nonTerminatingMachinesDiff < 0 {
 		// If MachineSet is frozen and no deletion timestamp, don't process it
 		if machineSet.Labels["freeze"] == "True" && machineSet.DeletionTimestamp == nil {


pkg/controller/machineset.go
```diff
@@ -335,19 +335,28 @@ func (c *controller) manageReplicas(ctx context.Context, allMachines []*v1alpha1
...
+		}
+	}
+
+	for _, m := range nonTerminatingMachines {
 		if m.Labels[v1alpha1.LabelKeyNodeUpdateResult] != v1alpha1.LabelValueNodeUpdateSuccessful {
 			machinesWithoutUpdateSuccessfulLabel = append(machinesWithoutUpdateSuccessfulLabel, m)
 		}
 	}
-	allMachinesDiff := len(allMachines) - int(machineSet.Spec.Replicas)
+	nonTerminatingMachinesDiff := len(nonTerminatingMachines) - int(machineSet.Spec.Replicas)
```
thiyyakat: Since `nonTerminatingMachines` is only being used to compute the number of non-terminating machines, can you skip the extra for loop and do this instead?

```
	var (
		numNonTerminatingMachines            int
		machinesWithoutUpdateSuccessfulLabel []*v1alpha1.Machine
	)
	for _, m := range allMachines {
		if m.Status.CurrentStatus.Phase != v1alpha1.MachineTerminating {
			numNonTerminatingMachines++
			if m.Labels[v1alpha1.LabelKeyNodeUpdateResult] != v1alpha1.LabelValueNodeUpdateSuccessful {
				machinesWithoutUpdateSuccessfulLabel = append(machinesWithoutUpdateSuccessfulLabel, m)
			}
		}
	}
	nonTerminatingMachinesDiff := numNonTerminatingMachines - int(machineSet.Spec.Replicas)
```

Issue #1008, opened 2025-07-01
Support Preservation of Failed Machines for diagnostics


elankath:
**How to categorize this issue?**

/area control-plane
/kind enhancement
/priority 3

**What would you like to be added**:

#### Background
 Currently, the `machine-controller-manager` moves `Running` machines to `Unknown` phase in case of errors and then to `Failed` phase after the configured `machine-health-timeout`. `Failed` machines are swiftly moved to the `Terminating` phase, the node is drained and the machine object deleted.

#### Need
There is a need for preserving VM's corresponding to Machines so that the operator/support/SRE can analyze and diagnose root cause of failure.  However, there should be a limit to the number of machines that are preserved for the worker pool. There should also be a configurable timeout beyond which the MCM goes ahead with Machine termination. 

We propose enhancing [MachineConfiguration](https://github.com/gardener/machine-controller-manager/blob/95884c4be1686543981ad04349ae775f255345de/pkg/apis/machine/v1alpha1/shared_types.go#L30) with `FailedMachineTimeout *metav1.Duration` and the [MachineDeploymentSpec](https://github.com/gardener/machine-controller-manager/blob/95884c4be1686543981ad04349ae775f255345de/pkg/apis/machine/types.go#L446) with the `FailedMachinePreserveMax *int32`. (Exact field names/locations are subject to change after design)

In addition, we will enhance the gardener `machineControllerManager` settings in the shoot spec to support operator configuration of the above fields  in the worker pool in a separate gardener PR.

```yaml
machineControllerManager:
   failedMachinePreserveMax: 2
   failedMachinePreserveTimeout: 3h
```

- The MCM will annotate all preserved failed machines with `node.machine.sapcloud.io/preserve-when-failed=true`
- The user/operator can also explicitly mark a `Machine` or its associated `Node` with the annotation `node.machine.sapcloud.io/preserve-when-failed=true`.  
- If the current count of preserved `Failed` machines is at or exceeds `failedMachinePreserveMax` then the annotation will not be accepted. (The annotation will be deleted)
- If the current count  of preserved `Failed` machines is at or exceeds `failedMachinePreserveMax`, then any `Unknown` Machines that move to the `Failed` phase will not be preserved and will be terminated. 
- The `failedMachinePreserveMax` MUST be set in the shoot spec, otherwise annotation `node.machine.sapcloud.io/preserve-when-failed=true` added by operator/support to a `Machine` has no effect. 
[...]

PR #1092, opened 2026-04-06
Track preserved failed machines in MCS and MCD status

thiyyakat:
This PR introduces 2 changes:
1. It tracks the number of preserved failed machines in the MCSs and MCDs in a new field `preservedFailedReplicas` added to the MCS and MCD Status.
2. It prevents the MCD from being marked as unhealthy by excluding preserved failed machines from `failedMachine` in the MCS and MCD Status. This will also prevent shoot reconciliation from getting stuck. 

Additionally, the PR modifies the sorting logic for `ActiveMachines` -- if two machines are preserved, one auto-preserved and one explicitly preserved through annotation by user/operator, the sorting logic de-prioritizes explicitly preserved machines for deletion.

The usage doc for preservation has been updated with a warning regarding the behaviour of DWD when number of preserved failed machines exceeds the threshold set for DWD.

Extends #1008

MCM unit and integration tests passed with changes. 
The changes were manually tested out using virtual provider. Below log shows MCD status when it has 1 preserved failed replica.
```
status:
  conditions:
  - lastTransitionTime: "2026-04-06T07:20:38Z"
    lastUpdateTime: "2026-04-06T07:20:58Z"
    message: MachineSet "shoot--xxx--preserve-pr-preserve-pr-cpu-z1-5cc55" has
      successfully progressed.
    reason: NewMachineSetAvailable
    status: "True"
    type: Progressing
  - lastTransitionTime: "2026-04-06T11:27:16Z"
    lastUpdateTime: "2026-04-06T11:27:16Z"
    message: Deployment has minimum availability.
    reason: MinimumReplicasAvailable
    status: "True"
    type: Available
  observedGeneration: 1
  preservedFailedReplicas: 1
  replicas: 1
  unavailableReplicas: 1
  updatedReplicas: 1
  ```
**Release note**:

```other operator
Track preserved failed machines in field `preservedFailedReplicas` in the Status of MCS and MCD, and prevent these machines from resulting in MCD being marked as unhealthy.
```


pkg/controller/deployment_machineset_util.go
```diff
@@ -68,8 +70,9 @@ func updateMachineSetStatus(ctx context.Context, machineClient machineapi.Machin
 			fmt.Sprintf("fullyLabeledReplicas %d->%d, ", is.Status.FullyLabeledReplicas, newStatus.FullyLabeledReplicas)+
 			fmt.Sprintf("readyReplicas %d->%d, ", is.Status.ReadyReplicas, newStatus.ReadyReplicas)+
 			fmt.Sprintf("availableReplicas %d->%d, ", is.Status.AvailableReplicas, newStatus.AvailableReplicas)+
-			fmt.Sprintf("sequence No: %v->%v,", is.Status.ObservedGeneration, newStatus.ObservedGeneration)+
-			fmt.Sprintf("autoPreserveFailedMachineCount %v->%v", is.Status.AutoPreserveFailedMachineCount, newStatus.AutoPreserveFailedMachineCount))
+			fmt.Sprintf("sequence No: %v->%v, ", is.Status.ObservedGeneration, newStatus.ObservedGeneration)+
```
takoverflow: Although its not part of your changes
I don't know why this log prints `sequence No:`, can it be changed to `observerGeneration:` instead?

pkg/controller/deployment_machineset_util.go
```diff
@@ -118,7 +122,9 @@ func calculateMachineSetStatus(is *v1alpha1.MachineSet, filteredMachines []*v1al
 				readyReplicasCount++
 			}
 		}
-		if machine.Status.LastOperation.State == v1alpha1.MachineStateFailed {
+		isMachinePreserved := !machine.Status.CurrentStatus.PreserveExpiryTime.IsZero()
+		// exclude failed preserved machines from failedMachines to prevent MCD from being marked as unavailable
+		if machine.Status.LastOperation.State == v1alpha1.MachineStateFailed && !isMachinePreserved {
```
takoverflow: Rather than inverting the variable definition and then inverting it again, consider calling it just `isMachineNotPreserved`?

```suggestion
		isMachineNotPreserved := machine.Status.CurrentStatus.PreserveExpiryTime.IsZero()
		// exclude failed preserved machines from failedMachines to prevent MCD from being marked as unavailable
		if machine.Status.LastOperation.State == v1alpha1.MachineStateFailed && isMachineNotPreserved {
```

pkg/controller/deployment_machineset_util.go
```diff
@@ -127,8 +133,11 @@ func calculateMachineSetStatus(is *v1alpha1.MachineSet, filteredMachines []*v1al
 				machineSummary.OwnerRef = controller.Name
 			}
 			failedMachines = append(failedMachines, machineSummary)
+		} else if isMachinePreserved && machine.Status.CurrentStatus.Phase == v1alpha1.MachineFailed {
```
takoverflow: Why is this checking `machine.Status.CurrentStatus.Phase` but the above is checking `machine.Status.LastOperation.State`? Can the latter not be checked here as well? (Or the former for that matter)

That is:
```go
if machine.Status.LastOperation.State == v1alpha1.MachineFailed {
    if isMachinePreserved {
        preservedFailedReplicasCount++
    } else {
        // Update MachineSummary
    }
}

File: pkg/controller/deployment_machineset_util.go [hunk 4]
```diff
@@ -224,6 +234,10 @@ func isMachineAvailable(machine *v1alpha1.Machine) bool {
 	return false
 }
 
+func isMachinePreservedAndFailed(machine *v1alpha1.Machine) bool {
```
takoverflow, line 237: This function is defined but is not being used above.

pkg/controller/deployment_machineset_util.go
thiyyakat: I don't quite know the semantic difference between `LastOperation.State` being `MachineFailed` and `CurrentStatus.Phase` being `MachineFailed`. Is Phase also Failed whenever State is Failed, and vice versa?

pkg/controller/deployment_machineset_util.go
thiyyakat: I'm guessing you mean "observedGeneration", right?

takoverflow: Yes, it was a typo

pkg/controller/deployment_machineset_util.go
takoverflow: I would assume this distinction was required since you introduced it here, that's why I wanted to understand what was the rationale behind using either of the fields.

Regarding your second question, that would require checking the source to be sure if both are set to Failed atomically.

thiyyakat: I don't think they are necessarily set to Failed at the same time. If the last operation fails, the State may be set to Failed, but the Phase may still be Unknown or CLBO. For the preservedFailedReplica counting, it is essential that the _Phase_ be Failed. We cannot club the two. 

To me it seems the purpose of failedMachines is to hold a summary of machines for which the last operation has failed. The machines may not necessarily be in Failed Phase. Please correct me if I am wrong.


Example: https://github.com/gardener/machine-controller-manager/blob/62c3070366d8d10bc1379e67ba2248c7d32d933d/pkg/util/provider/machinecontroller/machine_util.go#L819-L834

thiyyakat: Now that I think of it, maybe all preserved machines should not be excluded from failedMachines summary, and only Failed Preserved machines should be. Basically:
```
if isMachinePreserved && machine.Status.CurrentStatus.Phase == v1alpha1.MachineFailed {
    preservedFailedReplicasCount++
} else if machine.Status.LastOperation.State == v1alpha1.MachineStateFailed {
    machineSummary.Name = machine.Name
    machineSummary.ProviderID = machine.Spec.ProviderID
    machineSummary.LastOperation = machine.Status.LastOperation
    //ownerRef populated here, so that deployment controller doesn't have to add it separately
    if controller := metav1.GetControllerOf(machine); controller != nil {
	    machineSummary.OwnerRef = controller.Name
     }
    failedMachines = append(failedMachines, machineSummary)
} 
```
WDYT?

pkg/controller/deployment_util.go
```diff
@@ -1041,6 +1042,17 @@ func GetAvailableReplicaCountForMachineSets(MachineSets []*v1alpha1.MachineSet)
 	return totalAvailableReplicas
 }
 
+// GetPreservedFailedReplicaCountForMachineSets returns the number of available machines corresponding to the given machine sets.
```
aaronfern: Docstring needs to be corrected

pkg/controller/deployment_machineset_util.go
takoverflow: Hmm right, yeah that makes sense to me. Maybe the others can share their opinion as well.

thiyyakat: @aaronfern , @elankath, @r4mek, @gagan16k. Please let me know your thoughts as well. Thanks!

pkg/controller/deployment_machineset_util.go
thiyyakat: Have not removed the function, because the code now calls it. I prefer using the function to improve readability, though it is used only once.

pkg/controller/deployment_sync.go
```diff
@@ -654,7 +656,7 @@ func calculateDeploymentStatus(allISs []*v1alpha1.MachineSet, newIS *v1alpha1.Ma
 	// Copy conditions one by one so we won't mutate the original object.
 	status.Conditions = append(status.Conditions, deployment.Status.Conditions...)
 
-	if availableReplicas >= (deployment.Spec.Replicas)-MaxUnavailable(*deployment) {
+	if availableReplicas >= (deployment.Spec.Replicas)-MaxUnavailable(*deployment)-totalPreservedFailedReplicas {
```
gagan16k: Since `totalPreservedFailedReplicas` is calculated across all machineSets, would it be possible for the whole right-hand side of this condition to go below zero?
And then the condition will always be true

pkg/apis/machine/v1alpha1/machinedeployment_types.go
```diff
@@ -213,9 +213,13 @@ type MachineDeploymentStatus struct {
 	// +optional
 	AvailableReplicas int32 `json:"availableReplicas,omitempty"`
 
+	// PreservedFailedReplicas is the number of preserved machines in Failed phase targeted by this MachineDeployment
+	PreservedFailedReplicas int32 `json:"preservedFailedReplicas,omitempty"`
```
gagan16k, line 218:
```suggestion
	// PreservedFailedReplicas is the number of preserved machines in Failed phase targeted by this MachineDeployment
	// +optional
	PreservedFailedReplicas int32 `json:"preservedFailedReplicas,omitempty"`
```
Should this have an optional marker? All the other fields have it

pkg/controller/deployment_machineset_util.go
takoverflow: Discussed with @aaronfern and @thiyyakat offline, they were in favor of excluding only failed preserved machines from `FailedMachines`.

pkg/controller/controller_utils.go
```diff
@@ -745,6 +745,24 @@ func (s ActiveMachines) Less(i, j int) bool {
...
+			return m.Annotations[machineutils.PreserveMachineAnnotationKey] == machineutils.PreserveMachineAnnotationValuePreservedByMCM
+		}
+		isAutoPreservedI := isAutoPreserved(s[i])
+		isAutoPreservedJ := isAutoPreserved(s[j])
+		if isAutoPreservedI != isAutoPreservedJ {
+			return isAutoPreservedI // if s[i] is auto-preserved, it is "less", and should be deleted first
+		}
+		// if both are auto-preserved or both are explicitly preserved, we decide based on whichever machine
+		// has an earlier preserve expiry time
+		diff := s[j].Status.CurrentStatus.PreserveExpiryTime.Time.Sub(s[i].Status.CurrentStatus.PreserveExpiryTime.Time)
```
r4mek: Consider using `Before()` function

thiyyakat: I had originally written this with `Before()`. But I am performing 3 checks- Before, After and Equality. Calculating the diff instead makes the code less verbose.

takoverflow (approved):

Issue #1064, opened 2026-01-16
Distinguish between a failed machine creation and a failed node creation


Kostov6:
**How to categorize this issue?**

/area ops-productivity
/kind bug

**What happened**:
```
❯ k -n ... get machine <machine> -o yaml | yq .status
currentStatus:
  lastUpdateTime: "2026-01-16T10:04:42Z"
  phase: Pending
  timeoutActive: true
lastOperation:
  description: Creating machine on cloud provider
  lastUpdateTime: "2026-01-16T10:04:42Z"
  state: Processing
  type: Create

❯ aws ec2 describe-instances ... --query 'Reservations[].Instances[].[Tags[?Key==`Name`] | [0].Value, State.Name]'
[
    [
        "<machine>",
        "running"
    ]
]

❯ k get nodes
(empty)
```

**What you expected to happen**:

The above example shows a successfully created machine on the cloud provider but a failure in kubelet registering the node. In this case a more appropriate description would be `Waiting for kubelet to create node object`. 

This message will improve debugging failing shoot clusters as it is much more accurate

**Environment**:

- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- Others:

PR #1065, opened 2026-01-21
Skip finalizers for non-MCM nodes

gagan16k:
**What this PR does / why we need it**:
This PR ensures that finalizers are not added to non-MCM nodes lacking the `NotManagedByMCM` annotation, as this can lead to testing issues in kubernetes environments.
This is done by verifying that an associated machine exists for the node before adding finalizers.

Changes:
- Skip processing node updates and cluster node reconciliation if there is no associated machine
- Replaced the check for a removed finalizer(`hasNodeFinalizerBeenRemoved`) with a direct check for the presence of the finalizer

**Special notes for your reviewer**:
<details>
<summary>IT logs</summary>

```
Random Seed: 1769002937

Will run 10 of 10 specs
------------------------------
[BeforeSuite]
/Users/I765230/go/src/github.com/gagan16k/machine-controller-manager-provider-aws/test/integration/controller/controller_test.go:47
  > Enter [BeforeSuite] TOP-LEVEL @ 01/21/26 19:12:27.07
  STEP: Checking for the clusters if provided are available @ 01/21/26 19:12:27.071
  2026/01/21 19:12:27 Control cluster kube-config - /Users/I765230/go/src/github.com/gagan16k/machine-controller-manager-provider-aws/dev/kube-configs/kubeconfig_control.yaml
  2026/01/21 19:12:27 Target cluster kube-config  - /Users/I765230/go/src/github.com/gagan16k/machine-controller-manager-provider-aws/dev/kube-configs/kubeconfig_target.yaml
  STEP: Killing any existing processes @ 01/21/26 19:12:29.937
  STEP: Checking Machine-Controller-Manager repo is available at: ../../../dev/mcm @ 01/21/26 19:12:30.212
  STEP: Scaledown existing machine controllers @ 01/21/26 19:12:30.212
  STEP: Starting Machine Controller  @ 01/21/26 19:12:30.411
  STEP: Starting Machine Controller Manager @ 01/21/26 19:12:30.424
  STEP: Cleaning any old resources @ 01/21/26 19:12:30.429
  2026/01/21 19:12:30 machinedeployments.machine.sapcloud.io "test-machine-deployment" not found
  2026/01/21 19:12:30 machines.machine.sapcloud.io "test-machine" not found
  2026/01/21 19:12:31 machineclasses.machine.sapcloud.io "test-mc-v1" not found
  2026/01/21 19:12:31 machineclasses.machine.sapcloud.io "test-mc-v2" not found
[...]

PR #1102, opened 2026-05-18
Trigger node deletion when machine is deleted

### LINKED ISSUE #1044: MCM should always delete Node if corresponding Machine object is not present during termination flow

/area robustness
/kind bug
/priority 2

**What happened**:

There are cases where if a `Machine` is `force-deleted`, the `Node` cleanup does not occur. This happens if the Machine object is explicitly deleted by support.  And the node lifecycle controller fails to clear the node.

Difficult to reproduce. However, the `Node` object should be deleted ASAP after the corresponding `Machine` object is removed instead of needlessly waiting for node lifecycle controller.

We should not skip node deletion in MCM if corresponding `Machine` object is missing. This was long known gap since we were depending upon clean-up by standard k8s controller.

This problem occurred in Azure.

gagan16k:
1. Machine controller now explicitly deletes the finalizer and backing node from the target cluster on machine delete events
2. The machine status description during creation has been updated from `"Creating machine on cloud provider"` to `"VM created on cloud provider. Waiting for node registration"`, so the status accurately reflects which stage the machine is in.

Fixes #1044, #1064

**Tests**:
Passed unit tests

Manual tests performed by deleting node manually
```
(⎈|garden--aws-ha-external:garden shoot--i765230--demo)➜  machine-controller-manager git:(pr/gagan16k/1102) k get mc
NAME                                              STATUS    AGE     NODE
shoot--i765230--demo-worker-cpu-z1-7988d-hnw4w    Running   7h23m   ip-10-180-3-122.eu-west-1.compute.internal
shoot--i765230--demo-worker-etcd-z1-5dc7f-ttqd5   Running   7h23m   ip-10-180-137-245.eu-west-1.compute.internal

(⎈|garden-i765230--demo-external:default garden-i765230)➜  machine-controller-manager git:(pr/gagan16k/1102) k get no
NAME                                           STATUS   ROLES    AGE     VERSION
ip-10-180-137-245.eu-west-1.compute.internal   Ready    worker   7h22m   v1.34.9
ip-10-180-3-122.eu-west-1.compute.internal     Ready    worker   7h22m   v1.34.9



# Force-delete machine by removing finalizer
(⎈|garden--aws-ha-external:garden shoot--i765230--demo)➜  machine-controller-manager git:(pr/gagan16k/1102) k -n shoot--i765230--demo patch machine shoot--i765230--demo-worker-cpu-z1-7988d-hnw4w --type=merge -p '{"metadata":{"finalizers":null}}'
machine.machine.sapcloud.io/shoot--i765230--demo-worker-cpu-z1-7988d-hnw4w patched
(⎈|garden--aws-ha-external:garden shoot--i765230--demo)➜  machine-controller-manager git:(pr/gagan16k/1102) k -n shoot--i765230--demo delete machine shoot--i765230--demo-worker-cpu-z1-7988d-hnw4w
machine.machine.sapcloud.io "shoot--i765230--demo-worker-cpu-z1-7988d-hnw4w" deleted



# Machine and node deleted
(⎈|garden-i765230--demo-external:default garden-i765230)➜  machine-controller-manager git:(pr/gagan16k/1102) k get no
NAME                                           STATUS   ROLES    AGE     VERSION
ip-10-180-137-245.eu-west-1.compute.internal   Ready    worker   7h24m   v1.34.9

(⎈|garden--aws-ha-external:garden shoot--i765230--demo)➜  machine-controller-manager git:(pr/gagan16k/1102) k get mc
NAME                                              STATUS    AGE     NODE
shoot--i765230--demo-worker-cpu-z1-7988d-h299t              3s
shoot--i765230--demo-worker-etcd-z1-5dc7f-ttqd5   Running   7h25m   ip-10-180-137-245.eu-west-1.compute.internal
```

For #1064, machine during creation has message
```
Last Operation:
    Description:       VM created on cloud provider. Waiting for node registration
    Last Update Time:  2026-08-12T06:20:01Z
    State:             Processing
    Type:              Create
```

**Release note**:
  
  ```bugfix operator
MCM now deletes the backing node when its machine is force-deleted, and clears finalizers on nodes terminating without a backing machine. 
Machine creation status is also modified to distinguish cloud provider failures from node registration failures.
```

### CODE REVIEW

File: pkg/util/provider/machinecontroller/machine.go [hunk 1]
```diff
@@ -99,6 +99,14 @@ func (c *controller) deleteMachine(obj any) {
 		}
 	}
 	c.enqueueMachineTermination(machine, "handling terminating machine object DELETE event")
+
+	if c.targetCoreClient != nil {
+		if nodeName := machine.Labels[v1alpha1.NodeLabelKey]; nodeName != "" {
+			if err := c.targetCoreClient.CoreV1().Nodes().Delete(context.Background(), nodeName, metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
+				klog.Errorf("Machine %q: failed to delete node %q: %v", machine.Name, nodeName, err)
```
thiyyakat: Could you please rephrase this? Maybe: `"failed to delete backing node %q of machine %q queued for deletion"`? Please start error message with lowercase when possible.

pkg/util/provider/machinecontroller/node.go
```diff
@@ -153,6 +157,10 @@ func (c *controller) reconcileClusterNodeKey(key string) error {
 	machine, err := c.getMachineFromNode(node.Name)
 	if err != nil {
 		if errors.Is(err, errNoMachineMatch) {
+			if node.DeletionTimestamp != nil {
```
thiyyakat: Could you please update the docstring above since we are no longer ignoring such updates.

thiyyakat (commented):
Thanks for the changes. Just two comments. PTAL.

pkg/util/provider/machinecontroller/machine_test.go
```diff
@@ -677,7 +677,7 @@ var _ = Describe("machine", func() {
 								LastUpdateTime: metav1.Now(),
 							},
 							LastOperation: v1alpha1.LastOperation{
-								Description:    "Creating machine on cloud provider",
+								Description:    "Creating machine on cloud provider. Waiting for node object to register",
```
r4mek: maybe consider changing it to `Created machine on cloud provider. Waiting for node object to register`

gagan16k: Changed message

pkg/util/provider/machinecontroller/node.go
gagan16k: Changed docstring here and in `updateNode` handler

pkg/util/provider/machinecontroller/node.go
```diff
@@ -148,16 +153,23 @@ func (c *controller) reconcileClusterNodeKey(key string) error {
...
+				klog.V(2).Infof("ClusterNode %q: Node is being deleted with no backing machine, removing finalizer", key)
+				return c.removeNodeFinalizers(ctx, node)
+			}
 			klog.Errorf("ClusterNode %q: No machine found matching node, skipping adding finalizers", key)
 			return nil
+		} else {
+			klog.Errorf("ClusterNode %q: error fetching machine for node: %v", key, err)
+			return err
```
thiyyakat: This doesn't require an `else` block, since you are always returning on entering the `if` block.

thiyyakat (commented):
Just one small nit.

pkg/util/provider/machinecontroller/machine.go
thiyyakat: Apologies. My suggestion is incorrect since the machine object is already deleted. Maybe something like `"failed to delete backing node %q of deleted machine %q : %v"` is more appropriate.

thiyyakat:
Thanks for addressing the comments.
/lgtm

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -100,6 +100,15 @@ func (c *controller) deleteMachine(obj any) {
 		}
 	}
 	c.enqueueMachineTermination(machine, "handling terminating machine object DELETE event")
+
+	// Node retains finalizer here, removed by reconcileClusterNodeKey once DeletionTimestamp is set
```
takoverflow: Why is the finalizer being retained here, wouldn't it make sense to speed up cleanup by removing it as well? Why wait for node reconciliation to happen?

Is it for convenience or there's some value in preserving the finalizer still?

pkg/util/provider/machinecontroller/node.go
```diff
@@ -148,11 +153,17 @@ func (c *controller) reconcileClusterNodeKey(key string) error {
 		return nil
 	}
 
-	// Ignore node updates without an associated machine. Retry only for errors other than errNoMachineMatch;
-	// transient fetch errors will be eventually requeued by the update handler due to kubelet updates.
+	// Ignore node updates without an associated machine, except when the node is being
+	// deleted, then remove the MCM finalizer to unblock deletion. Retry only for
+	// errors other than errNoMachineMatch; transient fetch errors will be eventually
```
takoverflow: If there's a transient error while fetching the machine for the node, and node deletion was triggered via `kubectl delete node`, we can prematurely remove the finalizers now right?

Or there's no possibility of getting no machine match (when there is a machine) because we're relying on listers?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -530,7 +539,7 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 	if machine.Status.CurrentStatus.Phase == "" || machine.Status.CurrentStatus.Phase == v1alpha1.MachineCrashLoopBackOff {
 		clone := clone.DeepCopy()
 		clone.Status.LastOperation = v1alpha1.LastOperation{
-			Description:    "Creating machine on cloud provider",
+			Description:    "VM created on cloud provider. Waiting for node registration",
```
takoverflow, line 570: When we have these standard `LastOperation.Description` always, does it make sense to declare all such strings as semantic constants and refer to those instead?

This would ensure that in future if any description is changed, it would only require changing the constant and not replacing all occurences manually, WDYT?

takoverflow (commented):
Thanks for the PR, have some queries/suggestion PTAL

pkg/util/provider/machinecontroller/machine.go
gagan16k: There was a pre-existing mechanism to handle the finalizer removal (So that any nodes that do not have a machine do not get a finalizer, added as part of #1065).
So I did not add it here, as the finalizer is removed anyways

pkg/util/provider/machinecontroller/machine.go
gagan16k: I agree, but should I do it as part of this PR?

takoverflow: Can be tackled as a separate cleanup PR.

pkg/util/provider/machinecontroller/node.go
gagan16k: `NoMachineMatch` (when there is a machine) is theoretically possible but is probably very rare. 
Could add an extra API server fetch to actually check if the machine is truly gone, but this might be unnecessary? Not sure, WDYT

pkg/util/provider/machinecontroller/machine.go
takoverflow: Yes but AFAICS the current `reconcileClusterNodeKey` doesn't have any finalizer removal logic https://github.com/gardener/machine-controller-manager/blob/580d98fae362683f29c976ca546cd4535d05f9e0/pkg/util/provider/machinecontroller/node.go#L133

Please correct me if I'm wrong but the only place we currently actively remove the finalizer from the node and issue deletion is `deleteNodeObject` which is part of `triggerDeletionFlow`.

pkg/util/provider/machinecontroller/node.go
takoverflow: That's why I thought maybe the finalizer removal and node deletion could happen as a unit as part of machine on delete event similar to how its done in `deleteNodeObject`. So all this checking here becomes unnecessary.

Not proposing that this is how it should be done, just pitching an alternative to see if there's any pros/cons of doing it either way.

pkg/util/provider/machinecontroller/machine.go
gagan16k: Oh, really sorry for the confusion, #1065 had no finalizer removal steps (The change was to never add it in the first place).

But we do have it now now as part of this [PR](https://github.com/gagan16k/machine-controller-manager/blob/2cf292ef3d95e504616bebf1d70b9cfa4512a704/pkg/util/provider/machinecontroller/node.go#L165-L166)

gagan16k:
Pushed some major changes, needs re-review @takoverflow @thiyyakat @r4mek

pkg/util/provider/machinecontroller/node.go
gagan16k: Made these changes as part of [f8410ea](https://github.com/gardener/machine-controller-manager/pull/1102/commits/f8410ea28130eb3daf92e06131c35d3ef045ed24)

pkg/util/provider/machinecontroller/machine.go
gagan16k: Changed this behaviour as part of [f8410ea](https://github.com/gardener/machine-controller-manager/pull/1102/commits/f8410ea28130eb3daf92e06131c35d3ef045ed24)

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -100,6 +101,25 @@ func (c *controller) deleteMachine(obj any) {
...
+	// Remove the MCM finalizer from and delete the backing node
+	// The orphan-VM safety net eventually removes the finalizer from nodes if this still fails.
+	nodeName := machine.Labels[v1alpha1.NodeLabelKey]
+	go func() {
+		retryErr := retry.OnError(retry.DefaultBackoff, func(error) bool { return true }, func() error {
+			_, _, err := c.removeFinalizerAndDeleteNode(context.Background(), machine)
+			return err
+		})
+		if retryErr != nil {
+			klog.Errorf("failed to delete backing node %q of deleted machine %q after retries: %v", nodeName, machine.Name, retryErr)
```
takoverflow:
```suggestion
			klog.Errorf("failed to delete backing node %q of deleted machine %q after multiple retries: %v", nodeName, machine.Name, retryErr)
```

pkg/util/provider/machinecontroller/node.go
```diff
@@ -242,6 +242,70 @@ func (c *controller) getMachineFromNode(nodeName string) (*v1alpha1.Machine, err
...
+	var node *corev1.Node
+
+	// nodeName label is expected to be present on machine object, but in case it is not present
+	// we should not block deletion of machine and should move forward with flow.
+	nodeName := machine.Labels[v1alpha1.NodeLabelKey]
+	if nodeName == "" {
+		description = fmt.Sprintf("Label %q not present on machine %q, continuing deletion flow. %s", v1alpha1.NodeLabelKey, machine.Name, machineutils.InitiateFinalizerRemoval)
+		klog.Warning(description)
+		state = v1alpha1.MachineStateProcessing
+		return description, state, nil
```
takoverflow: Just for consistency with other fields and explicit clarification can you set `err = nil` and then just call return i.e.

```
err = nil
return
```

pkg/util/provider/machinecontroller/node.go
```diff
@@ -242,6 +242,70 @@ func (c *controller) getMachineFromNode(nodeName string) (*v1alpha1.Machine, err
...
+			description = fmt.Sprintf("No node object found for %q, continuing deletion flow. %s", nodeName, machineutils.InitiateFinalizerRemoval)
+			klog.Warning(description)
+			state = v1alpha1.MachineStateProcessing
+			err = nil
+		default:
+			description = fmt.Sprintf("Retrieval of Node Object %q failed due to error: %s, Retrying. %s", nodeName, err, machineutils.InitiateNodeDeletion)
+			klog.Error(description)
+			state = v1alpha1.MachineStateFailed
+		}
+		return description, state, err
```
takoverflow:
```suggestion
		return
```
Can simply be return since you've declared return values already.

pkg/util/provider/machinecontroller/node.go
```diff
@@ -242,6 +242,70 @@ func (c *controller) getMachineFromNode(nodeName string) (*v1alpha1.Machine, err
...
+		description = fmt.Sprintf("Label %q not present on machine %q, continuing deletion flow. %s", v1alpha1.NodeLabelKey, machine.Name, machineutils.InitiateFinalizerRemoval)
+		klog.Warning(description)
+		state = v1alpha1.MachineStateProcessing
+		return description, state, nil
+	}
+
+	node, err = c.nodeLister.Get(nodeName)
+	if err != nil {
+		// If node object is not found, then it could have already been deleted,
+		// therefore we should not block deletion of machine. But if error is other than NotFound,
```
takoverflow:
```suggestion
		// therefore we should not block deletion of machine. But if error is something other than NotFound,
```

pkg/util/provider/machinecontroller/node.go
```diff
@@ -242,6 +242,70 @@ func (c *controller) getMachineFromNode(nodeName string) (*v1alpha1.Machine, err
...
+	}
+
+	// Remove finalizers from node object before deletion.
+	klog.V(3).Infof("Removing finalizers from node %q associated with machine %q", nodeName, machine.Name)
+	err = c.removeNodeFinalizers(ctx, node)
+	if err != nil {
+		description = fmt.Sprintf("Removal of finalizers from Node Object %q failed due to error: %s, Retrying. %s", nodeName, err, machineutils.InitiateNodeDeletion)
+		klog.Error(description)
+		state = v1alpha1.MachineStateFailed
+		return description, state, err
```
takoverflow: Same as https://github.com/gardener/machine-controller-manager/pull/1102/changes#r3794727642

pkg/util/provider/machinecontroller/node.go
```diff
@@ -242,6 +242,70 @@ func (c *controller) getMachineFromNode(nodeName string) (*v1alpha1.Machine, err
...
+		description = fmt.Sprintf("No node object found for %q, continuing deletion flow. %s", nodeName, machineutils.InitiateFinalizerRemoval)
+		klog.Warning(description)
+		state = v1alpha1.MachineStateProcessing
+		err = nil
+	default:
+		description = fmt.Sprintf("Deletion of Node Object %q failed due to error: %s. %s", nodeName, err, machineutils.InitiateNodeDeletion)
+		klog.Error(description)
+		state = v1alpha1.MachineStateFailed
+	}
+	return description, state, err
```
takoverflow: Same as https://github.com/gardener/machine-controller-manager/pull/1102/changes#r3794727642

takoverflow (commented):
Thanks for the changes, have requested a bunch of nits PTAL. LGTM otherwise.

Also, once all the changes have been reviewed by other reviewers, can you once more do the manual testing that you initially did for this PR and share the logs/results.

pkg/util/provider/machinecontroller/machine.go
gagan16k: Would it be worth counting the number of retries and printing it in this message?

takoverflow: Does it matter how many times it was retried, we'll be printing either a warning or an error every time something errors out anyways. IMO, it doesn't really make a difference.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -100,6 +101,25 @@ func (c *controller) deleteMachine(obj any) {
 		}
 	}
 	c.enqueueMachineTermination(machine, "handling terminating machine object DELETE event")
```
r4mek: IIUC `deleteMachine()` is called when the object is removed from the etcd, so do we need to enqueue the machine here?

pkg/util/provider/machinecontroller/node.go
```diff
@@ -242,6 +242,71 @@ func (c *controller) getMachineFromNode(nodeName string) (*v1alpha1.Machine, err
 	return machines[0], nil
 }
 
+// removeFinalizersAndDeleteNode removes the MCM finalizer from and deletes the node object backed by the machine.
+func (c *controller) removeFinalizerAndDeleteNode(ctx context.Context, machine *v1alpha1.Machine) (description string, state v1alpha1.MachineState, err error) {
```
r4mek: we can do the machine status update here itself rather than returning the description, state, etc back.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -100,6 +101,25 @@ func (c *controller) deleteMachine(obj any) {
...
+	go func() {
+		retryErr := retry.OnError(retry.DefaultBackoff, func(error) bool { return true }, func() error {
+			_, _, err := c.removeFinalizerAndDeleteNode(context.Background(), machine)
+			return err
+		})
+		if retryErr != nil {
+			klog.Errorf("failed to delete backing node %q of deleted machine %q after multiple retries: %v", nodeName, machine.Name, retryErr)
+			return
+		}
+		klog.Infof("Successfully triggered deletion of backing node %q for deleted machine %q", nodeName, machine.Name)
```
thiyyakat: Wouldn't this log be misleading if the backing node object is not found? If I'm not wrong the logs will say
```
WARNING: No node object found for "<nodeName>", continuing deletion flow.
INFO: Successfully triggered deletion of backing node "<nodeName>" for deleted machine "<name>"
```

pkg/util/provider/machinecontroller/machine.go
r4mek: IIUC `deleteMachine()` is called when the object is removed from the etcd, so do we need to enqueue the machine here?

Edit: You can have log here instead to say `machine deletion successful` or something like that.

thiyyakat (commented):
Hi Gagan. I had one concern about the logs. If error can be handled differently in the callee and caller, maybe it can be improved.

pkg/util/provider/machinecontroller/machine.go
r4mek: you can remove this enqueue as well. Since, this doesn't really do much except taking space in queue.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -100,6 +101,25 @@ func (c *controller) deleteMachine(obj any) {
 		}
 	}
 	c.enqueueMachineTermination(machine, "handling terminating machine object DELETE event")
+
+	if c.targetCoreClient == nil || c.nodeLister == nil {
+		return
+	}
+
+	// Remove the MCM finalizer from and delete the backing node
```
r4mek: You would not want to do this all the time. Only when the machine object was deleted by removing the finalizer and node object was not removed. For standard flow, node object is always deleted before the machine object.

gagan16k: Yes, but we would never hit this part in the standard flow, as the node would be gone by then. I think it would be safe to assume this would only be necessary on a machine force-delete

r4mek: Essentially, we don't want to do all of these operations for standard flow. So, we should skip it.
Right now it will always call `removeFinalizerAndDeleteNode()` even when not needed.

pkg/util/provider/machinecontroller/machine.go
gagan16k: Fixed

pkg/util/provider/machinecontroller/machine.go
gagan16k: Addressed, PTAL

pkg/util/provider/machinecontroller/machine.go
gagan16k: Added log

takoverflow (commented):
/lgtm

pkg/util/provider/machinecontroller/machine.go
thiyyakat: Please correct me if I'm wrong, could this still happen?

gagan16k: I don't think it will. However, now it does not log when there is no node object found.

thiyyakat: Apologies. My checked out version of `node.go` was stale.

thiyyakat:
Thanks for the multiple rounds of changes!

/lgtm

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -100,7 +101,48 @@ func (c *controller) deleteMachine(obj any) {
...
+	}
+
+	// Remove the MCM finalizer from and delete the backing node.
+	// The orphan-VM safety net eventually removes the finalizer from nodes if this still fails.
+	nodeName := machine.Labels[v1alpha1.NodeLabelKey]
+	if nodeName == "" {
+		return
+	}
+
+	if _, err := c.nodeLister.Get(nodeName); apierrors.IsNotFound(err) {
```
r4mek, line 117: do we need this check? as in the retry function we are checking for it.

gagan16k: It can serve as a short exit path, so that we do not even spawn a goroutine in case the node is not present.

thiyyakat (approved):

PR #1107, opened 2026-06-19
Fix bug where `manageMachinePreservation` uncordons nodes of `Running` Machines unconditionally

thiyyakat:
This PR does the following:

- Fixes a bug where MCM unconditionally uncordoned the backing node of any active machine on every preservation reconcile pass, even when the node had been cordoned externally (e.g. by CA or an operator) for reasons unrelated to preservation.
- Moves the uncordon into `stopPreservationWithNode`, gated on `phase == MachineRunning`, so it only fires when preservation is actively being stopped and the machine has recovered.
- Orders the uncordon before `clearMachinePreserveExpiryTime`, so a failed uncordon is retried on the next reconcile instead of being silently dropped.
- Splits `manageMachinePreservation` into a gathering phase (`reconcileMachinePreservation`) and a transition phase (`manageMachinePreservation`).
- Adds an early exit for machines with no preservation state.
- Simplifies `getEffectivePreservationAnnotations` into a pure function.
- Removes `""` as a valid preservation annotation value.

Fixes #

IT run with `machine-controller-manager-provider-aws`

The key behavioral change: uncordoning is gated on preservation being stopped in the current reconcile iteration AND phase == MachineRunning. Machines that are actively preserved, or unrelated to preservation, will never have their node uncordoned by this flow.

**Release note**:

```bugfix operator
Fixed a bug where a cordoned node backing a non-preserved Running machine would be unconditionally uncordoned on every reconcile pass through the preservation flow.
```

### DIFF
--- a/pkg/controller/machineset.go
+++ b/pkg/controller/machineset.go
@@ -949,7 +949,7 @@ func (c *controller) manageAutoPreservationOfFailedMachines(ctx context.Context,
 	var others []*v1alpha1.Machine
 	for _, m := range machines {
 		// check if machine is already annotated for preservation, if yes, skip. Machine controller will take care of the rest.
-		if machineutils.IsMachineFailed(m) && !machineutils.PreventAutoPreserveAnnotationValues.Has(m.Annotations[machineutils.PreserveMachineAnnotationKey]) {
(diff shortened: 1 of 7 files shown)


thiyyakat:
Will reopen new PRs now that there is more clarity on what goes into the patch and what goes into master.

PR #1108, opened 2026-06-19
[rel-v0.62] Fix bug where manageMachinePreservation uncordons nodes of Running Machines unconditionally

thiyyakat:

The previous implementation called uncordonNodeIfCordoned whenever IsMachineActive returned true and a node name was present. This meant any Running machine with a cordoned backing node (cordoned for reasons unrelated to preservation, e.g: CA scale-down) would be unconditionally uncordoned on every reconcile pass. 

The PR fixes this by:
1. Introducing a check which ensures non-preservation-bound machines are not candidates for preservation-related operations
2. Performs uncordonIfCordoned only on machines for which preservation is being stopped, and the machine is in Running phase.

Fixes #

The following tests were added:
  - Preserved when-failed machine recovers to Running → node uncordoned, PreserveExpiryTime cleared.
  - Preserved when-failed machine expires while Running → node uncordoned, annotation removed.
  - Running machine with no preservation annotations and externally-cordoned node → node untouched.
  - Machine still in Failed phase with active preservation → node stays cordoned.
  - Actively-preserved machines (now, when-failed via node or machine annotation) with a cordoned node → node stays cordoned.
  - Non-preservation-bound machine → zero writes to machine status/annotations/labels/spec and to the backing node.

**Release note**:

```bugfix operator
Fixed a bug where a cordoned node backing a non-preserved Running machine would be unconditionally uncordoned on every reconcile pass through the preservation flow.
```

### DIFF
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -777,6 +777,14 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
 			return
 		}
 	}
+	machinePreserved := clone.Status.CurrentStatus.PreserveExpiryTime != nil
+	machineAnnotated := clone.Annotations[machineutils.PreserveMachineAnnotationKey] != ""
+	lastApplied := clone.Annotations[machineutils.LastAppliedNodePreserveValueAnnotationKey]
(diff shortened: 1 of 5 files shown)


thiyyakat:
Will reopen new PRs now that there is more clarity on what goes into the patch and what goes into master.

PR #1109, opened 2026-06-23
🐛 Fix machine loop block when backing node does not exist

viragvoros:
This PR fixes a bug in the machine preservation logic (`manageMachinePreservation`). Currently, the controller returns an error if the backing `Node` object does not exist yet in the cluster. 
When a machine is newly created or replacing an old instance, it takes time for the node to register. Returning an error here completely stops the machine's main update loop. This prevents the machine from changing its status phase, which causes downstream components to miss other checks.
By ignoring the `NotFound` error, the controller can assume there are no node annotations yet and continue updating the machine normally.

Fixes #

I added a new test case to the unit test table. It checks that when a node is missing from the cluster, the controller ignores the error and returns a normal `LongRetry` instead of failing.

**Release note**:
```
bugfix operator
Fixed a bug where a missing Node object stopped the machine controller loop from running, causing machine status phase updates to get stuck.
```

### DIFF
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -774,7 +774,15 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
 	if nodeName != "" {
 		nodeAnnotationValue, err = c.getNodePreserveAnnotationValue(nodeName)
 		if err != nil {
-			return
+			// If the Node object doesn't exist in the cluster yet,
+			// it cannot have a preservation annotation. Swallow the NotFound error
+			// so the machine can continue its status transition to Pending.
+			if apierrors.IsNotFound(err) {
+				err = nil
+				nodeAnnotationValue = ""
+			} else {
+				return
+			}
 		}
 	}
 	var effectivePreserveValue string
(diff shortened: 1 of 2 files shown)


gagan16k:
Was this guard intended to be on [`uncordonNodeIfCordoned()`](https://github.com/gardener/machine-controller-manager/blob/bd15a0127dce75ee533a6a917e664ffd0bab68c6/pkg/util/provider/machinecontroller/node.go#L355-L356), as `getNodePreserveAnnotation()`[already has a check for `IsNotFound`](https://github.com/gardener/machine-controller-manager/blob/bd15a0127dce75ee533a6a917e664ffd0bab68c6/pkg/util/provider/machinecontroller/node.go#L344-L348) errors in the function?

We are aware of that particular case, and are working on a fix for it. If this PR is intended to solve the same bug, we would prefer to close this in favor of the other one.
Refer to #1110 for tracking
Thanks!

Kumm-Kai:
@gagan16k 

> If this PR is intended to solve the same bug, we would prefer to close this in favor of the other one.

What/where is the "other one" PR?

thiyyakat:
@Kumm-Kai : The PR was raised only recently: https://github.com/gardener/machine-controller-manager/pull/1111. @gagan16k was unable to provide a link to it earlier.

thiyyakat:
Closing this PR in favour of https://github.com/gardener/machine-controller-manager/pull/1111

PR #1111, opened 2026-06-24
[rel-v0.62] Fix preservation reconciliation to prevent deviation from regular flow for non-preservation-bound machines and prevent machines getting stuck in creation flow beyond CreationTimeout

### LINKED ISSUE #1110: Machine stuck with no phase past `CreationTimeout`

/area robustness
/kind bug
/priority 3

**What happened**:
A machine was stuck with no phase for ~17 hours (exceeding creation timeout), when it's node failed to join the cluster. This machine was entirely missing it's status field

Root cause: The new helper`uncordonNodeIfCordoned`[returns node "X" not found errors unwrapped](https://github.com/gardener/machine-controller-manager/blob/bd15a0127dce75ee533a6a917e664ffd0bab68c6/pkg/util/provider/machinecontroller/node.go#L356-L359) instead of swallowing `IsNotFound` like `getNodePreserveAnnotationValue` does. It's invoked unconditionally from `manageMachinePreservation` (called on every reconcile, including when phase is blank), short-circuiting the entire reconcile loop before triggerCreationFlow can write any status updates (like `Phase: Pending`) on the machine.
`MachineCreationTimeout` is gated behind [Phase != ""](https://github.com/gardener/machine-controller-manager/blob/bd15a0127dce75ee533a6a917e664ffd0bab68c6/pkg/util/provider/machinecontroller/machine.go#L223-L224)  AND [Phase ∈ {Pending,Unknown,InPlaceUpdating,InPlaceUpdateFailed}](https://github.com/gardener/machine-controller-manager/blob/bd15a0127dce75ee533a6a917e664ffd0bab68c6/pkg/util/provider/machinecontroller/machine_util.go#L1069-L1071). 

- As phase is not written due to the above, empty phase escapes both gates, so the 20-minute creation timeout never fires.
- An added side-effect of this bug causes the `triggerCreationFlow` to be blocked till the node joins the cluster, and there are no status updates (`Pending` phase) on the machine until the node finally joins.

Any machine should move to `Pending` phase, and then move to `Failed` once it hits the `CreationTimeout`

**Environment**:

- Kubernetes version (use `kubectl version`): 1.32.13
- Cloud provider or hardware configuration: GCP

thiyyakat:
The previous implementation had several correctness issues in the machine preservation flow:

  1. Unconditional uncordon on every reconcile: uncordonNodeIfCordoned was called whenever IsMachineActive returned true and a node name was present. This meant any `Running` machine with a cordoned backing node — even one cordoned for reasons unrelated to preservation (e.g. Cluster Autoscaler scale-down) — would be unconditionally uncordoned on every reconcile pass.
  2. No gate for non-preservation-bound machines: Every machine was processed through the full preservation logic on every reconcile, even machines with no preservation state at all (no annotations, no PreserveExpiryTime). This caused unnecessary API calls on non-preserved machines.
  3. Stale clone causing conflict errors: The `LastAppliedNodePreserveValueAnnotation` sync in the defer compared annotations on a clone that had already been updated via UpdateStatus calls inside `preserveMachine/stopPreservationIfActive`. Because UpdateStatus returns the server's stored metadata (not the in-memory mutation), the  annotation diff was never detected and the sync was silently dropped. On paths where the sync did fire, it was using a stale ResourceVersion, producing spurious  conflict errors.  

This PR fixes all of the above by:

  - Introducing `preserveStateInfo struct` populated by `getPreserveStateInfo` at the start of each reconcile, reading node and machine annotation state in a single pass and storing it for the rest of the function. This eliminates the double lister read and decouples the annotation sync from the defer.
  - Adding an explicit `isMachinePreservationBound` gate: machines with no preservation state (no `PreserveExpiryTime`, no preserve annotation on machine or node, no `LastAppliedNodePreserveValueAnnotation`) skip the entire preservation flow and `return LongRetry, nil` immediately.
  - Moving the `LastAppliedNodePreserveValueAnnotation` sync to an explicit `updatePreserveAnnotationOnMachine` call at the end of the function via `shouldAnnotationsBeUpdatedOnMachine`, rather than a defer that compared potentially stale annotation maps.
  - Moving uncordon logic into `stopPreservationIfActive` (step 4), so it fires only when preservation is actively being stopped for a `Running machine` — not on every reconcile of any active machine. The `manageMachinePreservation` path retains a narrower uncordon call gated on `!preservationStopped && PreserveExpiryTime != nil && 
  Phase == Running` to handle the case where `preserveMachine` is called on a `Running` machine (e.g. `preserve=now` set by user).
  - Adding `IsNotFound` handling in the defer so transient not-found errors during preservation (e.g. machine deleted mid-reconcile) return `LongRetry`rather than `ShortRetry`.
  - Propagating the updated node object from `removePreservationRelatedAnnotationsOnNode` (now returns `*corev1.Node`) so that the subsequent uncordon step in `stopPreservationIfActive` operates on a current node object rather than a potentially stale one.
  - Removing the redundant `PreventAutoPreserveAnnotationValues` set in machineutils (it was identical to `AllowedPreserveAnnotationValues` minus ""); `manageAutoPreservationOfFailedMachines` in machineset.go now uses `AllowedPreserveAnnotationValues` directly.

Fixes #1110

**Release note**:

```bugfix operator
Fix preservation reconcile loop to avoid unconditionally uncordoning nodes unrelated to machine preservation, prevent spurious writes on non-preserved machines, and eliminate conflict errors caused by stale machine object comparison in the defer annotation sync.
```

### DIFF
--- a/pkg/controller/deployment_machineset_util.go
+++ b/pkg/controller/deployment_machineset_util.go
@@ -136,7 +136,7 @@ func calculateMachineSetStatus(is *v1alpha1.MachineSet, filteredMachines []*v1al
 			}
 			failedMachines = append(failedMachines, machineSummary)
 		}
-		// Count number of failed machines annotated with PreserveMachineAnnotationValuePreservedByMCM
+		// Count number of failed machines annotated with PreserveAnnotationValueAutoPreserved
 		// Cannot combine with above if block in case auto-preservation is not complete yet
 		if machine.Annotations[machineutils.PreserveMachineAnnotationKey] == machineutils.PreserveMachineAnnotationValueAutoPreserved {
 			autoPreserveFailedMachineCount++
--- a/pkg/controller/deployment_machineset_util.go
+++ b/pkg/controller/deployment_machineset_util.go
@@ -138,7 +138,7 @@ func calculateMachineSetStatus(is *v1alpha1.MachineSet, filteredMachines []*v1al
 		}
 		// Count number of failed machines annotated with PreserveMachineAnnotationValuePreservedByMCM
 		// Cannot combine with above if block in case auto-preservation is not complete yet
-		if machine.Annotations[machineutils.PreserveMachineAnnotationKey] == machineutils.PreserveMachineAnnotationValuePreservedByMCM {
+		if machine.Annotations[machineutils.PreserveMachineAnnotationKey] == machineutils.PreserveMachineAnnotationValueAutoPreserved {
 			autoPreserveFailedMachineCount++
 		}
 	}
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -752,9 +752,10 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
 		return
 	}
 	clone := machine.DeepCopy()
+	var removeAnnotations bool
 	defer func() {
 		// We compare annotation value in the clone with the original machine object to see if an update is required
-		if err == nil && clone.Annotations[machineutils.LastAppliedNodePreserveValueAnnotationKey] != machine.Annotations[machineutils.LastAppliedNodePreserveValueAnnotationKey] {
+		if err == nil && !removeAnnotations && clone.Annotations[machineutils.LastAppliedNodePreserveValueAnnotationKey] != machine.Annotations[machineutils.LastAppliedNodePreserveValueAnnotationKey] {
 			_, err = c.controlMachineClient.Machines(clone.Namespace).Update(ctx, clone, metav1.UpdateOptions{})
 			if err != nil {
 				klog.Errorf("error updating LastAppliedNodePreserveValueAnnotation value on machine %q: %v", machine.Name, err)
@@ -765,8 +766,7 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
 				klog.Warningf("Error during preservation flow:%v", err.Error())
 				retry = machineutils.LongRetry
 				err = nil
-			}
-			if apierrors.IsConflict(err) {
+			} else if apierrors.IsConflict(err) {
 				retry = machineutils.ConflictRetry
 			} else {
 				retry = machineutils.ShortRetry
@@ -796,20 +796,23 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
 	}
 	switch effectivePreserveValue {
 	case "", machineutils.PreserveMachineAnnotationValueFalse:
-		clone, err = c.stopPreservationIfActive(ctx, clone, false)
+
+		clone, err = c.stopPreservationIfActive(ctx, clone, removeAnnotations)
 	case machineutils.PreserveMachineAnnotationValueWhenFailed:
 		// on timing out, remove preserve annotation to prevent incorrect re-preservation
 		if machineutils.IsMachinePreservationExpired(clone) {
-			clone, err = c.stopPreservationIfActive(ctx, clone, true)
+			removeAnnotations = true
+			clone, err = c.stopPreservationIfActive(ctx, clone, removeAnnotations)
 		} else if !machineutils.IsMachineFailed(clone) {
-			clone, err = c.stopPreservationIfActive(ctx, clone, false)
+			clone, err = c.stopPreservationIfActive(ctx, clone, removeAnnotations)
 		} else {
 			clone, err = c.preserveMachine(ctx, clone, effectivePreserveValue)
 		}
 	case machineutils.PreserveMachineAnnotationValueNow:
 		if machineutils.IsMachinePreservationExpired(clone) {
 			// on timing out, remove preserve annotation to prevent incorrect re-preservation
-			clone, err = c.stopPreservationIfActive(ctx, clone, true)
+			removeAnnotations = true
+			clone, err = c.stopPreservationIfActive(ctx, clone, removeAnnotations)
 		} else {
 			clone, err = c.preserveMachine(ctx, clone, effectivePreserveValue)
 		}
@@ -818,7 +821,8 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
 			// To prevent incorrect re-preservation of a recovered, previously auto-preserved machine on future failures
 			// (since the autoPreserveFailedMachineCount maintained by the machineSetController, may have changed),
 			// in addition to stopping preservation, we also remove the preservation annotation on the machine.
-			clone, err = c.stopPreservationIfActive(ctx, clone, true)
+			removeAnnotations = true
+			clone, err = c.stopPreservationIfActive(ctx, clone, removeAnnotations)
 		} else {
 			clone, err = c.preserveMachine(ctx, clone, effectivePreserveValue)
 		}
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -754,27 +754,11 @@ type preserveStateInfo struct {
 	machineValue          string
 	lastAppliedNodeValue  string
 	PreserveExpiryTimeSet bool
-	effectiveValue        string
 }
 
 // manageMachinePreservation manages machine preservation based on the preserve annotation values on the node and machine objects.
 func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1alpha1.Machine) (retry machineutils.RetryPeriod, err error) {
-	var preservationBound bool
-	preservationBound, err = c.isMachinePreservationBound(ctx, machine)
-	if !preservationBound {
-		// we clear the error here to prevent preservation logic from interfering with non-preservation-bound machines
-		return machineutils.LongRetry, nil
-	}
-	clone := machine.DeepCopy()
-	var removeAnnotations bool
 	defer func() {
-		// We compare annotation value in the clone with the original machine object to see if an update is required
-		if err == nil && !removeAnnotations && clone.Annotations[machineutils.LastAppliedNodePreserveValueAnnotationKey] != machine.Annotations[machineutils.LastAppliedNodePreserveValueAnnotationKey] {
-			_, err = c.controlMachineClient.Machines(clone.Namespace).Update(ctx, clone, metav1.UpdateOptions{})
-			if err != nil {
-				klog.Errorf("error updating LastAppliedNodePreserveValueAnnotation value on machine %q: %v", machine.Name, err)
-			}
-		}
 		if err != nil {
 			if apierrors.IsNotFound(err) {
 				klog.Warningf("Error during preservation flow:%v", err.Error())
@@ -789,25 +773,30 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
 			retry = machineutils.LongRetry
 		}
 	}()
-
-	nodeName := clone.Labels[v1alpha1.NodeLabelKey]
-	nodeAnnotationValue := ""
-	if nodeName != "" {
-		nodeAnnotationValue, err = c.getNodePreserveAnnotationValue(nodeName)
-		if err != nil {
-			return
-		}
-	}
-	var effectivePreserveValue string
-	effectivePreserveValue, clone.Annotations = getEffectivePreservationAnnotations(nodeAnnotationValue, clone.Annotations)
-	if !machineutils.AllowedPreserveAnnotationValues.Has(effectivePreserveValue) {
-		if effectivePreserveValue == nodeAnnotationValue {
-			klog.Warningf("Preserve annotation value %q on node %q is invalid", effectivePreserveValue, nodeName)
-		} else {
-			klog.Warningf("Preserve annotation value %q on machine %q is invalid", effectivePreserveValue, clone.Name)
-		}
+	nodeName := machine.Labels[v1alpha1.NodeLabelKey]
+	preserveInfo, err := c.getPreserveStateInfo(machine)
+	if preserveInfo.machineAnnotated && !machineutils.AllowedPreserveAnnotationValues.Has(preserveInfo.machineValue) {
+		klog.Warningf("Preserve annotation value %q on machine %q is invalid", preserveInfo.machineValue, machine.Name)
 		return
 	}
+	if preserveInfo.nodeAnnotated && !machineutils.AllowedPreserveAnnotationValues.Has(preserveInfo.nodeValue) {
+		klog.Warningf("Preserve annotation value %q on node %q backing machine %q is invalid", preserveInfo.nodeValue, nodeName, machine.Name)
+		return
+	}
+	preservationBound := c.isMachinePreservationBound(preserveInfo)
+	if !preservationBound {
+		// we clear the error here to prevent preservation logic from interfering with non-preservation-bound machines
+		err = nil
+		return
+	} else if err != nil {
+		return
+	}
+
+	var effectivePreserveValue string
+	effectivePreserveValue = getEffectivePreservationAnnotations(preserveInfo)
+
+	var removeAnnotations, preservationStopped bool
+	clone := machine.DeepCopy()
 	switch effectivePreserveValue {
 	case "", machineutils.PreserveMachineAnnotationValueFalse:
 
@@ -849,49 +838,99 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
 	if clone.Status.CurrentStatus.Phase == v1alpha1.MachineRunning && nodeName != "" {
 		err = c.uncordonNodeIfCordoned(ctx, nodeName)
 	}
+
+	if shouldAnnotationsBeUpdatedOnMachine(removeAnnotations, preserveInfo) {
+		err = c.updatePreserveAnnotationOnMachine(ctx, preserveInfo.nodeValue, clone)
+	}
 	return
 }
 
-// getEffectivePreservationAnnotations returns the effective preservation value, and updates the machine Annotations related to preservation
-func getEffectivePreservationAnnotations(nodeAnnotationValue string, machineAnnotations map[string]string) (string, map[string]string) {
+// getEffectivePreservationAnnotations returns the effective preservation value.
+//
+// If there is no active node annotation AND no previously-applied node annotation,
+// enforce machine's preserve annotation.
+// Otherwise, the node annotation takes precedence (even if now empty/removed).
+//
+// This is required to handle the following scenario:
+//
+//	T1: Node and Machine both have the same annotation with the same value. (MCM is up and running).
+//	T2 (T2 > T1): MCM went down.
+//	T3 (T3 > T2): Node annotation was removed.
+//	T4 (T4 > T3): MCM came back up.
+//	At T4 it sees a Node with no preserve annotation but a Machine with a preserve annotation.
+//	It continues to preserve the machine.
+func getEffectivePreservationAnnotations(info *preserveStateInfo) string {
 	// If there is no active node annotation AND no previously-applied node annotation,
 	// enforce machine's preserve annotation.
 	// Otherwise, the node annotation takes precedence (even if now empty/removed).
-	if nodeAnnotationValue == "" && machineAnnotations[machineutils.LastAppliedNodePreserveValueAnnotationKey] == "" {
-		return machineAnnotations[machineutils.PreserveMachineAnnotationKey], machineAnnotations
+	if info.nodeValue == "" && info.lastAppliedNodeValue == "" {
+		return info.machineValue
 	}
-	clonedMachineAnnotations := make(map[string]string)
-	maps.Copy(clonedMachineAnnotations, machineAnnotations)
-	delete(clonedMachineAnnotations, machineutils.PreserveMachineAnnotationKey)
-	clonedMachineAnnotations[machineutils.LastAppliedNodePreserveValueAnnotationKey] = nodeAnnotationValue
-	return nodeAnnotationValue, clonedMachineAnnotations
+	return info.nodeValue
 }
 
 func (c *controller) isMachinePreservationBound(ctx context.Context, machine *v1alpha1.Machine) (bool, error) {
-	var err error
-	var machineAnnotated, nodeAnnotated bool
-	var laNodeAV string
-	var node *corev1.Node
-
-	if machine.Annotations != nil {
-		_, machineAnnotated = machine.Annotations[machineutils.PreserveMachineAnnotationKey]
-		laNodeAV = machine.Annotations[machineutils.LastAppliedNodePreserveValueAnnotationKey]
+func (c *controller) isMachinePreservationBound(info *preserveStateInfo) bool {
+	// if machine has no preservation state, the machine is not preservation-bound
+	if !info.PreserveExpiryTimeSet && !info.nodeAnnotated && !info.machineAnnotated && info.lastAppliedNodeValue == "" {
+		return false
 	}
+	return true
+}
 
+func (c *controller) getPreserveStateInfo(machine *v1alpha1.Machine) (*preserveStateInfo, error) {
+	var info preserveStateInfo
+	var err error
+	if machine.Annotations != nil {
+		info.machineValue, info.machineAnnotated = machine.Annotations[machineutils.PreserveMachineAnnotationKey]
+		info.lastAppliedNodeValue = machine.Annotations[machineutils.LastAppliedNodePreserveValueAnnotationKey]
+	}
+	if machine.Status.CurrentStatus.PreserveExpiryTime != nil {
+		info.PreserveExpiryTimeSet = true
+	}
 	nodeName := machine.Labels[v1alpha1.NodeLabelKey]
 	if nodeName != "" {
+		var node *corev1.Node
 		node, err = c.nodeLister.Get(nodeName)
 		if err != nil {
 			if apierrors.IsNotFound(err) {
 				err = nil
 			}
-		} else {
-			_, nodeAnnotated = node.Annotations[machineutils.PreserveMachineAnnotationKey]
+			return &info, err
 		}
+		info.nodeValue, info.nodeAnnotated = node.Annotations[machineutils.PreserveMachineAnnotationKey]
 	}
-	// if machine has no preservation state, the machine is not preservation-bound
-	if machine.Status.CurrentStatus.PreserveExpiryTime == nil && !nodeAnnotated && !machineAnnotated && laNodeAV == "" {
-		return false, err
+	return &info, err
+}
+
+// shouldAnnotationsBeUpdatedOnMachine returns true when the machine's annotation tracking needs
+// to be synced after a preservation action.
+func shouldAnnotationsBeUpdatedOnMachine(removeAnnotations bool, preserveInfo *preserveStateInfo) bool {
+	// annotations were already removed by stopPreservationIfActive — nothing left to sync
+	if removeAnnotations {
+		return false
 	}
-	return true, err
+	// node annotation is not in control — machine annotation prevails, no sync needed
+	if !preserveInfo.nodeAnnotated && preserveInfo.lastAppliedNodeValue == "" {
+		return false
+	}
+	// node value is unchanged and machine has no annotation to clear — nothing has changed
+	if preserveInfo.nodeValue == preserveInfo.lastAppliedNodeValue && !preserveInfo.machineAnnotated {
+		return false
+	}
+	return true
+}
+
+// updatePreserveAnnotationOnMachine clears the machine's PreserveMachineAnnotationKey and sets
+// LastAppliedNodePreserveValueAnnotationKey to nodeValue.
+func (c *controller) updatePreserveAnnotationOnMachine(ctx context.Context, nodeValue string, machine *v1alpha1.Machine) error {
+	clone := machine.DeepCopy()
+	if clone.Annotations == nil {
+		clone.Annotations = make(map[string]string)
+	} else {
+		delete(clone.Annotations, machineutils.PreserveMachineAnnotationKey)
+	}
+	clone.Annotations[machineutils.LastAppliedNodePreserveValueAnnotationKey] = nodeValue
+	_, err := c.controlMachineClient.Machines(clone.Namespace).Update(ctx, clone, metav1.UpdateOptions{})
+	return err
 }
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -799,24 +799,21 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
 		klog.Warningf("Couldn't find node %q for machine %q", nodeName, machine.Name)
 		err = nil
 	}
-	// Note: when the backing node  cannot be found, we assume the machine's annotation value needs to be enforced to enable
+	// Note: when the backing node cannot be found, we assume the machine's annotation value needs to be enforced to enable
 	// preservation of the machine object.
 	effectivePreserveValue := getEffectivePreservationAnnotations(preserveInfo, getErr)
 
-	var removeAnnotations, preservationStopped bool
+	var removeAnnotations bool
 	clone := machine.DeepCopy()
 	switch effectivePreserveValue {
 	case "", machineutils.PreserveMachineAnnotationValueFalse:
-		preservationStopped = true
 		clone, err = c.stopPreservationIfActive(ctx, clone, removeAnnotations)
 	case machineutils.PreserveMachineAnnotationValueWhenFailed:
 		// on timing out, remove preserve annotation to prevent incorrect re-preservation
 		if machineutils.IsMachinePreservationExpired(clone) {
 			removeAnnotations = true
-			preservationStopped = true
 			clone, err = c.stopPreservationIfActive(ctx, clone, removeAnnotations)
 		} else if !machineutils.IsMachineFailed(clone) {
-			preservationStopped = true
 			clone, err = c.stopPreservationIfActive(ctx, clone, removeAnnotations)
 		} else {
 			clone, err = c.preserveMachine(ctx, clone, effectivePreserveValue)
@@ -825,7 +822,6 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
 		if machineutils.IsMachinePreservationExpired(clone) {
 			// on timing out, remove preserve annotation to prevent incorrect re-preservation
 			removeAnnotations = true
-			preservationStopped = true
 			clone, err = c.stopPreservationIfActive(ctx, clone, removeAnnotations)
 		} else {
 			clone, err = c.preserveMachine(ctx, clone, effectivePreserveValue)
@@ -836,7 +832,6 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
 			// (since the autoPreserveFailedMachineCount maintained by the machineSetController, may have changed),
 			// in addition to stopping preservation, we also remove the preservation annotation on the machine.
 			removeAnnotations = true
-			preservationStopped = true
 			clone, err = c.stopPreservationIfActive(ctx, clone, removeAnnotations)
 		} else {
 			clone, err = c.preserveMachine(ctx, clone, effectivePreserveValue)
@@ -845,9 +840,9 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
 	if err != nil {
 		return
 	}
-	// For the preserveMachine path (preservation still active), uncordon the node if the machine
-	// has recovered to Running. The stopPreservationIfActive path handles uncordon internally.
-	if !preservationStopped && clone.Status.CurrentStatus.PreserveExpiryTime != nil && clone.Status.CurrentStatus.Phase == v1alpha1.MachineRunning && nodeName != "" {
+	// For the preserve=now path (preservation still active), uncordon the node if the machine
+	// has recovered to Running. For all other cases, stopPreservationIfActive path handles uncordon internally.
+	if clone.Status.CurrentStatus.PreserveExpiryTime != nil && clone.Status.CurrentStatus.Phase == v1alpha1.MachineRunning && nodeName != "" {
 		var node *corev1.Node
 		node, err = c.nodeLister.Get(nodeName)
 		if err != nil {
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -744,6 +744,19 @@ func (c *controller) isCreationProcessing(machine *v1alpha1.Machine) bool {
 	Machine Preservation operations
 */
 
+// preserveStateInfo encapsulates the preservation annotation values found
+// on the machine and node objects, along with the effective preservation value for the machine
+// and the last applied node preserve value by MCM.
+type preserveStateInfo struct {
+	nodeAnnotated         bool
+	machineAnnotated      bool
+	nodeValue             string
+	machineValue          string
+	lastAppliedNodeValue  string
+	PreserveExpiryTimeSet bool
+	effectiveValue        string
+}
+
 // manageMachinePreservation manages machine preservation based on the preserve annotation values on the node and machine objects.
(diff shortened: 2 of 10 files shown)


pkg/util/provider/machineutils/utils.go
elankath: after we changed literal value we should change the name to the constant to also `PreserveMachineAnnotationValueAutoPreserved` instead. This package will be exposed to consumers like DWD so best to get semantic name changes finalized.

thiyyakat: Addressed in https://github.com/gardener/machine-controller-manager/pull/1111/changes/200ce00ac429dea837992b1ab47851f961a6d0f2

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2419,17 +2419,21 @@ func (c *controller) stopPreservationIfActive(ctx context.Context, machine *v1al
 		// if node is not found and error is simply returned, then preservation will never be stopped on machine
 		// therefore, this error is handled specifically
 		if apierrors.IsNotFound(err) {
-			// Node not found, proceed to clear preserveExpiryTime on machine
-			klog.Warningf("Node %q of machine %q not found. Proceeding to clear PreserveExpiryTime on machine.", nodeName, machine.Name)
-			_, err := c.clearMachinePreserveExpiryTime(ctx, machine)
+			klog.Warningf("Node %q of machine %q not found. Proceeding to stop preservation on machine.", nodeName, machine.Name)
+			// Node not found, proceed to delete annotations and clear preserveExpiryTime on machine
+			machine, err = c.removePreserveAnnotationOnMachine(ctx, machine)
```
gagan16k: Since this is happening unconditionally, this could remove annotations off machines even when `removePreservationAnnotations` is false. Would that be acceptable if the node is gone?

Additionally what if a machine which has just entered creation flow (node has not joined) has been preserved using `preserve:now`. This code flow would remove the annotation from this machine.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -768,76 +773,182 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
+		return
+	}
+	preservationBound := c.isMachinePreservationBound(preserveInfo)
+	if !preservationBound {
+		// we clear the error here to prevent preservation logic from interfering with non-preservation-bound machines
+		err = nil
+		return
+	} else if err != nil {
+		return
+	}
```
gagan16k, line 801:
```suggestion
    // machine has no preservation state; defer maps err==nil to LongRetry.
	if !preservationBound {
		return
	} 
```

It appears as if the `err` will always be nil here, as the only above this which sets `err` (returned from `getPreserveStateInfo`) checks if it is nil, and returns if it is.

pkg/controller/deployment_machineset_util.go
```diff
@@ -138,7 +138,7 @@ func calculateMachineSetStatus(is *v1alpha1.MachineSet, filteredMachines []*v1al
 		}
 		// Count number of failed machines annotated with PreserveMachineAnnotationValuePreservedByMCM
```
gagan16k:
```suggestion
		// Count number of failed machines annotated with PreserveMachineAnnotationValueAutoPreserved
```

nit

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -744,22 +744,27 @@ func (c *controller) isCreationProcessing(machine *v1alpha1.Machine) bool {
...
+// preserveStateInfo encapsulates the preservation annotation values found
+// on the machine and node objects, along with the effective preservation value for the machine
+// and the last applied node preserve value by MCM.
+type preserveStateInfo struct {
+	nodeAnnotated         bool
+	machineAnnotated      bool
+	nodeValue             string
+	machineValue          string
+	lastAppliedNodeValue  string
+	PreserveExpiryTimeSet bool
```
gagan16k:
```suggestion
	preserveExpiryTimeSet bool
```
nit, as it is not exported

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -768,76 +773,182 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
-			klog.Warningf("Preserve annotation value %q on machine %q is invalid", effectivePreserveValue, clone.Name)
-		}
+	if preserveInfo.machineAnnotated && !machineutils.AllowedPreserveAnnotationValues.Has(preserveInfo.machineValue) {
+		klog.Warningf("Preserve annotation %q=%q on machine %q is invalid", machineutils.PreserveMachineAnnotationKey, preserveInfo.machineValue, machine.Name)
 		return
 	}
+	if preserveInfo.nodeAnnotated && !machineutils.AllowedPreserveAnnotationValues.Has(preserveInfo.nodeValue) {
+		klog.Warningf("Preserve annotation %q=%q on node %q backing machine %q is invalid", machineutils.PreserveMachineAnnotationKey, preserveInfo.nodeValue, nodeName, machine.Name)
+		return
+	}
```
gagan16k, line 788: As node annotations have been given priority over machine annotations for preservation, having an invalid preservation annotation on the machine here can block this function, even if a valid annotation is placed on the node.
This is a minor issue, but it can be easily fixed by checking if the node annotation is valid first, and skipping the check on machine annotation if it is.

pkg/util/provider/machinecontroller/node.go
```diff
@@ -340,40 +340,26 @@ func addedOrRemovedEssentialTaints(oldNode, node *corev1.Node, taintKeys []strin
...
-	}
-	return node.Annotations[machineutils.PreserveMachineAnnotationKey], nil
-}
-
-func (c *controller) uncordonNodeIfCordoned(ctx context.Context, nodeName string) error {
-	node, err := c.nodeLister.Get(nodeName)
-	if err != nil {
-		return err
-	}
+func (c *controller) uncordonNodeIfCordoned(ctx context.Context, node *corev1.Node) error {
```
gagan16k, line 343: Should there be a check for `ClusterAutoscalerScaleDownDisabledAnnotationByMCMKey` annotation before node operations, similar to how it is done in `removePreservationRelatedAnnotationsOnNode()`?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -768,76 +773,182 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
+	preservationBound := c.isMachinePreservationBound(preserveInfo)
+	if !preservationBound {
+		// we clear the error here to prevent preservation logic from interfering with non-preservation-bound machines
+		err = nil
```
aaronfern, line 792: Why do you need to set `err` to `nil` here?
I see `err` gets populated before this only when calling `c.getPreserveStateInfo()`, and even then if `err != nil`, we return

pkg/util/provider/machinecontroller/machine.go
aaronfern: Similar here. `err` is populated only by `c.getPreserveStateInfo()`, and there we return if `err != nil`

pkg/util/provider/machinecontroller/machine.go
aaronfern: Need not be exported
```suggestion
	preserveExpiryTimeSet bool
```

pkg/util/provider/machinecontroller/machine_util.go
aaronfern: Is there a reason this is not gated with `removePreservationAnnotations`? I see that every other invocation is `removePreserveAnnotationOnMachine()` within this function is gated

pkg/util/provider/machineutils/utils.go
```diff
@@ -108,23 +109,19 @@ const (
...
 	PreserveMachineAnnotationValueFalse = "false"
 )
 
 // AllowedPreserveAnnotationValues contains the allowed values for the preserve annotation
-var AllowedPreserveAnnotationValues = sets.New(PreserveMachineAnnotationValueNow, PreserveMachineAnnotationValueWhenFailed, PreserveMachineAnnotationValuePreservedByMCM, PreserveMachineAnnotationValueFalse, "")
-
-// PreventAutoPreserveAnnotationValues contains the values to check if a machine is already annotated for preservation,
-// in which case it should not be auto-preserved.
-var PreventAutoPreserveAnnotationValues = sets.New(PreserveMachineAnnotationValueNow, PreserveMachineAnnotationValueWhenFailed, PreserveMachineAnnotationValuePreservedByMCM, PreserveMachineAnnotationValueFalse)
+var AllowedPreserveAnnotationValues = sets.New(PreserveMachineAnnotationValueNow, PreserveMachineAnnotationValueWhenFailed, PreserveMachineAnnotationValueAutoPreserved, PreserveMachineAnnotationValueFalse)
```
aaronfern, line 124: Is `""` removed intentionally?

pkg/util/provider/machinecontroller/machine_test.go
```diff
@@ -4281,25 +4302,13 @@ var _ = Describe("machine", func() {
 						Type:   v1alpha1.NodePreserved,
 						Status: corev1.ConditionTrue},
 					retry:                  machineutils.LongRetry,
-					machineAnnotationValue: machineutils.PreserveMachineAnnotationValuePreservedByMCM,
-				},
-			}),
-			Entry("when node is annotated and preservation times out, should stop preservation", testCase{
```
aaronfern, line 4287: Why is this test removed? This looks like a valid case

pkg/util/provider/machinecontroller/machine_test.go
```diff
@@ -4323,23 +4333,10 @@ var _ = Describe("machine", func() {
 					retry:                   machineutils.LongRetry,
 				},
 			}),
-			Entry("when invalid preserve annotation is added on node of un-preserved machine, should do nothing  ", testCase{
```
aaronfern, line 4326: This looks like a valid case too

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -768,76 +773,182 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
+
+func (c *controller) isMachinePreservationBound(info *preserveStateInfo) bool {
+	// if machine has no preservation state, the machine is not preservation-bound
+	if !info.PreserveExpiryTimeSet && !info.nodeAnnotated && !info.machineAnnotated && info.lastAppliedNodeValue == "" {
+		return false
+	}
+	return true
+}
+
+func (c *controller) getPreserveStateInfo(machine *v1alpha1.Machine) (*preserveStateInfo, error) {
```
aaronfern: Please consider writing some unit tests for the new helper methods that you introduced

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -744,22 +744,27 @@ func (c *controller) isCreationProcessing(machine *v1alpha1.Machine) bool {
...
-		if err == nil && !machineAnnotationsSynced && clone.Annotations[machineutils.LastAppliedNodePreserveValueAnnotationKey] != machine.Annotations[machineutils.LastAppliedNodePreserveValueAnnotationKey] {
-			_, err = c.controlMachineClient.Machines(clone.Namespace).Update(ctx, clone, metav1.UpdateOptions{})
-			if err != nil {
-				klog.Errorf("error updating LastAppliedNodePreserveValueAnnotation value on machine %q: %v", machine.Name, err)
-			}
-		}
 		if err != nil {
-			if apierrors.IsConflict(err) {
+			if apierrors.IsNotFound(err) {
+				klog.Warningf("Error during preservation flow:%v", err.Error())
```
aaronfern: nit 
```suggestion
				klog.Warningf("Error during preservation flow: %v", err.Error())
```

gagan16k (commented):
Thanks for the fixes!

pkg/controller/deployment_machineset_util.go
thiyyakat: Thanks for catching this!

pkg/util/provider/machinecontroller/machine.go
thiyyakat: You're right. The earlier return defeats the purpose of this. I will probably need to buffer the error until we can determine whether or not the machine is preservation-bound. Thanks

pkg/util/provider/machinecontroller/machine.go
thiyyakat: You're right. I need to buffer the error  from [earlier](https://github.com/gardener/machine-controller-manager/pull/1111/changes/BASE..cf6d15fe683d504b7501cff09ae554d04ad7f764#diff-f3b8457c00ac7e0985403ca50f82c97e83e70d13431e96e03e23fa262290cc47R777) and only then this logic makes sense.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: ~If we don't do this, the machine will always be assumed to be preservation-bound every cycle, because of the LastAppliedNodeAnnotationValue annotation on the machine, which will never get removed.~

~Another way we can do this is by having a different function that only deletes the LastAppliedNodeAnnotationValue from the machine unconditionally, but deletes the PreserveMachineAnnotationKey only if `removeAnnotations` is true. WDYT?~

Update: checked the flow, the annotation value gets cleared, so this should not be done unconditionally.

pkg/util/provider/machinecontroller/node.go
thiyyakat: Do you mean to gate the uncordoning? Do you feel the preservation-bound check and Running phase checks are insufficient?

pkg/util/provider/machineutils/utils.go
thiyyakat: Yes! There is no use for `preserve=""`. The user can either delete the annotation, or set it to `false`.

pkg/util/provider/machinecontroller/machine_test.go
thiyyakat: The test following this one covers this test case and also checks for annotation removal.

pkg/util/provider/machinecontroller/machine_test.go
thiyyakat: The test following this one is a duplicate of this one.

pkg/util/provider/machinecontroller/machine.go
thiyyakat: Addressed in https://github.com/gardener/machine-controller-manager/pull/1111/changes/cb105e790a1155672570a97a19413b59b45b2b66. The error from `getPreserveStateInfo` is now buffered in `getErr` and is only returned if machine is preservation-bound.

pkg/util/provider/machinecontroller/machine.go
thiyyakat: Addressed in https://github.com/gardener/machine-controller-manager/pull/1111/changes/cb105e790a1155672570a97a19413b59b45b2b66

pkg/util/provider/machinecontroller/machine.go
thiyyakat: Addressed in https://github.com/gardener/machine-controller-manager/pull/1111/changes/cb105e790a1155672570a97a19413b59b45b2b66.

pkg/util/provider/machinecontroller/machine.go
thiyyakat: Addressed in https://github.com/gardener/machine-controller-manager/pull/1111/changes/cb105e790a1155672570a97a19413b59b45b2b66

pkg/util/provider/machinecontroller/machine.go
thiyyakat: Addressed in https://github.com/gardener/machine-controller-manager/pull/1111/changes/cb105e790a1155672570a97a19413b59b45b2b66

pkg/util/provider/machinecontroller/machine.go
thiyyakat: Addressed inhttps://github.com/gardener/machine-controller-manager/pull/1111/changes/cb105e790a1155672570a97a19413b59b45b2b66. The error from `getPreserveStateInfo` is buffered and only returned if machine is preservation bound. It is cleared otherwise.

pkg/controller/deployment_machineset_util.go
thiyyakat: Addressed in https://github.com/gardener/machine-controller-manager/pull/1111/changes/cb105e790a1155672570a97a19413b59b45b2b66.

pkg/util/provider/machinecontroller/machine.go
thiyyakat: Addressed in https://github.com/gardener/machine-controller-manager/pull/1111/changes/cb105e790a1155672570a97a19413b59b45b2b66.

pkg/util/provider/machinecontroller/machine.go
thiyyakat: Addressed in https://github.com/gardener/machine-controller-manager/pull/1111/changes/cb105e790a1155672570a97a19413b59b45b2b66. If machine annotation value is invalid, we clear the value in `preserveInfo` so that the effective value can be computed accordingly.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Addressed in https://github.com/gardener/machine-controller-manager/pull/1111/changes/cb105e790a1155672570a97a19413b59b45b2b66.

thiyyakat: Addressed in https://github.com/gardener/machine-controller-manager/pull/1111/changes/cb105e790a1155672570a97a19413b59b45b2b66

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -768,76 +773,190 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
+	if machine.Status.CurrentStatus.PreserveExpiryTime != nil {
+		info.preserveExpiryTimeSet = true
+	}
+	nodeName := machine.Labels[v1alpha1.NodeLabelKey]
+	if nodeName != "" {
+		node, err := c.nodeLister.Get(nodeName)
+		if err != nil {
+			return &info, err
+		}
+		info.nodeValue, info.nodeAnnotated = node.Annotations[machineutils.PreserveMachineAnnotationKey]
```
gagan16k, line 919: Since `""` is not a valid preservation value anymore, `info.nodeAnnotated` will be set to true on empty string as well. While it will be caught correctly in the checks later, should we add a check here?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -768,76 +773,190 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
-	if nodeName != "" {
-		nodeAnnotationValue, err = c.getNodePreserveAnnotationValue(nodeName)
-		if err != nil {
-			return
-		}
+	nodeName := machine.Labels[v1alpha1.NodeLabelKey]
+	// We buffer the error returned here until we can tell if the machine is preservation-bound.
+	preserveInfo, getErr := c.getPreserveStateInfo(machine)
+	if preserveInfo == nil { // machine object not annotated with preservation state and error fetching node object
+		return
```
gagan16k: This can never be true as it returns a pointer from  `&info`. It can be deleted, as the case it is supposed to guard is checked during `isPreservationBound`

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -768,76 +773,190 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
-	delete(clonedMachineAnnotations, machineutils.PreserveMachineAnnotationKey)
-	clonedMachineAnnotations[machineutils.LastAppliedNodePreserveValueAnnotationKey] = nodeAnnotationValue
-	return nodeAnnotationValue, clonedMachineAnnotations
+	if info.nodeValue == "" && info.lastAppliedNodeValue == "" {
+		return info.machineValue
+	}
+	return info.nodeValue
+}
+
+func (c *controller) isMachinePreservationBound(info *preserveStateInfo) bool {
```
gagan16k:
```suggestion
func isMachinePreservationBound(info *preserveStateInfo) bool {
```

Similar to `getEffectivePreservationAnnotations`, it doesn't use the reciever

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -768,76 +773,190 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
 	}
 	if err != nil {
 		return
 	}
-	// This is to handle the case where a preserved machine recovers from Failed to Running
-	// in which case, pods should be allowed to be scheduled onto the node
-	if machineutils.IsMachineActive(clone) && nodeName != "" {
-		err = c.uncordonNodeIfCordoned(ctx, nodeName)
+	// For the preserveMachine path (preservation still active), uncordon the node if the machine
+	// has recovered to Running. The stopPreservationIfActive path handles uncordon internally.
```
gagan16k: Similar to how `stopPreservationIfActive` handles uncordon, could this be moved into `preserveMachine` itself?
Or is there some other case which I am not thinking of?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -768,76 +773,190 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
+		}
+		klog.Warningf("Couldn't find node %q for machine %q", nodeName, machine.Name)
+		err = nil
+	}
+	// Note: when the backing node  cannot be found, we assume the machine's annotation value needs to be enforced to enable
+	// preservation of the machine object.
+	effectivePreserveValue := getEffectivePreservationAnnotations(preserveInfo, getErr)
+
+	var removeAnnotations, preservationStopped bool
+	clone := machine.DeepCopy()
```
gagan16k, line 807: Is the deepcopy required, as `getEffectivePreservationAnnotations` no longer mutates the annotation map.
This can directly use the `*machine` object from the args

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -768,76 +773,190 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
+			return
+		}
+		klog.Warningf("Couldn't find node %q for machine %q", nodeName, machine.Name)
+		err = nil
+	}
+	// Note: when the backing node  cannot be found, we assume the machine's annotation value needs to be enforced to enable
+	// preservation of the machine object.
+	effectivePreserveValue := getEffectivePreservationAnnotations(preserveInfo, getErr)
+
+	var removeAnnotations, preservationStopped bool
```
gagan16k: Are these two variables required?
`preservationStopped` can be removed, as `CurrentStatus.PreserveExpiryTime` is removed by `stopPreservationIfActive`, and the check for the time is enough in `L853` IMO, but with `removeAnnotations` it is not as obvious and the changes might be more involved.

pkg/util/provider/machinecontroller/node.go
gagan16k: I think it might be worth adding, as the more we can limit CA interactions with this the better. However, could ask for a second opinion

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -744,22 +744,27 @@ func (c *controller) isCreationProcessing(machine *v1alpha1.Machine) bool {
...
-			_, err = c.controlMachineClient.Machines(clone.Namespace).Update(ctx, clone, metav1.UpdateOptions{})
-			if err != nil {
-				klog.Errorf("error updating LastAppliedNodePreserveValueAnnotation value on machine %q: %v", machine.Name, err)
-			}
-		}
 		if err != nil {
-			if apierrors.IsConflict(err) {
+			if apierrors.IsNotFound(err) {
+				klog.Warningf("Error during preservation flow: %v", err.Error())
+				retry = machineutils.LongRetry
```
gagan16k, line 765: Might be too large of a delay for a `LongRetry`,  `ShortRetry` could be used even if a node is not found.
Not sure about this, however

gagan16k (commented):
Some more comments, PTAL

pkg/util/provider/machinecontroller/machine.go
thiyyakat: I'd rather not do it again here. Since we are doing the validity check nearly immediately after this function I don't see a potential issue in skipping the duplicate check. WDYT?

pkg/util/provider/machinecontroller/machine.go
thiyyakat: This can be done. I will do it as a part of the PR to change the drain from cordon+drain to taint+drain. Since this is a hotfix, unless buggy, I would rather not touch this bit. Is that fine? Or do you think there is a potential for issues here?

pkg/util/provider/machinecontroller/machine.go
thiyyakat: The deep copy is required for the remaining preservation functions such as `stopPreservationIfActive` and `preserveMachine` since they update the machine object and we also need the updated machine object later to update the annotations.

pkg/util/provider/machinecontroller/machine.go
thiyyakat: removeAnnotations serves 2 purposes:
1. It indicates to stopPreservationIfActive whether the machine and not should be unannotated - in the cases where machine preservation expires, or auto-preserved machine has recovered to Running.
2. if annotations have been removed by stopPreservationIfActive, updatePreserveAnnotationOnMachine should not run because it will again set LastAppliedNodePreserveValueAnnotation.

pkg/util/provider/machinecontroller/machine.go
thiyyakat: Since NotFound error is cleared, the retry period doesn't matter. `triggerCreationFlow` may be called next and if it encounters an error, it will set the retry to a suitable period. If it isn't called, then anyway because there is no error, LongRetry is returned from `reconcileClusterMachine`. This way the normal flow is not disrupted because of the preservation flow.

pkg/util/provider/machinecontroller/node.go
thiyyakat: The two may not be related though. If we do what you suggest, then there is a chance that the _user_ has set `scale-down-disabled` on the node in which case we would never uncordon the node even after it recovers from Failed to Running. 

The check in `removePreservationRelatedAnnotationsOnNode` was done to ensure that we don't clear the scale-down-disabled annotation if it has been set by the user.

We limit CA interaction by :
1. By-passing preservation logic if machine is not preservation-bound
2. Ensuring CA does not scale down a preserved node. We ensure preserved nodes have the scale-down-disabled annotation and in CA's ForceDeleteNode, we skip nodes if they have the scale-down-disabled annotation. 

@aaronfern do you think this is sufficient?

pkg/util/provider/machinecontroller/machine.go
thiyyakat: Addressed in https://github.com/gardener/machine-controller-manager/pull/1111/changes/91b947694df96f6ab89e8b8e129d175a5efa870e.

pkg/util/provider/machinecontroller/machine.go
thiyyakat: Addressed in https://github.com/gardener/machine-controller-manager/pull/1111/changes/91b947694df96f6ab89e8b8e129d175a5efa870e.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -744,22 +744,27 @@ func (c *controller) isCreationProcessing(machine *v1alpha1.Machine) bool {
...
+// on the machine and node objects, along with the effective preservation value for the machine
+// and the last applied node preserve value by MCM.
+type preserveStateInfo struct {
+	nodeAnnotated         bool
+	machineAnnotated      bool
+	nodeValue             string
+	machineValue          string
+	lastAppliedNodeValue  string
+	preserveExpiryTimeSet bool
+}
```
r4mek, line 757: Since `""` annotation value is not valid anymore, we can remove `nodeAnnotated` and `machineAnnotated` booleans. Annotation value is `""` implies that annotation is not set.

Also, can we rename `nodeValue`, `machineValue` to `nodeAnnotationValue`, `machineAnnotationValue` resp.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -768,76 +773,187 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
-	return nodeAnnotationValue, clonedMachineAnnotations
+	if info.nodeValue == "" && info.lastAppliedNodeValue == "" {
+		return info.machineValue
+	}
+	return info.nodeValue
+}
+
+func isMachinePreservationBound(info *preserveStateInfo) bool {
+	// if machine has no preservation state, the machine is not preservation-bound
+	if !info.preserveExpiryTimeSet && !info.nodeAnnotated && !info.machineAnnotated && info.lastAppliedNodeValue == "" {
```
r4mek, line 898: I am a bit confused. So in the parent function we allow the machine to not have preserve annotation, but the node must have preservation annotation.

But then here, we don't allow machine to be unannotated.

Also, why is it required for `lastAppliedNodeValue` to not be empty.

So, essentially we don't require this function and a simple `info.preserveExpiryTimeSet` check would suffice.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -768,76 +773,187 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
+		err = nil
 		return
+	} else if getErr != nil {
+		if !apierrors.IsNotFound(getErr) {
+			err = getErr
+			return
+		}
+		klog.Warningf("Couldn't find node %q for machine %q", nodeName, machine.Name)
+		err = nil
 	}
```
r4mek, line 801: You can move this check right after `isMachinePreservationBound()` call.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -768,76 +773,187 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
 	// Otherwise, the node annotation takes precedence (even if now empty/removed).
-	if nodeAnnotationValue == "" && machineAnnotations[machineutils.LastAppliedNodePreserveValueAnnotationKey] == "" {
-		return machineAnnotations[machineutils.PreserveMachineAnnotationKey], machineAnnotations
-	}
-	clonedMachineAnnotations := make(map[string]string)
-	maps.Copy(clonedMachineAnnotations, machineAnnotations)
-	delete(clonedMachineAnnotations, machineutils.PreserveMachineAnnotationKey)
-	clonedMachineAnnotations[machineutils.LastAppliedNodePreserveValueAnnotationKey] = nodeAnnotationValue
-	return nodeAnnotationValue, clonedMachineAnnotations
+	if info.nodeValue == "" && info.lastAppliedNodeValue == "" {
```
r4mek, line 890: Why do we have to also check for `lastAppliedNodeValue`?

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2376,41 +2376,41 @@ func (c *controller) preserveMachine(ctx context.Context, machine *v1alpha1.Mach
...
 }
 
 // stopPreservationIfActive stops the preservation of the machine and node, if preserved, and returns true if machine object has been updated
-func (c *controller) stopPreservationIfActive(ctx context.Context, machine *v1alpha1.Machine, removePreservationAnnotations bool) (bool, error) {
+func (c *controller) stopPreservationIfActive(ctx context.Context, machine *v1alpha1.Machine, removePreservationAnnotations bool) (*v1alpha1.Machine, error) {
+	var err error
 	// removal of preserveExpiryTime is the last step of stopping preservation
 	// therefore, if preserveExpiryTime is not set, machine is not preserved
 	nodeName := machine.Labels[v1alpha1.NodeLabelKey]
 	if machine.Status.CurrentStatus.PreserveExpiryTime == nil {
```
r4mek, line 2396: We have the `info.preserveExpiryTimeSet` and we already check for it in the parent function. So, no need for this.

If we still want to be check it again then shouldn't we clear the annotations before returning?

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2376,41 +2376,41 @@ func (c *controller) preserveMachine(ctx context.Context, machine *v1alpha1.Mach
...
-		machine, err := c.clearMachinePreserveExpiryTime(ctx, machine)
+		machine, err = c.clearMachinePreserveExpiryTime(ctx, machine)
 		if err != nil {
-			return false, err
+			return nil, err
 		}
 		klog.V(2).Infof("Preservation of machine %q with no backing node has stopped.", machine.Name)
-		return true, nil
+		return machine, nil
 	}
```
r4mek, line 2414: We can extract this block outside, so that we always remove preserve annotation from machine and clear clear machine expiry time. Later we can return if `nodeName == nil`

Then we won't have to do the same operation again if node is `NotFound`.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2441,27 +2446,33 @@ func (c *controller) stopPreservationIfActive(ctx context.Context, machine *v1al
 	// Step 1: change node condition to reflect that preservation has stopped
 	updatedNode, err := nodeops.AddOrUpdateConditionsOnNode(ctx, c.targetCoreClient, node.Name, preservedConditionFalse)
 	if err != nil {
-		return false, err
+		return nil, err
 	}
 	// Step 2: remove annotations from node
-	err = c.removePreservationRelatedAnnotationsOnNode(ctx, updatedNode, removePreservationAnnotations)
+	updatedNode, err = c.removePreservationRelatedAnnotationsOnNode(ctx, updatedNode, removePreservationAnnotations)
```
r4mek, line 2452: Can we combine this and the condition update on node as a single operation?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -744,22 +744,27 @@ func (c *controller) isCreationProcessing(machine *v1alpha1.Machine) bool {
...
+	nodeAnnotated         bool
+	machineAnnotated      bool
+	nodeValue             string
+	machineValue          string
+	lastAppliedNodeValue  string
+	preserveExpiryTimeSet bool
+}
+
 // manageMachinePreservation manages machine preservation based on the preserve annotation values on the node and machine objects.
 func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1alpha1.Machine) (retry machineutils.RetryPeriod, err error) {
```
r4mek, line 760: We can split this function such that one handles the case where `nodeName` is set and other where `nodeName` is not set.

pkg/util/provider/machinecontroller/machine.go
gagan16k: Agreed, it is not necessary.

pkg/util/provider/machinecontroller/machine.go
gagan16k: `removeAnnotations` should stay as, if it is removed,  `stopPreservationIfActive` will need to actively check this.`preservationStopped` can be removed (Based on discussion)

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2376,41 +2376,41 @@ func (c *controller) preserveMachine(ctx context.Context, machine *v1alpha1.Mach
 		_, err = nodeops.AddOrUpdateConditionsOnNode(ctx, c.targetCoreClient, updatedNode.Name, *newCond)
 		if drainErr != nil {
 			klog.Errorf("error draining preserved node %q for machine %q : %v", nodeName, machine.Name, drainErr)
-			return drainErr
+			return nil, drainErr
 		}
```
r4mek, line 2380: I don't think we will reach this codepath as we are gonna skip the `if needsUpdate {` if we have `drainErr`. So this block can be shifted after `c.drainPreservedNode()` call.

Hence, we would be returning without an error even if drain was unsuccessful.

r4mek (commented):
Thanks for the PR! I have posted few comments, some of which are more related to code cleanup which you can choose to do now or defer to a later PR.

pkg/util/provider/machinecontroller/machine.go
thiyyakat: The only way to distinguish between node/machine being annotated invalidly with "", or whether the annotation is unset(valid- means deletion of annotation), is by checking nodeAnnotated/machineAnnotated.

pkg/util/provider/machinecontroller/machine.go
thiyyakat: I'm sorry. I'm not sure I understand what you mean by "allow" in this context because it is only checking if a machine needs preservation logic to be run on it. Could you please explain?

pkg/util/provider/machinecontroller/machine.go
thiyyakat: We want to check if a machine is preservation-bound at all before we return an error. This is being done so that we don't alter the flow for machines that are not preservation-bound. This is the current issue that the PR is trying to solve.

pkg/util/provider/machinecontroller/machine.go
thiyyakat: Please check the comment above (sorry, it's long). It gives an example scenario of why lastAppliedNodeValue is needed. If MCM goes down, we won't be able to tell if node was never annotated or if node annotation was deleted.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: `info.preserveExpiryTimeSet` is only being used to check if a machine is preservation-bound. We do not pass it to stopPreservationIfActive.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: We cannot unfortunately. Annotations are updated using `Update` and conditions using `UpdateStatus`.

pkg/util/provider/machinecontroller/machine.go
thiyyakat: Will raise a separate PR for refactoring. Since the purpose of this is to play it safe and quickly deliver a fix without introducing any new regressions, I would like to leave this as-is for now.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Why would we skip `if needsUpdate{` if we have `drainErr` that is not `nil`? In `computeNewNodePreservedCondition` we set `needsUpdate` to `true` if it is the first time that drain is failing.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Will raise a separate PR for any refactoring needed. Since the purpose of this is to play it safe and quickly deliver a fix without introducing any new regressions, I would like to leave this as-is for now.

pkg/util/provider/machinecontroller/machine.go
r4mek: Ignore the previous comment, I misunderstood the preservation bound.

pkg/util/provider/machinecontroller/machine.go
elankath: `getPreserveStateInfo` should not return a pointer. It is a pure value object which is always non-nil. 

PS: it's state should not be changed in the caller - it should be immutable once created for a specific Machine.

thiyyakat: Addressed in commit: [Address review comment: modify getPreserveStateInfo to return a value object.](https://github.com/gardener/machine-controller-manager/pull/1111/changes/8cb22d206e8ed29a330eb994e4562fa98098a4e2)

elankath: OK, let us address the other change - immutability in the code simplification PR which will not be part of this fix PR.

thiyyakat: There were a few other comments as well on this PR which are related to refactoring and code simplification. I will make a note of them and address them in the refactoring PR.

aaronfern (commented):
/lgtm

aaronfern (approved):

PR #1119, opened 2026-07-15
Decouple Cordoning from Draining to Avoid Uncordoning User-Cordoned Nodes in Preservation Flow

### LINKED ISSUE #1114: De-couple drain and uncordoning

/area auto-scaling
/kind enhancement
/priority 3

Currently, the machine controller performs cordoning and drain together in a single `RunDrain()`:
https://github.com/gardener/machine-controller-manager/blob/bd15a0127dce75ee533a6a917e664ffd0bab68c6/pkg/util/provider/drain/drain.go#L218-L238

Instead, the proposal is to de-couple such that cordoning can be performed before drain, and outside of `RunDrain`.

**Why is this needed**:

Drain and Cordoning are semantically different operations and must not be coupled together. Moreover, separating them out also allows cordoning to be done via taints, if required, as in the machine preservation case. 

In preservation, we uncordon the backing nodes of preserved machines when the machine transitions from Failed-->Running. We need a way to ensure a manual cordon of the node is not inadvertently undone by the preservation flow. Applying a unique taint before draining preserved machines lets this uncordoning happen safely, without overriding a user's explicit cordon.

thiyyakat:

This PR removes the cordoning activity from `RunDrain` allowing us to control the two activities separately. This also enables the cordoning during machine preservation's drain to be done without altering `spec.Unschedulable`, thereby ensuring a user's manual cordoning/uncordoning is always honoured.

The changes in the PR are:
- Remove function `RunCordonOrUncordon`
- Add a new function `cordonNode()` and call it in the callers of `RunDrain()`.
- For preservation of failed machines, use custom taint to cordon.
- Add helper function `updateMachineStatusForDrain` to update machine status during drain across all drain-xxx functions.

Fixes #1114 

IT was run with provider-AWS

The changes were tested by running mcm and mcm-provider-aws locally, targeting a shoot cluster.

Scenarios tested:
- Drain during rolling update 
- Drain during inplace update
- Drain during scale-down
- Drain when a machine transitions to Failed and is preserved
- Drain when a preserved machine transitions to Failed
- `kubectl delete mc <machine-name>` after labelling machine with force-deletion:True 
- Drain when machineDrainTimeout lapses (force drain)

**Release note**:

```other operator
De-couple cordoning and drain in the various operations that involve draining such as updates and machine preservation. 

```

### DIFF
--- a/pkg/util/provider/machinecontroller/machine_util.go
+++ b/pkg/util/provider/machinecontroller/machine_util.go
@@ -1555,7 +1555,6 @@ func (c *controller) drainNode(ctx context.Context, deleteMachineRequest *driver
 		forceDeletePods                                 bool
 		forceDeleteMachine                              bool
 		timeOutOccurred                                 bool
-		skipDrain                                       bool
 		description                                     string
 		state                                           v1alpha1.MachineState
 		readOnlyFileSystemCondition, nodeReadyCondition v1.NodeCondition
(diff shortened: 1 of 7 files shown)


pkg/util/provider/machinecontroller/machine_util.go
takoverflow: Since you've gotten rid of this `skipDrain` check elsewhere by returning early, why not do it here as well, this would allow for removal of another `if-else` at L1612. WDYT?

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1655,97 +1645,142 @@ func (c *controller) drainNode(ctx context.Context, deleteMachineRequest *driver
...
+
+// cordonNode sets node.Spec.Unschedulable to true, if not already set to true
+func (c *controller) cordonNode(ctx context.Context, nodeName string) error {
+	node, err := c.targetCoreClient.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
+	if err != nil {
+		// Deletion could be triggered when machine is just being created, no node present then
+		return nil
+	}
+	if node.Spec.Unschedulable {
+		klog.V(3).Infof("Scheduling state for node %q is already in desired state", node.Name)
```
takoverflow:
```suggestion
		klog.V(3).Infof("Node %q is already marked as Unschedulable", node.Name)
```
The log stating `desired state` without context doesn't really inform what the desired state actually is.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2443,19 +2480,29 @@ func (c *controller) stopPreservationIfActive(ctx context.Context, machine *v1al
...
 		machine, err = c.removePreserveAnnotationOnMachine(ctx, machine)
 		if err != nil {
 			return false, err
 		}
 	}
-	// Step 4: update machine status to set preserve expiry time to nil
+
+	// Step 4: remove preservation-related taint regardless of machine phase.
+	// If machine is in Running, workload can get scheduled onto it.
+	err = nodeops.RemoveTaintOffNode(ctx, c.targetCoreClient, updatedNode.Name, updatedNode, &v1.Taint{Key: machineutils.NodePreservedTaintKey, Effect: v1.TaintEffectNoSchedule})
```
takoverflow:
```suggestion
	err = nodeops.RemoveTaintOffNode(ctx, c.targetCoreClient, updatedNode.Name, updatedNode, &v1.Taint{
	Key: machineutils.NodePreservedTaintKey, 
	Effect: v1.TaintEffectNoSchedule,
	})
```

takoverflow: What happens if `c.targetCoreClient` is `nil`?

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2443,19 +2480,29 @@ func (c *controller) stopPreservationIfActive(ctx context.Context, machine *v1al
...
+	if err != nil {
+		return true, err
```
takoverflow: Why is this returning `true`?

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2591,13 +2638,17 @@ func (c *controller) drainPreservedNode(ctx context.Context, machine *v1alpha1.M
 			readOnlyFileSystemCondition = condition
 		}
 	}
-
+	if nodeName == "" {
```
takoverflow: This can be moved above the conditions check, since that is useless if the `nodeName` is `""`

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2591,13 +2638,17 @@ func (c *controller) drainPreservedNode(ctx context.Context, machine *v1alpha1.M
...
+		klog.Warningf("(drainNode) machine %q has no node name. Skipping drain.", machine.Name)
+		return nil
+	}
 	// verify and log node object's existence
 	_, err = c.nodeLister.Get(nodeName)
 	if err == nil {
 		klog.V(3).Infof("(drainNode) For node %q, machine %q, nodeReadyCondition: %s, readOnlyFileSystemCondition: %s", nodeName, machine.Name, nodeReadyCondition, readOnlyFileSystemCondition)
 	} else if apierrors.IsNotFound(err) {
 		klog.Warningf("(drainNode) Node %q for machine %q doesn't exist, so drain will finish instantly", nodeName, machine.Name)
```
takoverflow:
```suggestion
		klog.Warningf("(drainNode) Node %q for machine %q doesn't exist. Skipping drain.", nodeName, machine.Name)
```

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2637,6 +2688,14 @@ func (c *controller) drainPreservedNode(ctx context.Context, machine *v1alpha1.M
 		)
 	}
 
+	// since we do not wish to accidentally uncordon a user-cordoned node after preservation stops,
+	// we add a taint with effect NoSchedule, before draining the node, instead of setting Spec.Unschedulable = true
```
takoverflow:
```suggestion
	// we add a taint with effect 'NoSchedule' before draining the node, instead of cordoning it.
```

takoverflow (dismissed):
Thanks for the changes, would it be possible as part of this PR to combine the three separate `drain` methods? Since it's getting harder to keep track of which one does what specifically.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1655,97 +1645,142 @@ func (c *controller) drainNode(ctx context.Context, deleteMachineRequest *driver
...
+
+func (c *controller) updateMachineStatusForDrain(ctx context.Context, machine *v1alpha1.Machine, description string, state v1alpha1.MachineState, opType v1alpha1.MachineOperationType) (machineutils.RetryPeriod, error) {
+	retryPeriod, err := c.machineStatusUpdate(
 		ctx,
 		machine,
 		v1alpha1.LastOperation{
 			Description:    description,
 			State:          state,
-			Type:           v1alpha1.MachineOperationDelete,
+			Type:           opType,
```
aaronfern, line 1768: There is no need to pass `opType` to a function that is specific to updating machine statuses for drains, it should always be `v1alpha1.MachineOperationDrainNode`

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1655,97 +1645,142 @@ func (c *controller) drainNode(ctx context.Context, deleteMachineRequest *driver
...
-
-				// Return error even when machine object is updated
-			} else if err != nil && forceDeleteMachine {
-				// Drain failed on force deletion
-				klog.Warningf("Drain failed for machine %q. However, since it's a force deletion shall continue deletion of VM. \nBuf:%v \nErrBuf:%v \nErr-Message:%v", machine.Name, buf, errBuf, err)
-
+		if err = c.cordonNode(ctx, nodeName); err != nil {
+			if forceDeleteMachine {
+				// Cordoning failed on force deletion
+				klog.Warningf("Cordoning failed for machine %q. However, since it's a force deletion shall continue deletion of VM. \nErr-Message:%v", machine.Name, err)
```
aaronfern: Please also mention node details in this warning

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1655,97 +1645,142 @@ func (c *controller) drainNode(ctx context.Context, deleteMachineRequest *driver
...
 				description = fmt.Sprintf("Drain failed due to - %s. However, since it's a force deletion shall continue deletion of VM. %s", err.Error(), machineutils.DelVolumesAttachments)
 				state = v1alpha1.MachineStateProcessing
 			} else {
-				klog.Errorf("drain failed for machine %q , providerID %q ,backing node %q. \nBuf:%v \nErrBuf:%v \nErr-Message:%v", machine.Name, getProviderID(machine), getNodeName(machine), buf, errBuf, err)
-
-				description = fmt.Sprintf("Drain failed due to - %s. Will retry in next sync. %s", err.Error(), machineutils.InitiateDrain)
+				klog.Errorf("cordoning of backing node %q for machine %q, with providerID %q, failed with error: %v", nodeName, machine.Name, getProviderID(machine), err)
+				description = fmt.Sprintf("Cordoning failed due to - %s. Will retry in next sync. %s", err.Error(), machineutils.InitiateDrain)
```
aaronfern: Did you mean `Drain failed due to...`?

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1655,97 +1645,142 @@ func (c *controller) drainNode(ctx context.Context, deleteMachineRequest *driver
...
 				state = v1alpha1.MachineStateFailed
 			}
+			updateRetryPeriod, updateErr := c.updateMachineStatusForDrain(ctx, machine, description, state, v1alpha1.MachineOperationDelete)
+			if updateErr != nil {
+				return updateRetryPeriod, updateErr
+			}
+			return machineutils.ShortRetry, err
```
aaronfern, line 1674: Why are we returning unconditionally here? Iiuc if cordoning fails for a machine that is marked for force drain, this return is still applicable which is not our intention, not what the printed warning prints

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1655,97 +1645,142 @@ func (c *controller) drainNode(ctx context.Context, deleteMachineRequest *driver
...
+			state = v1alpha1.MachineStateFailed
 		}
 	}
 
-	updateRetryPeriod, updateErr := c.machineStatusUpdate(
+	updateRetryPeriod, updateErr := c.updateMachineStatusForDrain(ctx, machine, description, state, v1alpha1.MachineOperationDelete)
+	if updateErr != nil {
+		return updateRetryPeriod, updateErr
+	}
+	return machineutils.ShortRetry, fmt.Errorf("%s", description)
```
aaronfern: Is it right to always return as error? even for cases when `RunDrain` ran successfully?

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1655,97 +1645,142 @@ func (c *controller) drainNode(ctx context.Context, deleteMachineRequest *driver
...
+		if err == nil {
+			// Drain successful
+			klog.V(2).Infof("Drain successful for machine %q ,providerID %q, backing node %q. \nBuf:%v \nErrBuf:%v", machine.Name, getProviderID(machine), getNodeName(machine), buf, errBuf)
+
+			if forceDeletePods {
+				description = fmt.Sprintf("Force Drain successful. %s", machineutils.DelVolumesAttachments)
+			} else { // regular drain already waits for vol detach and attach for another node.
+				description = fmt.Sprintf("Drain successful. %s", machineutils.InitiateVMDeletion)
+			}
+			err = fmt.Errorf("%s", description)
```
aaronfern: I do not see this err value being used anywhere, and it seems odd (or rather wrong) that a "Drain successful" message is added to an error

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1655,97 +1645,142 @@ func (c *controller) drainNode(ctx context.Context, deleteMachineRequest *driver
...
+	return machineutils.ShortRetry, fmt.Errorf("%s", description)
+}
+
+// cordonNode sets node.Spec.Unschedulable to true, if not already set to true
+func (c *controller) cordonNode(ctx context.Context, nodeName string) error {
+	node, err := c.targetCoreClient.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
+	if err != nil {
+		// Deletion could be triggered when machine is just being created, no node present then
+		return nil
+	}
```
aaronfern: Use `apierrors.IsNotFound(err)` to filter out cases where node is not found. For all other errors, return the error

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2637,6 +2688,14 @@ func (c *controller) drainPreservedNode(ctx context.Context, machine *v1alpha1.M
 		)
 	}
 
+	// since we do not wish to accidentally uncordon a user-cordoned node after preservation stops,
+	// we add a taint with effect NoSchedule, before draining the node, instead of setting Spec.Unschedulable = true
+	err = nodeops.AddOrUpdateTaintOnNode(ctx, c.targetCoreClient, nodeName, &v1.Taint{Key: machineutils.NodePreservedTaintKey, Effect: v1.TaintEffectNoSchedule})
+	if err != nil {
+		klog.Errorf("cordoning of backing node %q for machine %q, with providerID %q, failed with error: %v", nodeName, machine.Name, getProviderID(machine), err)
```
aaronfern: Since this is not cordoning a node but rater tainting it
```suggestion
		klog.Errorf("tainting of backing node %q for machine %q, with providerID %q, failed with error: %v", nodeName, machine.Name, getProviderID(machine), err)
```

aaronfern:
> would it be possible as part of this PR to combine the three separate drain methods? Since it's getting harder to keep track of which one does what specifically.

This is needed, but I think it should not be done in this PR. The scope and complexity will increase considerably. Not to mention additional time to properly test it.
I can open an issue for it, so that it can be picked up separately, and I can pick this up
wdyt? @takoverflow

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: I've tried not to deviate from original behaviour. Will make this change since two of you feel this should be done 👍

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Should not be returning `true`. Will make the change.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: The `opType` was passed to cover this case: https://github.com/gardener/machine-controller-manager/blob/59134c3ae73d1da0696405e6d6b16f3068349010/pkg/util/provider/machinecontroller/machine_util.go#L1728-L1742

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: ~I made it "Cordoning failed .." because it seemed in line with the purpose of the PR – to distinguish between draining and cordoning.  However, a case can be made for logging `"Drain failed due to"` since the cordoning is still being called in drainXXX functions, and it retains previous behaviour. I am okay either way.~

Update: I see I have anyway not been consistent about this. Will make the change to `Drain failed due to..`

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: By force drain, do you mean `forceDeleteMachine` set to `true`? In that case, I thought the behaviour is identical to https://github.com/gardener/machine-controller-manager/blob/59134c3ae73d1da0696405e6d6b16f3068349010/pkg/util/provider/machinecontroller/machine_util.go#L1713-L1718

Towards the end of the function the machine status is updated. 

Could you please point out where it diverges, I am unable to figure out the difference. 😅

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: From what I could see, we always return an error. In the successful case:

https://github.com/gardener/machine-controller-manager/blob/59134c3ae73d1da0696405e6d6b16f3068349010/pkg/util/provider/machinecontroller/machine_util.go#L1709

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: This is the original behaviour, if I am not wrong: https://github.com/gardener/machine-controller-manager/blob/59134c3ae73d1da0696405e6d6b16f3068349010/pkg/util/provider/machinecontroller/machine_util.go#L1709 

 This is aligned with the other functions called in triggerDeletionFlow : https://github.com/gardener/machine-controller-manager/blob/59134c3ae73d1da0696405e6d6b16f3068349010/pkg/util/provider/machinecontroller/machine_util.go#L1856
https://github.com/gardener/machine-controller-manager/blob/59134c3ae73d1da0696405e6d6b16f3068349010/pkg/util/provider/machinecontroller/machine_util.go#L1944

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: This is the behaviour of the earlier cordoning function: https://github.com/gardener/machine-controller-manager/blob/59134c3ae73d1da0696405e6d6b16f3068349010/pkg/util/provider/drain/drain.go#L1169-L1174

I can change this if you like.

takoverflow:
> This is needed, but I think it should not be done in this PR.

Alright, fair enough. That can be tackled later.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Shall I make it `"Node %q is already cordoned"`? To maintain consistency with the remaining logs and comments?

takoverflow: Sure, that's better

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Addressed in https://github.com/gardener/machine-controller-manager/pull/1119/changes/6f42253e9e29ad3adbef22e8a7495ff80f2f6dac. The function returns early now when the node name is empty.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Addressed in https://github.com/gardener/machine-controller-manager/pull/1119/changes/96309f232a0dcfe9b5b01286a60e3d3d5301492b.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Changed return value to `false`. Addressed in https://github.com/gardener/machine-controller-manager/pull/1119/changes/b036aacb07230fa031ca030ad1ccc75569b314e2.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Addressed in https://github.com/gardener/machine-controller-manager/pull/1119/changes/b036aacb07230fa031ca030ad1ccc75569b314e2.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Addressed in https://github.com/gardener/machine-controller-manager/pull/1119/changes/96309f232a0dcfe9b5b01286a60e3d3d5301492b. I've made the changes across all drain functions.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Changed the comment. Addressed in https://github.com/gardener/machine-controller-manager/pull/1119/changes/b036aacb07230fa031ca030ad1ccc75569b314e2.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Added node name and providerID. Addressed in https://github.com/gardener/machine-controller-manager/pull/1119/changes/96309f232a0dcfe9b5b01286a60e3d3d5301492b.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Addressed in https://github.com/gardener/machine-controller-manager/pull/1119/changes/b036aacb07230fa031ca030ad1ccc75569b314e2.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Changed the log. Addressed in https://github.com/gardener/machine-controller-manager/pull/1119/changes/b036aacb07230fa031ca030ad1ccc75569b314e2.

pkg/util/provider/machinecontroller/machine_util.go
aaronfern: Okay, fair enough

pkg/util/provider/machinecontroller/machine_util.go
aaronfern: Okay, yes, I see it now. 
My apologies, this is fine

pkg/util/provider/machinecontroller/machine_util.go
aaronfern: Okay, you're right
Returning error is the status quo.

pkg/util/provider/machinecontroller/machine_util.go
aaronfern: In the original behaviour this was needed because `err` was returned. Since the final return uses `fmt.Errorf("%s", description)` now, `err` ends up not being used

pkg/util/provider/machinecontroller/machine_util.go
aaronfern: Yes, please, I would like this change to be added.
Any other error should be caught and returned

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: You're right. The assignment is useless. Will make the change 👍

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Have changed `cordonNode()` (now moved to `node.go`) to return all errors except for `NotFound` errors. Addressed in : https://github.com/gardener/machine-controller-manager/pull/1119/changes/cbdbb69cac361bdbacfbe5bcf13519e3e1319b29

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Removed the assignment in commit: https://github.com/gardener/machine-controller-manager/pull/1119/changes/cbdbb69cac361bdbacfbe5bcf13519e3e1319b29

pkg/util/provider/machinecontroller/node.go
```diff
@@ -352,6 +352,30 @@ func (c *controller) getNodePreserveAnnotationValue(nodeName string) (string, er
 	return node.Annotations[machineutils.PreserveMachineAnnotationKey], nil
 }
 
+// cordonNode sets `node.Spec.Unschedulable` to true, if not already set to true
+func (c *controller) cordonNode(ctx context.Context, nodeName string) error {
```
gagan16k, line 344: The function below  this one, `uncordonNodeIfCordoned` sets the value for `spec.Unschedulable = false`.
This could remove a manually set value, as [`manageMachinePreservation()`](https://github.com/thiyyakat/machine-controller-manager/blob/cbdbb69cac361bdbacfbe5bcf13519e3e1319b29/pkg/util/provider/machinecontroller/machine.go#L824-L826) calls this function when a machine goes from `Failed` back into `Running`.

pkg/util/provider/machinecontroller/node.go
```diff
@@ -352,6 +352,30 @@ func (c *controller) getNodePreserveAnnotationValue(nodeName string) (string, er
...
+		klog.V(3).Infof("Node %q is already cordoned.", node.Name)
+		return nil
+	}
+	clone := node.DeepCopy()
+	clone.Spec.Unschedulable = true
+	_, err = c.targetCoreClient.CoreV1().Nodes().Update(ctx, clone, metav1.UpdateOptions{})
+	if err != nil {
+		return err
+	}
+	return nil
```
gagan16k, line 365:
```suggestion
	return err
```

Nit

thiyyakat: Ha ha. Thanks! Good catch!

thiyyakat:
/hold

Putting this PR on hold until #1118 gets merged. This PR will require some preservation related cordoning-uncordoning changes, and requires the changes in #1118 to be included first.

pkg/util/provider/machinecontroller/node.go
thiyyakat: Thanks. Will make the change. I think I'll put this PR on hold until #1118 gets merged. After that will take all this into consideration.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: As discussed offline, this case is handled by the `nodename==""` at the beginning of the function.

pkg/util/provider/machinecontroller/node.go
thiyyakat: `uncordonNodeIfCordoned` has been removed and the code now removes the taint instead. After rebasing, the flow in `manageMachinePreservation()` has changed. PTAL. Ref: 117fb4be0cdc053060b8536b338e268099a8f0ab

pkg/util/provider/machinecontroller/node.go
thiyyakat: Corrected in 117fb4be0cdc053060b8536b338e268099a8f0ab.

thiyyakat:
/unhold

The PR has been rebased with the preservation flow changes in master. The manual tests mentioned in the PR description have been repeated with the latest changes.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1504,6 +1489,13 @@ func (c *controller) drainNodeForInPlace(ctx context.Context, machine *v1alpha1.
 		)
 	}
 
+	if err = c.cordonNode(ctx, nodeName); err != nil {
```
r4mek, line 1505: should we be cordoning the node here or would it be better to be a step of its own? since here it is not really de-coupled and is still part of drain.

thiyyakat: It should ideally be a step of its own outside all the drainXXX functions. If we're combining the drains into one function in the future, we can do this then. Since it may require some amount of refactoring and more thorough testing. This PR is to fix a bug, and the fix needs to be delivered quickly. 

Maybe I should make the PR title clearer. It doesn't read like it's a bug fix 😅

pkg/util/provider/machinecontroller/node.go
```diff
@@ -340,19 +340,25 @@ func addedOrRemovedEssentialTaints(oldNode, node *corev1.Node, taintKeys []strin
...
 
-func (c *controller) uncordonNodeIfCordoned(ctx context.Context, node *corev1.Node) error {
-	if !node.Spec.Unschedulable {
-		return nil
-	}
-	nodeClone := node.DeepCopy()
-	nodeClone.Spec.Unschedulable = false
-	_, err := c.targetCoreClient.CoreV1().Nodes().Update(ctx, nodeClone, metav1.UpdateOptions{})
+// cordonNode sets `node.Spec.Unschedulable` to true, if not already set to true
+func (c *controller) cordonNode(ctx context.Context, nodeName string) error {
```
gagan16k, line 344: Any reason the argument was changed(old function  `uncordonNodeIfCordoned`) from `node *corev1.Node` to `nodeName string`?
Both the callers already have the node object, which can be directly passed.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Changed the title to "Decouple Cordoning from Draining to Avoid Uncordoning User-Cordoned Nodes in Preservation Flow". Scope was not clear otherwise.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2646,6 +2681,18 @@ func (c *controller) drainPreservedNode(ctx context.Context, machine *v1alpha1.M
 			timeOutDuration,
 			maxEvictRetries,
 		)
+	} // since we do not wish to change a user's explicit cordoning of a node, for preservation, we make use of
```
gagan16k:
```suggestion
	}
	// since we do not wish to change a user's explicit cordoning of a node, for preservation, we make use of
```

Nit, this might read as a comment for the else block.

pkg/util/provider/machinecontroller/node.go
thiyyakat: > Both the callers already have the node object, which can be directly passed.

You are right. But if you look at the deleted `RunCordonOrUncordon`, the function was fetching the latest node object from the API server. I wanted to retain the behaviour so that the code doesn't introduce new Conflict errors.

gagan16k: Node fetch before operations is a common pattern across the codebase. After discussion, it was decided to not change this.

gagan16k:
/lgtm

Thanks for the work on preservation fixes @thiyyakat !

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Addressed in fa150102172354882672cd71503a7a9cbeb281b2

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2621,6 +2649,11 @@ func (c *controller) drainPreservedNode(ctx context.Context, machine *v1alpha1.M
 		nodeNotReadyDuration                        = 5 * time.Minute
 		ReadonlyFilesystem     v1.NodeConditionType = "ReadonlyFilesystem"
 	)
+	if nodeName == "" {
+		klog.Warningf("(drainNode) machine %q has no node name. Skipping drain.", machine.Name)
```
takoverflow:
```suggestion
		klog.Warningf("(drainPreservedNode) machine %q has no node name. Skipping drain.", machine.Name)
```
Keeping in-line with the other statements.

takoverflow (commented):
Just one minor nit, thanks for the changes!

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Thanks. Committed your change: https://github.com/gardener/machine-controller-manager/pull/1119/commits/6f3a032b0cf89296249d16d17ae3529a8fae7651

pkg/util/provider/machinecontroller/node.go
```diff
@@ -340,19 +340,25 @@ func addedOrRemovedEssentialTaints(oldNode, node *corev1.Node, taintKeys []strin
...
-	klog.Infof("Successfully uncordoned node %q", node.Name)
-	return nil
+	if node.Spec.Unschedulable {
+		klog.V(3).Infof("Node %q is already cordoned.", node.Name)
+		return nil
+	}
+	clone := node.DeepCopy()
+	clone.Spec.Unschedulable = true
+	_, err = c.targetCoreClient.CoreV1().Nodes().Update(ctx, clone, metav1.UpdateOptions{})
+	return err
```
aaronfern: Please add a log that logs successful cordoning before returning from this func

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1517,6 +1502,13 @@ func (c *controller) drainNodeForInPlace(ctx context.Context, machine *v1alpha1.
 		)
 	}
 
+	if err = c.cordonNode(ctx, nodeName); err != nil {
+		klog.Errorf("cordoning of backing node %q for machine %q, with providerID %q, failed with error: %v", nodeName, machine.Name, getProviderID(machine), err)
+		description = fmt.Sprintf("Drain failed due to - %s. Will retry in next sync. %s", err.Error(), machineutils.InitiateDrain)
+		state = v1alpha1.MachineStateProcessing
+		return c.updateMachineStatusAndNodeCondition(ctx, machine, description, state, err)
+	}
+
```
aaronfern, line 1511: Before this PR, cordoning was handled by `RunDrain()`. Over there if there was an error cordoning a node an error was returned which was processed by either of the cases [here](https://github.com/gardener/machine-controller-manager/blob/master/pkg/util/provider/machinecontroller/machine_util.go#L1728-L1739)

There were 2 cases (based on whether force deletion was enabled or not), and the major distinction was the state (`MachineStateProcessing` vs `MachineStateFailed`).
Now I see that state will always be set to `MachineStateProcessing`.

Could this not cause issues?

thiyyakat: From what I can see for `drainNodeForInPlace` in the existing code, the state is always set to `MachineStateProcessing`.

https://github.com/gardener/machine-controller-manager/blob/d8557e46858b0a083b33dd4ba97b209f0321b3ac/pkg/util/provider/machinecontroller/machine_util.go#L1549-L1565

Apologies if I'm overlooking something. 😅 

For `drainNode` existing code sets different states which has been retained in this PR.

pkg/util/provider/machinecontroller/node.go
thiyyakat: Added the log in d9580d334c23a49f5b5146995303a2a952d6fa50. PTAL.

pkg/util/provider/machinecontroller/machine_util.go
aaronfern: My bad, I confused `drainNode()` with `drainNodeForInPlace()` 🤦🏻 
All the more reason we need https://github.com/gardener/machine-controller-manager/issues/1128 😅

aaronfern:
Changes seem fine to me, thanks for making the changes!
/lgtm

aaronfern (approved):

Issue #1120, opened 2026-07-21
Introduce Simulation Provider for integration testing


takoverflow:
**How to categorize this issue?**

/area testing
/kind enhancement

**Issue Description**:

Currently, all MCM integration tests (IT) run only with an actual cloud provider, making it difficult to validate changes or perform stress tests with a large number of machines. Moreover, the tests don't run as part of the repository's CI/CD workflows since that requires a running kubernetes cluster with actual backing infrastructure.

**What would you like to be added**:

Having a simulation provider that could emulate CSP (cloud service provider) responses for the Machine `Driver` interface with support for injection of various constraints such as instance quota, mock delays or targeted failures.

The introduction of the simulation provider would be achieved in multiple stages:

- [ ] Proposal detailing out the requirement and the plan to achieve them. 
      https://github.com/gardener/machine-controller-manager/pull/1121
- [ ] Introduction of bare-bones provider that allows running the IT as is.
      https://github.com/gardener/machine-controller-manager/pull/1129
- [ ] Standalone `clustersim` tool to create virtual clusters with manually specified workload consisting of `MachineClass`es and `MachineDeployment`s to aid in manual testing.
- [ ] Addition of `simulation` modification hooks to improve and standardize the process of extending the testing scenarios.

**Why is this needed**:

This would allow for testing any MCM changes by having automated workflows that run the IT with the simulation provider without needing real infrastructure.

Furthermore, this would greatly increase developer productivity by reducing the friction of manually testing the changes since the CSP responses can be mocked and additional constraints be introduced as per their need.

PR #1121, opened 2026-07-21
Add proposal for simulation provider for testing

takoverflow:
This PR introduces a proposal to introduce a machine controller simulation provider to enhance developer experience by making manual testing easier and having IT run as part of MCM CI/CD pipelines without the need of real CSP infrastructure.

Fixes part of #1120

**Release note**:

```other operator

```

### DIFF
--- a/docs/proposals/simulation-provider.md
+++ b/docs/proposals/simulation-provider.md
@@ -1,0 +1,116 @@
+# Simulation Provider for Testing
+
+## Objective
+
+Currently, all Machine Controller Manager (MCM) integration tests (IT) run only with an actual cloud provider, making it difficult to validate changes or perform stress tests with a large number of machines. Moreover, the tests don't run as part of the repository's CI/CD workflows since that requires a running kubernetes cluster with actual backing infrastructure. This leads to manual effort on either the developer or the reviewer to actually run the integration tests to ensure that there aren't any regressions.
+
+This document proposes enhancing MCM with the capability to:
+- Run a simulated machine controller provider for integration testing purposes.
+- Introduce modifications to the simulated machine's lifecycle in the form of delays, errors, etc.
+- Specify custom `MachineDeployment`s (MCD) and `MachineClass`es (MCC) for the end user to perform manual testing.
+
+## Proposal
+
+In the MCM repository tree, introduce a new package (tentatively named) `provider-simulation` which implements the `Driver` interface required by each supported cloudprovider. Refer to [machine_error_codes.md#machine-provider-interface](../development/machine_error_codes.md#machine-provider-interface) for more details.
+
+### Initialisation
+
+Cluster creation and lifecycle to be managed via `e2e-framework`, it handles the initialisation and creation of the minimal control plane needed for running the simulation provider. This fetches the specified `kube-apiserver` and `etcd` binaries (if not present) and runs them as local processes.
+
+### Simulation Driver
+
+* `CreateMachine()`:
+  Creates a `node` for the specified `CreateMachineRequest`'s `Machine` and `MachineClass` as a representation for the existence of a backing VM. Returns a dummy `ProviderID`, `NodeName` and and `LastKnownState` message denoting successful instance creation. This can be modified to simulate quota (`ResourceExhausted`) errors or VM creation/node join delays etc.
+* `InitializeMachine()`:
+  Optional method that is used for network configuration for the VM. It can be skipped for the simulation provider. However, it can be used to simulate initialization errors/delays.
+* `DeleteMachine()`:
+  Since the deletion of the actual `node` is handled in `triggerDeletionFlow()`, for mocking the VM deletion from the CSP side, all that's needed is to stop tracking the machine (using `ProviderID`) as part of the quota (if defined) for the instance to which the machine belongs.
+* `GetMachineStatus()`:
+  Depending on whether any failures are to be injected for the simulation, this can either return success (denoted by returning the `ProviderID` and `NodeName`) or else `NotFound` or similar errors. For multiple VMs being returned for the queried machine, `OutOfRange` error is returned.
+* `ListMachines()`:
+  Returns a `map[providerID]machineName` for the specified `MachineClass`.
+* `GetVolumeIDs()`:
+  Skipped for the simulation provider for now. Can be implemented in the future if the testing requires it.
+
+### Simulation Modifications
+
+Machine lifecycle hooks where modifications to the `Driver` method implementation can be done for failure simulation (similar to [scheduling framework](https://kubernetes.io/docs/concepts/scheduling-eviction/scheduling-framework/)). This allows for convenient targeted manual testing and customized testing by choosing which failures to simulate and for how many machines.
+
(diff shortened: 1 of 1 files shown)


docs/proposals/simulation-provider.md
```diff
@@ -0,0 +1,116 @@
+# Simulation Provider for Testing
```
ashwani2k, line 1: Won't Emulator be a better fit. As you emulate the cloud provider here, or may be call it virtual, I think simulation means you will still call the actual provider but you just simulate the tests locally. 
or may be this is a like a fake provider.

docs/proposals/simulation-provider.md
```diff
@@ -0,0 +1,116 @@
...
+  - controller_suite_test.go
+  - controller_test.go
+  - resource_tracker.go
+```
+
+#### High level Flow
+
+1. Initialise control plane i.e. start `kube-apiserver`, `etcd` as local processes.
+2. Deploy MCM CRDs.
+3. Start `machine-controller-manager` and `machine-controller-simulation-provider` processes.
```
ashwani2k: Even if we keep it as simulation we should still name it as `machine-controller-manager-provider-simulator|virtual|fake` to be inline with other providers.

docs/proposals/simulation-provider.md
```diff
@@ -0,0 +1,116 @@
...
+4. Deploy user specified MCD and MCC (fallback to in-tree IT specific test MCD and MCC)
+5. Create dummy GNA secret (GNA secret, need to check if only passing a dummy environment variable for `GNA_SECRET_NAME` is sufficient or the secret itself needs to be created)
+6. Run IT. (Set some test related environment variables to true)
+
+## Target Scope
+
+Only offer functionality needed to run IT standalone (as part of CI/CD pipelines), with optional support for adding points of failures. The simulation config skaffold and supported modifications doesn't have to be exhaustive in the first iteration, however it should be extensible enough to allow someone to add their customizations to the simulation provider at their specified modification point to test edge cases (restricted still to modification related to driver interface calls).
```
ashwani2k, line 147: As currently we don't run IT standalone, it will be good to classify test added as part of this endeavor as mandatory vs non-mandatory for PR merge. Otherwise you might end up with IT test failing and never succeeding esp. if they validate scenarios which have time constraints.

ashwani2k (commented):
Thanks for collating the thoughts here. Just some observations.

docs/proposals/simulation-provider.md
elankath: @ashwani2k  It is exactly the other way around. 

> An emulator replicates both the hardware and software of a target device, acting as an exact clone (e.g., running an Android game on your PC). A simulator only mimics the software environment and its behavior, without replicating the underlying hardware (e.g., testing how an iOS app interface responds on your `Mac)

So, it is best to call it `simulated-provider`.

docs/proposals/simulation-provider.md
```diff
@@ -0,0 +1,116 @@
...
+- Introduce modifications to the simulated machine's lifecycle in the form of delays, errors, etc.
+- Specify custom `MachineDeployment`s (MCD) and `MachineClass`es (MCC) for the end user to perform manual testing.
+
+## Proposal
+
+In the MCM repository tree, introduce a new package (tentatively named) `provider-simulation` which implements the `Driver` interface required by each supported cloudprovider. Refer to [machine_error_codes.md#machine-provider-interface](../development/machine_error_codes.md#machine-provider-interface) for more details.
+
+### Initialisation
+
+Cluster creation and lifecycle to be managed via `e2e-framework`, it handles the initialisation and creation of the minimal control plane needed for running the simulation provider. This fetches the specified `kube-apiserver` and `etcd` binaries (if not present) and runs them as local processes.
```
elankath, line 18: What is this `e2e-framework` ? Why can't we just use a `make target` for this ? `Ex: make test-integration-setup` ?

docs/proposals/simulation-provider.md
```diff
@@ -0,0 +1,116 @@
...
+
+```json
+{
+  "preCreate": {
+    "minDelay": "10s",
+    "maxDelay": "30s",
+    "percentageOfMachines": "10%"
+  },
+  "preDelete": {
+    "rateLimitError": {
```
elankath, line 59: We need this configurable as a pattern for against MachineDeployments. ie we need a `machineDeploymentPattern` attribute somewhere in this json.

docs/proposals/simulation-provider.md
```diff
@@ -0,0 +1,116 @@
...
+  }
+}
+```
+
+### Integration with MCM
+
+The simulation provider would be added as a new package located at `pkg/simulator`.
+Proposed file structure:
+```
+pkg/simulator:
```
elankath: This should be `pkg/simulatedprovider`

docs/proposals/simulation-provider.md
```diff
@@ -0,0 +1,116 @@
...
+  - main.go
+  - Makefile
+  provider/
+  - driver.go
+  provider/simulation/
+  - simulation.go
+  - config.go
+  provider/simulation/injections/
+  - pre_create.go
```
elankath: I don't think files like `pre_create.go` are needed. What would even go inside this file ?  If someone wants to inject something outside of the MCM code base, then explicitly define an interface for this.

docs/proposals/simulation-provider.md
takoverflow: Clarified with [381bde0](https://github.com/gardener/machine-controller-manager/pull/1121/commits/381bde0f2205e183b68610569a2e7e7c9b823fa9) PTAL

docs/proposals/simulation-provider.md
takoverflow: As discussed offline, [kubernetes-sigs/e2e-framework](https://github.com/kubernetes-sigs/e2e-framework) is one of the recommended ways to create and consume environments for e2e k8s testing.

As part of the IT integration, its convenient helpers would be utilized for automated setup alongwith `make` targets for manual setup.

docs/proposals/simulation-provider.md
```diff
@@ -0,0 +1,155 @@
...
+     ```
+   - Run `clustersim setup <demo-cluster> --mcc=./mcc.yaml --mcd=./mcd.yaml` with the fetched data; this would create a directory `./gen/<demo-cluster>/` containing the following files
+     ```
+     ./gen/<demo-cluster>
+     + start-config-mcm.json # Contains the launch flags for mcm
+     + start-config-mc.json # Contains the launch flags for mc-provider-simulation
+     + mcc.yaml
+     + mcd.yaml
+     ```
+   - Additionally the `setup` would also build the required binaries for `mcm` and `mc-provider-simulation` (in `gen/bin` directories, using their corresponding `make build` targets) which can later be used when running the testing environment.
```
thiyyakat, line 126:
```suggestion
   - Additionally the `setup` would also build the required binaries for `mcm` and `mcm-provider-simulator` (in `gen/bin` directories, using their corresponding `make build` targets) which can later be used when running the testing environment.
```
Is this what you meant to write?

thiyyakat (commented):
Thanks for creating the proposal! Just one nit.

Issue #1123, opened 2026-07-27
Introduce additional integration tests


aaronfern:
**How to categorize this issue?**

/area testing
/kind test
/priority 3

**What would you like to be added**:
The integration tests that are currently present in the repo have not been updated in some time and no new tests have been added to verify and validate newer mcm features. 
This issue is to keep track of new integration tests that are needed.

The following integration tests are needed

- [ ] Auto machine preservation -> https://github.com/gardener/machine-controller-manager/pull/1124
  The following cases need to be validated
  - Machine should be preserved on failure and should be able to re-join the cluster when it recovers
  - Threshold should be respected, and failed machines crossing this should not be preserved
  - machinePreserveTimeout should be honoured for failed machines
  - Machine should not be preserved when the `preserve=false` annotation is present
  - Preserved machine should stop being preserved when the `preserve=false` annotation is added to it
- [ ] Manual machine preservation -> https://github.com/gardener/machine-controller-manager/pull/1124
  The following cases need to be validated
  - Machine should be preserved when the `preserve=when-failed` annotation is present on it
  - Machine marked for preservation using `when-failed` should be preserved when failed even if `autoPreserveMax` is crossed
  - Failed machine that has been preserved using `when-failed` should move back to `Running` phase when it recovers
  - Preserved failed machine should terminate if preserve annotation is removed
- [ ] Auto inPlace machine updates -> https://github.com/gardener/machine-controller-manager/pull/1144
- [ ] Manual inPlace machine updates -> https://github.com/gardener/machine-controller-manager/pull/1144
- [ ] Machine is deleted when `deletionTimestamp` added to node object -> https://github.com/gardener/machine-controller-manager/pull/1126

In addition, there will be an additional task. This will be re-looked after all tests mentioned above have been added
- [ ] Modernise existing integration tests to have them all use uniform helper methods and structure as the newly added tests

**Why is this needed**:
This is needed as we have not added integration tests for newly added features, and as the codebase gets more complicated, we need a way to ensure that existing features do not break

**Note**
Cases that have to be covered for each tests will be added/updated as they are being worked on

PR #1124, opened 2026-07-27
Add integration tests for machine preservation

aaronfern:
**What this PR does / why we need it**:
This PR adds integration tests for machine preservation.

The following cases have been covered:

Auto Machine Preservation:
- [x] Ensure failed machine is preserved and rejoins the cluster on recovery
- [x] Failed machines that cross the max threshold are not preserved
- [x] machinePreserveTimeout is honoured for failing machines
- [x] Machine should not be preserved when the `preserve=false` annotation is present
- [x] Preserved machine should stop being preserved when `preserve=false` annotation is added
- [x] When `AutoPreserveFailedMachineMax` is reduced, failed machines that exceed the new `AutoPreserveFailedMachineMax ` count are moved to terminating

Manual Machine Preservation:
- [x] Machine in an mcd without any preservation fields should be preserved when failed if it has the `when-failed` annotation and move back to `Running` phase when it recovers
- [x]  Machine should be preserved when it's corresponding node is annotated with `when-failed`
- [x] A machine marked for preservation using the `when-failed` annotation should be preserved when failed even if `autoPreserveMax` is crossed
- [x] Preserved machine should terminate if the preservation annotation is removed 

**Which issue(s) this PR fixes**:
Fixes partially #1123

**Special notes for your reviewer**:
To trigger machine failures, this PR takes the approach of applying a validating admission policy to block all kubelet updates. This causes kubelet not to be able to update it's node lease or it's node status and results in the node going into the `Unknown` state.
This is a k8s native way of achieving node failures and has no dependency on the provider or any other node/worker-pool setting

This might not work with the virtual provider of the virtual provider does not use kubelets, but imo this can be considered a special case and can be tackled in a separate PR

**Release note**:

```other developer
Added integration tests for machine preservation
```

PR #1125, opened 2026-07-27
Add integration tests for manual machine preservation

aaronfern:
This PR adds integration tests for manual machine preservation.

The following cases have been covered:
- [x] Machine in an mcd without any preservation fields should be preserved when failed if it has the `when-failed` annotation
- [x] Failed machine that has been preserved using `when-failed` should move back to `Running` phase when it recovers
- [x] A machine marked for preservation using the `when-failed` annotation should be preserved when failed even if `autoPreserveMax` is crossed
- [x] Preserved machine should terminate if the preservation annotation is removed 

Fixes partially #1123 

To trigger machine failures, this PR takes the approach of applying a validating admission policy to block all kubelet updates. This causes kubelet not to be able to update it's node lease or it's node status and results in the node going into the Unknown state.
This is a k8s native way of achieving node failures and has no dependency on the provider or any other node/worker-pool setting

This might not work with the virtual provider of the virtual provider does not use kubelets, but imo this can be considered a special case and can be tackled in a separate PR

**Release note**:

```other developer
Added integration tests for manual machine preservation
```

### DIFF
--- a/pkg/test/integration/common/framework.go
+++ b/pkg/test/integration/common/framework.go
@@ -26,6 +26,7 @@ import (
 	"github.com/onsi/gomega"
 	"github.com/onsi/gomega/gexec"
 	appsV1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/watch"
@@ -34,6 +35,7 @@ import (
 	"github.com/gardener/machine-controller-manager/pkg/apis/machine/v1alpha1"
 	"github.com/gardener/machine-controller-manager/pkg/test/integration/common/helpers"
 	"github.com/gardener/machine-controller-manager/pkg/test/utils/matchers"
+	mc_utils "github.com/gardener/machine-controller-manager/pkg/util/provider/machineutils"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


pkg/test/integration/common/framework.go
```diff
@@ -981,6 +983,272 @@ func (c *IntegrationTestFramework) ControllerTests() {
...
+				gomega.Eventually(
+					c.ControlCluster.IsMachineRunning,
+					c.timeout,
+					c.pollingInterval).
+					WithArguments(ctx, []string{machineList.Items[0].Name}).
+					Should(gomega.BeTrue())
+			})
+		})
+		ginkgo.Context("Ensure that machine annotated with `when-failed` are preserved", func() {
+			ginkgo.It("Should ensure that a machine marked for preservation using `when-failed` is preserved even if `autoPreserveMax` number of machines already preserved", func() {
```
thiyyakat, line 1066:
```suggestion
			ginkgo.It("Should ensure that a machine marked for preservation using `when-failed` is preserved even if `autoPreserveFailedMachineMax` number of machines are already preserved", func() {
```

pkg/test/integration/common/framework.go
```diff
@@ -981,6 +983,272 @@ func (c *IntegrationTestFramework) ControllerTests() {
...
+					Should(gomega.BeTrue())
+
+				// Fetch machine list again to ensure we have the latest state of the machine
+				machineList = c.ControlCluster.GetMachineList(ctx)
+
+				// Simulate kubelet failure for one node so that it can be auto preserved
+				ginkgo.By("deploy VAP and VAPB to simulate kubelet failure for one node")
+				gomega.Expect(c.TargetCluster.CreateVAPToBlockKubeletUpdates(ctx, []string{machineList.Items[0].ObjectMeta.Labels["node"]})).To(gomega.BeNil())
+
+				ginkgo.By("Waiting for machine to fail and be preserved")
```
thiyyakat, line 1114:
```suggestion
				ginkgo.By("Waiting for machine to fail and be auto-preserved")
```

pkg/test/integration/common/framework.go
```diff
@@ -981,6 +983,272 @@ func (c *IntegrationTestFramework) ControllerTests() {
...
+						},
+					},
+				}
+				patchBytes, err := json.Marshal(patch)
+				gomega.Expect(err).To(gomega.BeNil())
+				_, err = c.ControlCluster.McmClient.MachineV1alpha1().Machines(controlClusterNamespace).Patch(ctx, machineList.Items[1].Name, types.MergePatchType, patchBytes, metav1.PatchOptions{})
+				gomega.Expect(err).To(gomega.BeNil())
+
+				// Simulate kubelet failure for the second node.
+				// The first node is already auto preserved and counts AutoPreserveFailedMachineMax.
```
thiyyakat, line 1137:
```suggestion
				// The first node is already auto-preserved and counts towards AutoPreserveFailedMachineMax.
```

pkg/test/integration/common/framework.go
thiyyakat: Another useful test would be to make sure manual preservation can still be performed when AutoPreserveFailedMachineMax is zero. Since this will probably be the most common use case. You could either modify this test case itself, or add another.

thiyyakat (commented):
Thanks for adding the tests!

 There are 2 more tests that I could think of. You can take a call based on the trade-off between value and increase in IT time. First test would be to check if `when-failed` annotation works even when `AutoPreserveFailedMachineMax` is 0 or nil. Second would be to check if annotating the node object also works (since none of the tests are checking that flow at the moment.

thiyyakat:
Is it possible to configure `MachineHealthTimeout` to a lower value? We're not checking if the value is being honoured in the tests anyway. I'm asking because the values across tests will add up and may slow down IT. 


Since the shoot reconciliation depends on the machine deployment availability, can you modify one of the test cases to check for that as well? It need not be a separate test. With preserved failed machines (and no other failures), `MachineDeploymentAvailable` ConditionStatus must be set to True.

aaronfern:
Closing this PR as tests from this PR moved to #1124

PR #1126, opened 2026-07-27
Add an integration test for node deletion

aaronfern:
This PR add an integration test to verify if machine deletion happens when node deletion is triggered

Fixes partially #1123 

**Release note**:

```other developer
Added an integration test to validate machine deletion when it's corresponding node is deleted
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -43,6 +43,7 @@ require (
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/josharian/intern v1.0.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
+	github.com/kylelemons/godebug v1.1.0 // indirect
 	github.com/mailru/easyjson v0.7.7 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.3-0.20250322232337-35a7c28c31ee // indirect
--- a/go.mod
+++ b/go.mod
@@ -43,7 +43,6 @@ require (
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/josharian/intern v1.0.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
-	github.com/kylelemons/godebug v1.1.0 // indirect
 	github.com/mailru/easyjson v0.7.7 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.3-0.20250322232337-35a7c28c31ee // indirect
--- a/pkg/test/integration/common/framework.go
+++ b/pkg/test/integration/common/framework.go
@@ -768,23 +768,27 @@ func (c *IntegrationTestFramework) ControllerTests() {
 
 				ginkgo.By("Waiting until number of ready nodes is 1 more than initial nodes")
 				gomega.Eventually(
-					c.TargetCluster.GetNumberOfNodes,
-					c.timeout,
-					c.pollingInterval).
-					Should(gomega.BeNumerically("==", initialNodes+1))
-
-				gomega.Eventually(
 					c.TargetCluster.GetNumberOfReadyNodes,
 					c.timeout,
 					c.pollingInterval).
 					Should(gomega.BeNumerically("==", initialNodes+1))
+
+				ginkgo.By("wait till node label is present on the machine")
+				gomega.Eventually(func() bool {
+					mc, err := c.ControlCluster.McmClient.MachineV1alpha1().Machines(controlClusterNamespace).Get(ctx, testMachineName, metav1.GetOptions{})
+					if err != nil {
+						return false
+					}
+					_, exists := mc.GetLabels()[v1alpha1.NodeLabelKey]
+					return exists
+				}).Should(gomega.BeTrue())
 
 				// Get existing machine. This is needed to fetch node name
 				existingMachine, err := c.ControlCluster.McmClient.MachineV1alpha1().Machines(controlClusterNamespace).Get(ctx, testMachineName, metav1.GetOptions{})
 				gomega.Expect(err).To(gomega.BeNil())
 
 				ginkgo.By("Deleting node associated with test-machine")
-				err = c.TargetCluster.Clientset.CoreV1().Nodes().Delete(ctx, existingMachine.Labels["node"], metav1.DeleteOptions{})
+				err = c.TargetCluster.Clientset.CoreV1().Nodes().Delete(ctx, existingMachine.Labels[v1alpha1.NodeLabelKey], metav1.DeleteOptions{})
 				gomega.Expect(err).To(gomega.BeNil())
 
 				ginkgo.By("Waiting until number of ready nodes is equal to number of initial nodes")
--- a/pkg/test/integration/common/framework.go
+++ b/pkg/test/integration/common/framework.go
@@ -761,6 +761,47 @@ func (c *IntegrationTestFramework) ControllerTests() {
 				})
 			})
 		})
+		ginkgo.Context("node deletion", func() {
+			ginkgo.It("should delete machine resource when node resource is deleted", func() {
+				ginkgo.By("Create machine")
+				gomega.Expect(c.ControlCluster.CreateMachine(controlClusterNamespace, gnaSecretNameLabelValue)).To(gomega.BeNil())
+
+				ginkgo.By("Waiting until number of ready nodes is 1 more than initial nodes")
+				gomega.Eventually(
+					c.TargetCluster.GetNumberOfNodes,
+					c.timeout,
+					c.pollingInterval).
+					Should(gomega.BeNumerically("==", initialNodes+1))
+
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -43,6 +43,7 @@ require (
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/josharian/intern v1.0.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
+	github.com/kylelemons/godebug v1.1.0 // indirect
```
thiyyakat: This dependency gets removed on running `go mod tidy` like in the other PR.

pkg/test/integration/common/framework.go
```diff
@@ -735,6 +735,47 @@ func (c *IntegrationTestFramework) ControllerTests() {
...
+					c.TargetCluster.GetNumberOfNodes,
+					c.timeout,
+					c.pollingInterval).
+					Should(gomega.BeNumerically("==", initialNodes+1))
+
+				gomega.Eventually(
+					c.TargetCluster.GetNumberOfReadyNodes,
+					c.timeout,
+					c.pollingInterval).
+					Should(gomega.BeNumerically("==", initialNodes+1))
```
thiyyakat, line 748: Just out of curiosity, why are there separate `Eventually` checks for `GetNumberOfNodes` and `GetNumberOfReadyNodes`? Does the latter not subsume the first check? I can see it is being done elsewhere too, just wanted to understand rationale.

pkg/test/integration/common/framework.go
```diff
@@ -735,6 +735,47 @@ func (c *IntegrationTestFramework) ControllerTests() {
...
+
+				// Get existing machine. This is needed to fetch node name
+				existingMachine, err := c.ControlCluster.McmClient.MachineV1alpha1().Machines(controlClusterNamespace).Get(ctx, testMachineName, metav1.GetOptions{})
+				gomega.Expect(err).To(gomega.BeNil())
+
+				ginkgo.By("Deleting node associated with test-machine")
+				err = c.TargetCluster.Clientset.CoreV1().Nodes().Delete(ctx, existingMachine.Labels["node"], metav1.DeleteOptions{})
```
thiyyakat: Is there a race condition possible here? existingMachine.Labels["node"] may get set after a node is ready, in which case Delete will fail and leave an orphaned resource. You could avoid this by waiting for the label to get set on the machine before calling delete.

thiyyakat:
```suggestion
				err = c.TargetCluster.Clientset.CoreV1().Nodes().Delete(ctx, existingMachine.Labels[v1alpha1.NodeLabelKey], metav1.DeleteOptions{})
```

thiyyakat (commented):
Thanks for the changes! The test looks good to me. I just had a few questions about possible behaviour.

pkg/test/integration/common/framework.go
aaronfern: You're right, it does seem unnecessary. I think this pattern was added long ago and just not updated
Addressed in https://github.com/gardener/machine-controller-manager/pull/1126/commits/dbf9765b4ea368b15f6512309e780c8a13acf314

pkg/test/integration/common/framework.go
aaronfern: I think a check here is a bit unnecessary as it's not common for node labels not to be part of the machine.
That being said, this is a fair ask and I've addressed it in https://github.com/gardener/machine-controller-manager/pull/1126/commits/dbf9765b4ea368b15f6512309e780c8a13acf314

pkg/test/integration/common/framework.go
```diff
@@ -735,6 +735,51 @@ func (c *IntegrationTestFramework) ControllerTests() {
 				})
 			})
 		})
+		ginkgo.Context("node deletion", func() {
+			ginkgo.It("should delete machine resource when node resource is deleted", func() {
+				ginkgo.By("Create machine")
+				gomega.Expect(c.ControlCluster.CreateMachine(controlClusterNamespace, gnaSecretNameLabelValue)).To(gomega.BeNil())
```
gagan16k, line 741: Just a thought, to save time, could we have this machine creation along with the creation on the other `test-machine` (L659). This way, we can save time on machine creation, and still have it be part of the core machine tests.

pkg/test/integration/common/framework.go
```diff
@@ -735,6 +735,51 @@ func (c *IntegrationTestFramework) ControllerTests() {
...
+				gomega.Eventually(func() bool {
+					mc, err := c.ControlCluster.McmClient.MachineV1alpha1().Machines(controlClusterNamespace).Get(ctx, testMachineName, metav1.GetOptions{})
+					if err != nil {
+						return false
+					}
+					_, exists := mc.GetLabels()[v1alpha1.NodeLabelKey]
+					return exists
+				}).Should(gomega.BeTrue())
+
+				// Get existing machine. This is needed to fetch node name
```
takoverflow: Is this additional fetch necessary since above when we're polling for the `NodeLabelKey` we can get the node name there itself and then issue the delete call right?

aaronfern: Updated the step above to store returned mc in a variable and reusing that
ptal

pkg/test/integration/common/framework.go
aaronfern: Okay, makes sense considering it saves time
I've done that and moves this test so that the other tests also make sense
PTAL

pkg/test/integration/common/framework.go
```diff
@@ -698,21 +698,67 @@ func (c *IntegrationTestFramework) ControllerTests() {
...
+						c.TargetCluster.GetNumberOfNodes,
+						c.timeout,
+						c.pollingInterval).
+						Should(gomega.BeNumerically("==", initialNodes))
+					gomega.Eventually(
+						c.TargetCluster.GetNumberOfReadyNodes,
+						c.timeout,
+						c.pollingInterval).
+						Should(gomega.BeNumerically("==", initialNodes))
+				})
```
takoverflow: What is this check for? I couldn't understand the point when we're checking for the same above as well before machine deletion. Is there any difference?

pkg/test/integration/common/helpers/machine_resources.go
```diff
@@ -60,6 +62,36 @@ func (c *Cluster) CreateMachine(namespace string, gnaSecretName string) error {
 			},
 			metav1.CreateOptions{},
 		)
+	if err != nil {
+		return err
+	}
+
+	_, err = c.McmClient.
```
takoverflow: Rather than repeating the entire logic, why don't we just create a slice of machine names and run a loop that creates machines with those names and returns if there's any error.

Since the only thing changing between the two machine objects is just the name.

Alternatively this can be a function argument.

aaronfern: Yup, makes sense. Made the change, but didn't pass the mc names as args since this func is called only once.

takoverflow (commented):
/lgtm

pkg/test/integration/common/helpers/machine_resources.go
```diff
@@ -26,41 +26,49 @@ const (
...
-				ObjectMeta: metav1.ObjectMeta{
-					Name:      McName,
-					Namespace: namespace,
-				},
-				Spec: v1alpha1.MachineSpec{
-					Class: v1alpha1.ClassSpec{
-						Kind: "MachineClass",
-						Name: "test-mc-v1",
+// CreateMachines creates test-machines using machineclass "test-mc"
+func (c *Cluster) CreateMachines(namespace string, gnaSecretName string) error {
```
gagan16k: Adding to @takoverflow's [comment](https://github.com/gardener/machine-controller-manager/pull/1126#discussion_r4068510885), can we make it take a slice of machine names as input instead?

This is because in #1158 we are adding a new IT spec which needs and additional machine, and this way we can have a standard way to create machines in parallel at the start. WDYT?

aaronfern: Made this change with https://github.com/gardener/machine-controller-manager/pull/1126/commits/657efab7700141f45f7a1997085f9894cf9f14ec

pkg/test/integration/common/framework.go
```diff
@@ -671,8 +671,8 @@ func (c *IntegrationTestFramework) ControllerTests() {
 	ginkgo.Describe("machine resource", func() {
 		var initialNodes int16
 		ginkgo.Context("creation", func() {
-			ginkgo.It("should not lead to any errors and add 1 more node in target cluster", func() {
-				// In case of existing deployments creating nodes when starting simulated
+			ginkgo.It("should not lead to any errors and add 2 more node in target cluster", func() {
+				// In case of existing deployments creating nodes when starting virtual
```
thiyyakat: Just one nit. /lgtm otherwise.

```suggestion
				// In case of existing deployments creating nodes when starting simulated
```

aaronfern: Updated in https://github.com/gardener/machine-controller-manager/pull/1126/commits/9b69984c7193cb306e7c556a4cfc8e1925dfa399

thiyyakat:
/lgtm

Thanks for the changes!

takoverflow (approved):
/lgtm
/approve

PR #1129, opened 2026-07-31
feat: Simulated Provider for running IT

takoverflow:
This PR introduces Simulated provider to run IT as part of MCM repo's CI/CD. This can also be used as a mock provider with a virtual (kwok) cluster for testing/simulation purposes.

Fixes part of #1120 

Recommend reviewing PR commit-wise for an easy time.

**Release note**:

```feature developer
Introduce simulation provider to run integration tests for MCM without an actual cluster
```


pkg/simulatedprovider/cmd/machine-controller/main.go
```diff
@@ -0,0 +1,45 @@
...
+
+	s := options.NewMCServer()
+	s.AddFlags(pflag.CommandLine)
+
+	flag.InitFlags()
+	logs.InitLogs()
+	defer logs.FlushLogs()
+
+	if s.TargetKubeconfig == "" {
+		_ = fmt.Errorf("--target-kubeconfig must be provided")
```
gagan16k: Should this err be printed instead?

.ci/test
```diff
@@ -47,18 +50,22 @@ function test_with_coverage() {
...
   echo ">>>>> Finished executing unit tests"
 fi
 
-# Integration tests here will always be skipped.
 if [[ "${SKIP_INTEGRATION_TESTS}" != "" ]]; then
   echo ">>>>> Skipping integration tests"
+else
+  echo ">>>>> Invoking integration tests with the simulated provider"
+	ginkgo -v --show-node-events --poll-progress-after=300s --poll-progress-interval=60s pkg/simulatedprovider/test/integration/controller
+  echo ">>>>> Finished executing intergration test"
```
gagan16k:
```suggestion
  echo ">>>>> Finished executing integration test"
```
nit

pkg/simulatedprovider/provider/driver.go
```diff
@@ -0,0 +1,190 @@
...
+// Ref: ../../util/provider/driver/driver.go
+var _ driver.Driver = &DriverImpl{}
+
+// DriverImpl is the struct that implements the MCM driver.Driver interface
+// It also maintains a map `managedNodes` of all the nodes currently managed by MCM
+type DriverImpl struct {
+	mu            sync.Mutex
+	client        *kubernetes.Clientset
+	machineClient machineclientset.Interface
+	managedNodes  sync.Map
```
gagan16k: Every reference to `managedNodes` already holds the mutex in the current state of the code, so could the `sync.Map` just be a regular `map[string]corev1.Node`?

pkg/test/integration/common/framework.go
```diff
@@ -605,14 +625,25 @@ func (c *IntegrationTestFramework) SetupBeforeSuite() {
 
 	clusterName := targetClusterName
 
+	// Because the simulated provider creates fake secrets for the MachineClasses
+	// on-the-fly, we add an artifical delay to give it time to create the secrets
+	// before their existence is checked below.
+	if isSimulatedProvider {
+		time.Sleep(2 * time.Second)
```
gagan16k: Would this be redundant, as the block with `gomega.Eventually(.....`  polls for the secrets anyways and waits for it?

takoverflow: Yes, this was done initially when I wasn't polling. It worked for local testing however this artificial delay wasn't enough for the pipelines.

Will remove this.

pkg/simulatedprovider/provider/driver.go
takoverflow: I wanted to discuss with @elankath whether it makes sense to have the `mutex` if we could have this be a `sync.Map`. Will make the relevant required changes after confirming with him.

pkg/test/integration/common/framework.go
```diff
@@ -45,36 +44,25 @@ const (
...
 	targetDir = filepath.Join("..", "..", "..", ".ci", "controllers-test", "logs")
-	// Suffix for the`kubernetes-io-cluster` tag and cluster name for the orphan resource tracker. Used as ResourceGroupName for Azure clusters
-	targetClusterName = os.Getenv("TARGET_CLUSTER_NAME")
 	// machine-controller-manager log file
 	mcmLogFile = filepath.Join(targetDir, "mcm_process.log")
-
 	// machine-controller log file
 	mcLogFile = filepath.Join(targetDir, "mc_process.log")
 
+	// Suffix for the`kubernetes-io-cluster` tag and cluster name for the orphan resource tracker. Used as ResourceGroupName for Azure clusters
```
aaronfern:
```suggestion
	// Suffix for the `kubernetes-io-cluster` tag and cluster name for the orphan resource tracker. Used as ResourceGroupName for Azure clusters
```

pkg/simulatedprovider/cmd/machine-controller/main.go
takoverflow: Yes, will change to `fmt.Fprintf` like the other errors.

pkg/simulatedprovider/test/integration/controller/test-mcc.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: machine.sapcloud.io/v1alpha1
+credentialsSecretRef:
+  name: cloudprovider
+  namespace: test-ns
+kind: MachineClass
+metadata:
+  creationTimestamp: "2026-07-04T22:36:34Z"
```
aaronfern: `creationTimestamp` should not be needed

go.mod
```diff
@@ -25,11 +25,14 @@ require (
 	sigs.k8s.io/yaml v1.6.0
 )
 
+require sigs.k8s.io/e2e-framework v0.6.0
```
aaronfern: Why is this in a separate `require` block?

.ci/test
aaronfern:
```suggestion
  echo ">>>>> Finished executing intergration tests"
```

aaronfern (commented):
Thanks for the PR!
Looks good, I only have a few nits

go.mod
takoverflow: I don't know, `go mod tidy` formatted it that way! 😅
I'll add it to the existing require block.

pkg/simulatedprovider/cluster/cluster.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	createNamespaceFunc := envfuncs.CreateNamespace(env.Namespace)
+	env.Ctx, err = createNamespaceFunc(env.Ctx, env.Cfg)
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		return
+	}
+	return nil
+}
+
+func (env *Env) deployCRDs() (err error) {
```
r4mek, line 143: You can maybe implement this function like this:

```go
func (env *Env) deployCRDs() error {
	crdsDir, err := findCRDsDir()
	if err != nil {
		return err
	}
	_, err = envfuncs.SetupCRDs(crdsDir, "*")(env.Ctx, env.Cfg)
	return err
}

func findCRDsDir() (string, error) {
	dir, err := os.Getwd()
	if err != nil {
		return "", err
	}
	for {
		crdsDir := filepath.Join(dir, "kubernetes", "crds")
		if _, err := os.Stat(crdsDir); err == nil {
			return crdsDir, nil
		}
		parent := filepath.Dir(dir)
		if parent == dir {
			return "", fmt.Errorf("kubernetes/crds not found")
		}
		dir = parent
	}
}
```

pkg/simulatedprovider/cluster/cluster.go
```diff
@@ -0,0 +1,171 @@
...
+			mcc, ok := obj.(*v1alpha1.MachineClass)
+			if !ok || mcc.CredentialsSecretRef == nil || mcc.SecretRef == nil {
+				return
+			}
+			secretMap := map[string]string{
+				mcc.CredentialsSecretRef.Name: mcc.CredentialsSecretRef.Namespace,
+				mcc.SecretRef.Name:            mcc.SecretRef.Namespace,
+			}
+
+			for name, ns := range secretMap {
```
r4mek: maybe we can call the function twice for `credentialsSecretRef` and `secretRef` rather than having a for loop.

pkg/simulatedprovider/provider/node.go
```diff
@@ -0,0 +1,135 @@
...
+	node.Spec.Taints = slices.DeleteFunc(node.Spec.Taints, func(taint corev1.Taint) bool {
+		return taint.Key == corev1.TaintNodeNotReady
+	})
+
+	updatedNode, err = d.client.CoreV1().Nodes().Update(ctx, node, metav1.UpdateOptions{})
+	if err != nil {
+		err = fmt.Errorf("cannot update node with name %q: %w", node.Name, err)
+		return
+	}
+	updatedNode.Status.Conditions = buildNodeReadyConditions(corev1.ConditionTrue)
```
r4mek: you should create a deepcopy of the node before modifying it.

r4mek (commented):
Thanks for the PR!

pkg/simulatedprovider/provider/node.go
r4mek: you can ignore this. This is not needed.

pkg/simulatedprovider/provider/node.go
```diff
@@ -0,0 +1,135 @@
...
+	node.Labels = addNodeLabels(mc, mcc)
+	node.Status = buildNodeStatus(mcc)
+	node.Spec.ProviderID = fmt.Sprintf("fake://%s:%s", mcc.NodeTemplate.Region, mc.Name)
+	return
+}
+
+func (d *DriverImpl) transitionNodeToReady(ctx context.Context, name string) (updatedNode *corev1.Node, err error) {
+	node, err := d.client.CoreV1().Nodes().Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		err = fmt.Errorf("cannot get node with name %q: %w", node.Name, err)
```
r4mek: you should use `name` instead of `node.Name`

takoverflow:
Thanks for taking a look
Have addressed your review comments, PTAL @gagan16k @aaronfern @r4mek

pkg/simulatedprovider/cluster/cluster.go
```diff
@@ -0,0 +1,171 @@
...
+
+// Env specifies the virtual cluster details/configuration.
+type Env struct {
+	Name      string
+	Namespace string
+	Ctx       context.Context
+	Cfg       *envconf.Config
+}
+
+// New returns an kwokctl created cluster's Environment.
```
thiyyakat:
```suggestion
// New returns a kwokctl-created cluster's Environment.
```

pkg/simulatedprovider/provider/node.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return err
+	}
+	for _, n := range nodeList.Items {
+		d.managedNodes.Store(n.Name, n)
+	}
+	return nil
+}
+
+func (d *DriverImpl) buildNode(mc *v1alpha1.Machine, mcc *v1alpha1.MachineClass) (node corev1.Node) {
```
thiyyakat: Is there a reason why you're returning by value instead of as a pointer? Asking because if this provider is going to be used for stress testing, this change may make a difference. The same holds for `buildNodeStatus` as well.

pkg/simulatedprovider/provider/node.go
```diff
@@ -0,0 +1,135 @@
...
+	nodeList, err := d.client.CoreV1().Nodes().List(ctx, metav1.ListOptions{
+		LabelSelector: fmt.Sprintf(
+			"%s,%s!=", machineutils.MachineLabelKey, machineutils.MachineLabelKey,
+		),
+	})
+	if err != nil {
+		return err
+	}
+	for _, n := range nodeList.Items {
+		d.managedNodes.Store(n.Name, n)
```
thiyyakat: Can the sync.Map be converted to a store node name and pointer to a node instead? This would be similar to the k8s client-go informer cache, which relies on callers not mutating the object. The suggestion is keeping in mind that the provider can be used for stress testing.

thiyyakat (commented):
Thank you for adding the simulated provider! :heart: Just a few comments/questions from me. PTAL.

pkg/simulatedprovider/provider/node.go
takoverflow: Thanks for the review, I was thinking if it even makes sense to have the map be `map[node.Name]corev1.Node` or `map[node.Name]*corev1.Node`, since the only information that's actually used from the value is just the `ProviderID`. Wouldn't it make more sense for the map to be `map[node.Name]node.Spec.ProviderID` instead? And all the keys denote nodes which are managed by MCM. WDYT?

thiyyakat: If you don't have the node object in the sync map, how are you planning to modify ListMachines to do the label check?

takoverflow: Ah right, I could construct a struct with the ID and the class name for the map value but let me think about it.

pkg/simulatedprovider/provider/node.go
takoverflow: Changed it to a pointer.

pkg/simulatedprovider/provider/node.go
takoverflow: Using a new `managedNodeInfo` struct as the map value, it consists of both `ProviderID` and `MachineClassName`.

pkg/simulatedprovider/provider/driver.go
takoverflow: I've removed the mutex.

gagan16k: Was there any reason `sync.Map` was preferred over the mutex+map?

From the [sync.Map](https://pkg.go.dev/sync#Map) docs,
```
The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, 
for better type safety and to make it easier to maintain other invariants along with the map content.

The Map type is optimized for two common use cases: 
(1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or 
(2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. 
In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate [Mutex] or [RWMutex].
```

This map is written on every `CreateMachine/DeleteMachine` and read on every `GetMachineStatus/ListMachines`, which is general read/write mix.

takoverflow: > when multiple goroutines read, write, and overwrite entries for disjoint sets of keys.

Considering that most times we'd be making calls for different machines (i.e. keys), the mutex blocks the unrelated operation regardless of whether the operation is working with the same key. For example:
1. Create machine-a (mutex acquired)
2. Same time, delete machine-b -> This is blocked now even though the two operations have nothing in common.

Because there's a single mutex for the entire driver, it can increase the probability of needless blocks, which I wished to avoid.

pkg/simulatedprovider/provider/node.go
```diff
@@ -0,0 +1,131 @@
...
```
gagan16k: Since we only read `managedNodeInfo` from the map, why not just create an object here and store it explicitly here, instead of `corev1.Node`

pkg/simulatedprovider/provider/node.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	return nil
+}
+
+func (d *DriverImpl) buildNode(mc *v1alpha1.Machine, mcc *v1alpha1.MachineClass) *corev1.Node {
+	var node corev1.Node
+	node.Name = mc.Name
+	node.Labels = addNodeLabels(mc, mcc)
+	node.Status = buildNodeStatus(mcc)
+	node.Spec.ProviderID = fmt.Sprintf("fake://%s:%s", mcc.NodeTemplate.Region, mc.Name)
```
gagan16k, line 54: Would we need a nil check for `mcc.NodeTemplate`?

pkg/simulatedprovider/provider/node.go
```diff
@@ -0,0 +1,131 @@
...
+		labelMachineClassName: mcc.Name,
+	}
+
+	maps.Copy(labels, mc.Spec.NodeTemplateSpec.Labels)
+	return labels
+}
+
+func buildNodeStatus(mcc *v1alpha1.MachineClass) corev1.NodeStatus {
+	status := corev1.NodeStatus{
+		Capacity: maps.Clone(mcc.NodeTemplate.Capacity),
```
gagan16k, line 113: `mcc.NodeTemplate` could be `nil`, which will cause a panic at L91

pkg/simulatedprovider/cmd/machine-controller/main.go
```diff
@@ -0,0 +1,45 @@
+package main
```
gagan16k, line 5:
```suggestion
// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
//
// SPDX-License-Identifier: Apache-2.0

package main
```

Do we need to have this on this file?

pkg/simulatedprovider/provider/driver.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+		klog.Infof("Created node %q", node.Name)
+		// The node is considered 'managed' even when it hasn't transitioned to 'Ready'.
+		d.managedNodes.Store(node.Name, managedNodeInfo{
+			ProviderID:       node.Spec.ProviderID,
+			MachineClassName: req.MachineClass.Name,
+		})
+	}
+
+	node, err = d.client.CoreV1().Nodes().Get(ctx, req.Machine.Name, metav1.GetOptions{})
```
gagan16k: Can we not use the object returned by the `Create()` call instead of a fresh fetch using `Get()`?

pkg/simulatedprovider/provider/driver.go
gagan16k: Makes sense

pkg/simulatedprovider/provider/driver.go
takoverflow: What if the object already exists and we didnt create it, i.e. `found` was true, then you have to fetch the object.

Additionally, upon creation, the object is updated with some additional data (its status is updated by api server), which the built object might not have, so its best to fetch the latest state before updating it.

pkg/simulatedprovider/provider/node.go
takoverflow: Ah, this was a miss from me. Good catch, will fix it.

pkg/simulatedprovider/provider/node.go
takoverflow: NodeTemplate being `nil` means that there's no instance capacity information, no regional information. If it's `nil`, I'd prefer things to panic rather than creating nodes/machines having no data! Hope it makes sense :)

pkg/simulatedprovider/provider/node.go
takoverflow: Refer https://github.com/gardener/machine-controller-manager/pull/1129#discussion_r3835325323

pkg/simulatedprovider/provider/node.go
takoverflow: Incidentally, because of this panic only I found out that when creating v2 machineclass in IT framework, we weren't copying the `NodeTemplate`
Ref: https://github.com/gardener/machine-controller-manager/pull/1129/changes#diff-b1d981b90e7fa48758d064696fdc7f3c299500667ba01f2e48debcbbea2d09a2R472

gagan16k: Makes sense, but perhaps we could add a check with logging and graceful return instead of letting it panic?

takoverflow: Added.

gagan16k:
/lgtm

Thanks for the work on this topic!

thiyyakat:
@takoverflow , I was looking at the MCM provider code for AWS. The driver methods have some basic logs which may help with debugging. 
Examples:

https://github.com/gardener/machine-controller-manager-provider-aws/blob/f9c19c5bddc4a6f370d86cdaa93aa56e9a03aae8/pkg/aws/core.go#L85

https://github.com/gardener/machine-controller-manager-provider-aws/blob/f9c19c5bddc4a6f370d86cdaa93aa56e9a03aae8/pkg/aws/core.go#L434

Do you think that will be useful here, given that we are planning to use the simulated provider for reproducing issues and debugging as well? If you feel there are sufficient logs already in the controller, you needn't add any more.

takoverflow:
> Do you think that will be useful here, given that we are planning to use the simulated provider for reproducing issues and debugging as well? If you feel there are sufficient logs already in the controller, you needn't add any more.

@thiyyakat These logs have some use because it can help trace what happened in the driver call from the time the request was made till it finished. In the simulated provider, we don't have any real activity that's logged/takes time there wouldn't be much value I feel for these, since it can be interpreted from the machine controller logs when these calls are made.

But if you think there's use-case still, I don't have strong feelings, I can add the logs indicating what calls are being made.

thiyyakat:
> In the simulated provider, we don't have any real activity that's logged/takes time there wouldn't be much value I feel for these, since it can be interpreted from the machine controller logs when these calls are made.

Got it. No need to add the logs then.

pkg/simulatedprovider/provider/node.go
```diff
@@ -0,0 +1,134 @@
...
+			"%s,%s!=", machineutils.MachineLabelKey, machineutils.MachineLabelKey,
+		),
+	})
+	if err != nil {
+		return err
+	}
+	for _, n := range nodeList.Items {
+		d.managedNodes.Store(n.Name, managedNodeInfo{
+			ProviderID:       n.Spec.ProviderID,
+			MachineClassName: n.Name,
```
thiyyakat: Probably a stupid question, but why are you assigning the node's name as the machineClassName?

takoverflow: Not a stupid question, rather a stupid mistake.
Thanks for catching it.

takoverflow: There's nothing in the IT that actually checks if `ListMachines` is working as expected.

thiyyakat:
Thanks for the changes and for addressing my comments!

/lgtm

docs/development/integration_tests.md
```diff
@@ -21,7 +21,17 @@ Integration tests for `machine-controller-manager-provider-{provider-name}` can
...
  - If you are deploying the secret manually, a `Secret` named `test-mc-secret` (that contains the provider secret and cloud-config) in the `default` namespace of the Control Cluster should be created.
-3. The controllers log files (`mcm_process.log` and `mc_process.log`) are stored in `.ci/controllers-test/logs` repo and can be used later.
+3. The controllers log files (`mcm_process.log` and `mc_process.log`) are stored in `.ci/controllers-test/logs` directory and can be used later.
+
+### With the Simulated Provider
+
+To run the integration tests against the simulated provider, rather than running the `test-integration` target from the provider repository, just run it on the project root.
+That is, `machine-controller-manager/Makefile` provides the same helper target that runs IT with the simulated provider.
+```
+make test-integration
```
elankath, line 33: This currently fails with panics. Maybe you need to move this section up and also add a pre-requisites section.

pkg/simulatedprovider/provider/driver.go
```diff
@@ -0,0 +1,198 @@
...
+	defer driverAPIMetricRecorderFn(labelInitializeMachine, &err)()
+	err = status.Error(
+		codes.Unimplemented, "simulation provider does not implement InitializeMachine",
+	)
+	return
+}
+
+// DeleteMachine handles a machine deletion request. For the simulated provider, it just
+// removes the corresponding node from the managedNodes map.
+func (d *DriverImpl) DeleteMachine(_ context.Context, req *driver.DeleteMachineRequest) (resp *driver.DeleteMachineResponse, err error) {
```
elankath, line 135: Shouldn't we call `d.client.CoreV1().Nodes().Delete(ctx, req.Machine.Name, metav1.DeleteOptions{})`

takoverflow: Node deletion is handled by the machine deletion flow itself. As part of `Driver` interface `DeleteMachine` call, we don't actually delete the node object.

pkg/simulatedprovider/cluster/cluster.go
```diff
@@ -0,0 +1,197 @@
...
+	env.Ctx, err = createNamespaceFunc(env.Ctx, env.Cfg)
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		return
+	}
+	return nil
+}
+
+func (env *Env) deployCRDs() (err error) {
+	// TODO: think if there's a better caller-agnostic way to get this path
+	_, f, _, ok := runtime.Caller(0)
```
elankath: Even I agree this isn't great as it will break in a lot of cases: cross-compilation, when binary is built with trimpath, etc. Just embed the CRD yamls like you have already done for `kwok-config.yaml`

pkg/simulatedprovider/cluster/cluster.go
```diff
@@ -0,0 +1,197 @@
...
+	exportLogsFunc := envfuncs.ExportClusterLogs(env.Name, dir)
+	_, err = exportLogsFunc(env.Ctx, env.Cfg)
+	return
+}
+
+func (env *Env) createCluster() (err error) {
+	// Using the direct path doesn't work since it looks for the file
+	// relative to the caller, so the config file is embedded and a
+	// temporary file path is passed in order to create the cluster.
+	configPath := filepath.Join(os.TempDir(), "kwok-config.yaml")
```
elankath: Why not `f, err := os.CreateTemp("", "kwokctl-config-*")` to avoid future conflicts ?

elankath: It might be better to create a cluster sub-dir inside temp that can be deleted. I leave the choice upto you. Basically changes of 2 `kwow-config`'s are rather high.

pkg/simulatedprovider/provider/driver.go
```diff
@@ -0,0 +1,198 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package driver
```
elankath: package name does not match folder name.

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	k8s.io/klog/v2 v2.130.1
 	k8s.io/kube-openapi v0.0.0-20250710124328-f3f2b991d03b // keep this value in sync with k8s.io/apiserver
 	k8s.io/utils v0.0.0-20250604170112-4c0f3b243397
+	sigs.k8s.io/e2e-framework v0.6.0
```
elankath: Brings in heavy transitive dependencies in test only code, including `controller-runtime`, `websocket`,  etc to a pure `client-go` project which is ginkgo based. Should use separate go module under `pkg/simulatedprovider`.  Should be straightforward since it is fairly isolated.

pkg/simulatedprovider/cluster/cluster.go
elankath: EDIT: never mind. this is not possible using `go:embed` unlike Rust which supports this. I propose using the karpenter style mechanism and introduce a `crds.go` in `kubernetes/crds` that embeds the CRD's, which is then consumable in the simulated provider which can avoid this hack. Let us see if this works out.

pkg/simulatedprovider/provider/driver.go
```diff
@@ -0,0 +1,198 @@
...
+	defer driverAPIMetricRecorderFn(labelCreateMachine, &err)()
+
+	var node *corev1.Node
+	if _, found := d.managedNodes.Load(req.Machine.Name); !found {
+		if req.MachineClass.NodeTemplate == nil {
+			err = fmt.Errorf("cannot build a node as `machineClass.NodeTemplate` is nil")
+			return
+		}
+		node = d.buildNode(req.Machine, req.MachineClass)
+		_, err = d.client.CoreV1().Nodes().Create(ctx, node, metav1.CreateOptions{})
```
elankath, line 99: VERY MINOR: You can save the node object returned and use it instead of an additional call to get.

pkg/simulatedprovider/cluster/cluster.go
```diff
@@ -0,0 +1,197 @@
...
+	if err = env.createCluster(); err != nil {
+		return
+	}
+
+	if err = env.deployCRDs(); err != nil {
+		return
+	}
+	// This delay is added to give time to the CRDs to register
+	// before any MCM watches are started or objects are deployed.
+	time.Sleep(200 * time.Millisecond)
```
elankath: can replace with `wait.PollUntilContextTimeout`

pkg/simulatedprovider/provider/driver.go
takoverflow: But what if the node object is updated before we attempt to change it's status. Wouldn't it cause a conflictError if we then transition it to a `Ready` node.

pkg/simulatedprovider/cluster/cluster.go
takoverflow: I do clean up the configFile once the cluster is created, also as discussed offline, this file is just used for bootstrapping the cluster. But just to be on the safe side, I've made the requested change.

takoverflow: Same as https://github.com/gardener/machine-controller-manager/pull/1129#discussion_r3920764889

docs/development/integration_tests.md
aaronfern: +1
I had a panic too. Some prereqs are needed

takoverflow: Added a note stating that existing old kwokctl versions can cause these panics when fetching the control plane components' binaries. Now when creating a cluster, a new version is explicitly passed which ensures that this isn't observed.

elankath (approved):
thanks for this valuable feature. Looking forward to using it and the PR coming in next for daily dev

PR #1130, opened 2026-08-03
Make drain package reusable for controller-runtime based callers

### LINKED ISSUE #1133: Make drain package reusable for controller-runtime based callers

/kind enhancement
/priority 3

The drain package in machine-controller-manager is currently tightly coupled to the client-go informer/lister stack. Callers like Gardener that use a controller-runtime client (and do not run client-go informers) cannot reuse the drain logic directly.

  Extend the drain Options API to support external callers that:
  - Do not run a client-go informer/lister for pods
  - Need to plug in their own pod source
  - Want to customize pod selection with additional filters
  - Want to skip volume detach/reattach handling independently of ForceDeletePods

**Why is this needed**:

This is required to streamline the drain logic across gardener.

acumino:
/kind enhancement

This PR extends the drain Options API to support callers (e.g. Gardener) that do not run client-go informers and need to plug in their own pod source or customize pod selection.

- Add `PodProvider` interface (PodsForNode) and SetPodProvider setter — allows callers to supply pods via a controller-runtime client instead of a pod lister/informer.
- Add `AdditionalPodFilters` field — caller-supplied predicates evaluated alongside the built-in drain filters.
- Add `SkipVolumeDetach` flag — skips PV detach during eviction, independent of ForceDeletePods.
- Make `podSynced` cache-sync conditional — skipped when no informer is running (podSynced == nil).
- Make `pdbLister` nil-safe so callers that don't set one don't panic.

Fixes https://github.com/gardener/machine-controller-manager/issues/1133

**Release note**:

```feature developer
The drain package now exposes a `PodProvider` interface, `AdditionalPodFilters`, `SkipVolumeHandling`, and a nil-safe `pdbLister` guard, allowing external controllers (e.g. Gardener) to reuse the drain logic without running a client-go informer stack.
```

### DIFF
--- a/pkg/util/provider/drain/drain.go
+++ b/pkg/util/provider/drain/drain.go
@@ -77,6 +77,25 @@ type Options struct {
 	volumeAttachmentHandler      *VolumeAttachmentHandler
 	Timeout                      time.Duration
 	podSynced                    cache.InformerSynced
+
+	// AdditionalPodFilters are extra predicates evaluated alongside the built-in filters when selecting pods to
+	// drain.
+	AdditionalPodFilters []AdditionalPodFilter
+	// podProvider, if set, is used to list the pods on the node instead of podLister. It allows callers that do not
+	// run client-go informers (e.g. controller-runtime based controllers) to supply their own pod source.
+	podProvider PodProvider
+	// SkipVolumeHandling, if true, skips the PersistentVolume detach/reattach handling during eviction.
+	SkipVolumeHandling bool
+}
+
+// AdditionalPodFilter takes a pod and returns whether the pod should be included for draining (true) or excluded
+// (false). It is a caller-supplied predicate evaluated in addition to the built-in filters.
+type AdditionalPodFilter func(corev1.Pod) bool
+
+// PodProvider lists the pods running on a given node. It abstracts the pod source (client-go lister vs. a
+// controller-runtime client) used during drain.
+type PodProvider interface {
+	PodsForNode(ctx context.Context, nodeName string) ([]corev1.Pod, error)
 }
 
 // Takes a pod and returns a bool indicating whether or not to operate on the
@@ -214,6 +233,13 @@ func NewDrainOptions(
 	}
 }
 
+// SetPodProvider sets the PodProvider used to list the pods on the node instead of the client-go pod lister. It lets
+// callers that do not run client-go informers (e.g. controller-runtime based controllers) supply their own pod
+// source when using NewDrainOptions/RunDrain.
+func (o *Options) SetPodProvider(podProvider PodProvider) {
+	o.podProvider = podProvider
+}
+
 // RunDrain runs the 'drain' command
 func (o *Options) RunDrain(ctx context.Context) error {
 	o.drainStartedOn = time.Now()
@@ -239,7 +265,7 @@ func (o *Options) RunDrain(ctx context.Context) error {
 		klog.Errorf("Drain Error: Cordoning of node failed with error: %v", err)
 		return err
 	}
-	if !cache.WaitForCacheSync(drainContext.Done(), o.podSynced) {
+	if o.podSynced != nil && !cache.WaitForCacheSync(drainContext.Done(), o.podSynced) {
 		err := fmt.Errorf("timed out waiting for pod cache to sync")
 		return err
 	}
@@ -249,14 +275,14 @@ func (o *Options) RunDrain(ctx context.Context) error {
 }
 
 func (o *Options) deleteOrEvictPodsSimple(ctx context.Context) error {
-	pods, err := o.getPodsForDeletion()
+	pods, err := o.getPodsForDeletion(ctx)
 	if err != nil {
 		return err
 	}
 
 	err = o.deleteOrEvictPods(ctx, pods)
 	if err != nil {
-		pendingPods, newErr := o.getPodsForDeletion()
+		pendingPods, newErr := o.getPodsForDeletion(ctx)
 		if newErr != nil {
 			return newErr
 		}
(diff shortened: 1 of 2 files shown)


pkg/util/provider/drain/drain.go
```diff
@@ -77,6 +77,25 @@ type Options struct {
...
 	podSynced                    cache.InformerSynced
+
+	// AdditionalPodFilters are extra predicates evaluated alongside the built-in filters when selecting pods to
+	// drain.
+	AdditionalPodFilters []AdditionalPodFilter
+	// podProvider, if set, is used to list the pods on the node instead of podLister. It allows callers that do not
+	// run client-go informers (e.g. controller-runtime based controllers) to supply their own pod source.
+	podProvider PodProvider
+	// SkipVolumeHandling, if true, skips the PersistentVolume detach/reattach handling during eviction.
+	SkipVolumeHandling bool
```
takoverflow, line 90: Consider renaming this to `SkipVolumeDetach` instead.

`Handling` being very generic doesn't indicate without looking at the doc-string what's being skipped.

pkg/util/provider/drain/drain.go
```diff
@@ -79,6 +79,25 @@ type Options struct {
...
+}
+
+// AdditionalPodFilter takes a pod and returns whether the pod should be included for draining (true) or excluded
+// (false). It is a caller-supplied predicate evaluated in addition to the built-in filters.
+type AdditionalPodFilter func(corev1.Pod) bool
```
takoverflow, line 95:
```suggestion
type AdditionalPodFilter func(corev1.Pod) (include bool)
```
So the bool value makes sense even without the docstring, similar to how `podFilter` is defined.

pkg/util/provider/drain/drain.go
```diff
@@ -79,6 +79,25 @@ type Options struct {
...
+
+// PodProvider lists the pods running on a given node. It abstracts the pod source (client-go lister vs. a
+// controller-runtime client) used during drain.
+type PodProvider interface {
+	PodsForNode(ctx context.Context, nodeName string) ([]corev1.Pod, error)
```
takoverflow, line 100: Should it be `[]*corev1.Pod` to be consistent with the output from the lister as well. This is what is expected.
```suggestion
	PodsForNode(ctx context.Context, nodeName string) ([]*corev1.Pod, error)
```

Alternatively, if what I've suggested in [comment below](https://github.com/gardener/machine-controller-manager/pull/1130#discussion_r3879459452) is done, then it can return `[]corev1.Pod` as well (to keep things simple for gardener) with the lister `PodsForNode` taking care of converting `[]*corev1.Pod` to node-specific `[]corev1.Pod`.

pkg/util/provider/drain/drain.go
```diff
@@ -346,10 +372,21 @@ func (ps podStatuses) Message() string {
 
 // getPodsForDeletion returns all the pods we're going to delete.  If there are
 // any pods preventing us from deleting, we return that list in an error.
-func (o *Options) getPodsForDeletion() (pods []corev1.Pod, err error) {
-	podList, err := o.podLister.List(labels.Everything())
-	if err != nil {
-		return
+func (o *Options) getPodsForDeletion(ctx context.Context) (pods []corev1.Pod, err error) {
+	var podList []*corev1.Pod
+	if o.podProvider != nil {
```
takoverflow, line 377: Can `NewDrainOptions` be modified to construct the `podProvider` using the lister.

Something like
```go
type podProvider struct {
	Lister corelisters.PodLister
}

func (p *podProvider) PodsForNode(_ context.Context, nodeName string) ([]*corev1.Pod, error) {
	podList, err := p.Lister.List(labels.Everything())
	if err != nil {
		return nil, err
	}

	pods := make([]*corev1.Pod, 0)
	for _, pod := range podList {
		if pod.Spec.NodeName == nodeName {
			pods = append(pods, pod)
		}
	}
	return pods, nil
}
```

Then the `podProvider` would be initialised using the passed `podLister` in `NewDrainOptions`

Later on, any external consumer like gardener can then override the default. This would result in these `if-else` no longer being needed. Also the `podProvider` then actually works with the lister and is not just used in a one-off scenario.

This keeps `getPodsForDeletion` very simple and the loop below can also get rid of the nodename check, i.e. this can be removed.
```go
for _, pod := range podList {
		if pod.Spec.NodeName != o.nodeName {
			continue
		} 
...
```

WDYT?

takoverflow (commented):
Thanks for the PR, I've added some comments PTAL

pkg/util/provider/drain/drain.go
acumino: This is nice idea.

pkg/util/provider/drain/drain.go
takoverflow: This check is redundant now, `PodsForNode` already takes care of it. `podList` will never contain pods not scheduled on the specified node.

acumino:
cc @takoverflow 
PTAL.

pkg/util/provider/drain/drain.go
```diff
@@ -473,15 +530,15 @@ func (o *Options) evictPods(ctx context.Context, attemptEvict bool, pods []corev
...
 		podsToDrain := make([]*corev1.Pod, len(pods))
 		for i := range pods {
 			podsToDrain[i] = &pods[i]
 		}
 
-		klog.V(3).Infof("Forceful eviction of pods on the node: %q", o.nodeName)
+		klog.V(3).Infof("Eviction of pods on the node without volume handling: %q", o.nodeName)
 
-		// evict all pods in parallel without waiting for pods or volume detachment
+		// Evict all pods without waiting for volume detachment.
```
takoverflow: Was there a reason why this comment was changed? We're still evicting in parallel right?

pkg/util/provider/drain/drain_test.go
```diff
@@ -861,8 +860,175 @@ var _ = Describe("drain", func() {
...
+				Out:                          GinkgoWriter,
+				nodeName:                     nodeName,
+				IgnorePodsWithoutControllers: true,
+				IgnoreDaemonsets:             true,
+				DeleteLocalData:              true,
+			}
+		})
+
+		Context("PodProvider", func() {
+			It("should returns only pods on the target node from PodProvider", func() {
```
takoverflow:
```suggestion
			It("should return only pods on the target node from PodProvider", func() {
```

pkg/util/provider/drain/drain_test.go
```diff
@@ -861,8 +860,175 @@ var _ = Describe("drain", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+				Expect(pods).To(HaveLen(4))
+			})
+
+			It("filters pods with a single additional filter", func() {
+				// Keep only pods whose name ends with "0" or "1"
+				drain.AdditionalPodFilters = []AdditionalPodFilter{
+					func(pod corev1.Pod) bool {
+						return pod.Name == "pod-0" || pod.Name == "pod-1"
+					},
```
takoverflow, line 994: As per the comment above, the filter should be
```go
func(pod corev1.Pod) bool {
	return strings.HasSuffix(pod.Name, "0") || strings.HasSuffix(pod.Name, "1")
},

takoverflow (commented):
Apologies for the late response, just very minor nits for the test, rest looks good to me. Thanks for the changes!

pkg/util/provider/drain/drain.go
acumino: Yes.

takoverflow (commented):
/lgtm

PR #1135, opened 2026-08-20
Fix handling of NotFound errors when deciding if a Failed machine should be terminated

thiyyakat:

Fixes the handling of `NotFound` errors in `shouldFailedMachineBeTerminated()` and refactors the shared preservation helpers so both the machineset controller and the machine controller rely on the same logic.

Previously, `shouldFailedMachineBeTerminated()` treated *any* error while resolving the effective preserve value (including a `NotFound` on the backing node) as a signal to terminate the failed machine. As a result, a machine explicitly annotated for preservation (`preserve=now` / `when-failed` / `auto-preserved`) whose node is not found was terminated instead of being preserved.

This PR:
- Moves the preservation state helpers (`PreserveStateInfo`, `GetPreserveStateInfo`, `GetEffectivePreservationAnnotations`) into `pkg/util/provider/machineutils` so they can be shared, and adds `IsPositivePreserveValue` to distinguish values that actually request preservation (`now`/`when-failed`/`auto-preserved`) from `false`/empty/unrecognized values.
- Rewrites `shouldFailedMachineBeTerminated()` to:
  - fall through on a `NotFound` node error (enforcing the machine's own annotation), and only terminate on other errors,
  - terminate once preservation has expired,
  - keep the machine only when the effective preserve value is a positive value.

Behavior notes:
- A failed machine with `preserve=false`, an empty, or an unrecognized value is terminated (matching the prior `default` behavior).
- A failed machine annotated for preservation whose node is `NotFound` is now preserved (per the documented T1–T4 recovery semantics), rather than terminated.

Fixes #

- `NotFound` handling has two sub-cases worth calling out:
  - machine annotated with a positive preserve value + node `NotFound` → **preserved** (intended change from prior behavior).
  - only `last-applied-node-preserve-value` set (no active machine annotation) + node `NotFound` → **terminated**, unchanged; this is outcome-consistent with `manageMachinePreservation`, which stops preservation for the same state. `last-applied-node-preserve-value` is used as chronological history, not live intent.
- No functional change to `manageMachinePreservation` beyond consuming the relocated/exported helpers.
- IT passed for provider-aws

**Release note**:

```bugfix operator
Fixed handling of `NotFound` errors in the failed-machine termination check: a failed machine annotated for preservation whose backing node no longer exists is now preserved instead of being terminated.
```

### DIFF
--- a/pkg/controller/machineset.go
+++ b/pkg/controller/machineset.go
@@ -910,28 +910,23 @@ func (c *controller) shouldFailedMachineBeTerminated(machine *v1alpha1.Machine) 
 		node      *corev1.Node
 		err       error
 	)
+	if machineutils.HasMachinePreservationExpired(machine) {
+		klog.V(3).Infof("Preservation of failed machine %q has timed out at %v.", machine.Name, machine.Status.CurrentStatus.PreserveExpiryTime)
+		return true
+	}
+
 	nodeName := machine.Labels[v1alpha1.NodeLabelKey]
-	// We don't return on error until it is determined whether the machine has valid preservation state
 	if nodeName != "" {
 		node, err = c.nodeLister.Get(nodeName)
 		if err != nil {
-			if !apierrors.IsNotFound(err) {
-				klog.Errorf("error finding preservation state for machine %q: %v. Proceeding with termination of the machine.", machine.Name, err)
-				return true
-			}
-			klog.Warningf("node %q not found for machine %q.", nodeName, machine.Name)
+			klog.Warningf("Error fetching node %q of machine %q: %v", nodeName, machine.Name, err)
 		} else {
 			nodeFound = true
 		}
 	}
 
-	if machineutils.IsMachinePreservationExpired(machine) {
-		klog.V(3).Infof("Preservation of failed machine %q has timed out at %v", machine.Name, machine.Status.CurrentStatus.PreserveExpiryTime)
-		return true
-	}
-
 	preserveInfo := machineutils.GetPreserveStateInfo(node, machine)
-	if machineutils.IsPositivePreserveValue(machineutils.GetEffectivePreservationAnnotations(&preserveInfo, nodeFound)) {
+	if machineutils.IsPreservationRequested(machineutils.GetEffectivePreservationAnnotations(&preserveInfo, nodeFound)) {
 		klog.V(3).Infof("Failed machine %q is either preserved or in the process of being preserved.", machine.Name)
 		return false
 	}
--- a/pkg/controller/machineset.go
+++ b/pkg/controller/machineset.go
@@ -31,6 +31,7 @@ import (
 	"sync"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -904,29 +905,37 @@ func isMachineStatusEqual(s1, s2 v1alpha1.MachineStatus) bool {
 // or if it is a candidate for auto-preservation. If none of these conditions are met, it returns true indicating
 // that the failed machine should be terminated.
 func (c *controller) shouldFailedMachineBeTerminated(machine *v1alpha1.Machine) bool {
-	// if preserve expiry time is set and is in the future, machine is already preserved
-	if machine.Status.CurrentStatus.PreserveExpiryTime != nil {
-		if machine.Status.CurrentStatus.PreserveExpiryTime.After(time.Now()) {
-			klog.V(3).Infof("Failed machine %q is preserved until %v", machine.Name, machine.Status.CurrentStatus.PreserveExpiryTime)
-			return false
+	var (
+		nodeFound bool
+		node      *corev1.Node
+		err       error
+	)
+	nodeName := machine.Labels[v1alpha1.NodeLabelKey]
+	// We don't return on error until it is determined whether the machine has valid preservation state
+	if nodeName != "" {
+		node, err = c.nodeLister.Get(nodeName)
+		if err != nil {
+			if !apierrors.IsNotFound(err) {
+				klog.Errorf("error finding preservation state for machine %q: %v. Proceeding with termination of the machine.", machine.Name, err)
+				return true
+			}
+			klog.Warningf("node %q not found for machine %q.", nodeName, machine.Name)
+		} else {
+			nodeFound = true
 		}
+	}
+
+	if machineutils.IsMachinePreservationExpired(machine) {
 		klog.V(3).Infof("Preservation of failed machine %q has timed out at %v", machine.Name, machine.Status.CurrentStatus.PreserveExpiryTime)
 		return true
 	}
-	preserveValue, err := c.findEffectivePreserveValue(machine)
-	if err != nil {
-		// in case of error fetching node or annotations, we don't want to block deletion of failed machines, so we return true
-		klog.Errorf("error finding effective preserve value for machine %q: %v. Proceeding with termination of the machine.", machine.Name, err)
-		return true
-	}
-	switch preserveValue {
-	case machineutils.PreserveMachineAnnotationValueWhenFailed, machineutils.PreserveMachineAnnotationValueNow, machineutils.PreserveMachineAnnotationValueAutoPreserved: // this is in case preservation process is not complete yet
+
+	preserveInfo := machineutils.GetPreserveStateInfo(node, machine)
+	if machineutils.IsPositivePreserveValue(machineutils.GetEffectivePreservationAnnotations(&preserveInfo, nodeFound)) {
+		klog.V(3).Infof("Failed machine %q is either preserved or in the process of being preserved.", machine.Name)
 		return false
-	case machineutils.PreserveMachineAnnotationValueFalse:
-		return true
-	default:
-		return true
 	}
+	return true
 }
 
 // manageAutoPreservationOfFailedMachines annotates failed machines with preserve=auto-preserved annotation
@@ -1018,24 +1027,4 @@ func addAutoPreserveAnnotationOnMachine(machineToUpdate *v1alpha1.Machine) error
 func removeAutoPreserveAnnotationFromMachine(machineToUpdate *v1alpha1.Machine) error {
 	delete(machineToUpdate.Annotations, machineutils.PreserveMachineAnnotationKey)
 	return nil
-}
-
-func (c *controller) findEffectivePreserveValue(machine *v1alpha1.Machine) (string, error) {
-	var nodeAnnotationValue, machineAnnotationValue, lANodeAnnotationValue string
-	machineAnnotationValue = machine.Annotations[machineutils.PreserveMachineAnnotationKey]
-	lANodeAnnotationValue = machine.Annotations[machineutils.LastAppliedNodePreserveValueAnnotationKey]
-	nodeName := machine.Labels[v1alpha1.NodeLabelKey]
-	if nodeName != "" {
-		node, err := c.nodeLister.Get(nodeName)
-		if err != nil {
-			klog.Errorf("error fetching node %q for machine %q: %v", nodeName, machine.Name, err)
-			return "", err
-		}
-		nodeAnnotationValue = node.Annotations[machineutils.PreserveMachineAnnotationKey]
-	}
-	if nodeAnnotationValue == "" && lANodeAnnotationValue == "" {
-		return machineAnnotationValue, nil
-	} else {
-		return nodeAnnotationValue, nil
-	}
 }
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -774,14 +774,17 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
 		}
 	}()
 	var (
-		nodeFound bool
-		node      *corev1.Node
+		nodeFound         bool
+		node              *corev1.Node
+		removeAnnotations bool
 	)
 	nodeName := machine.Labels[v1alpha1.NodeLabelKey]
 	if nodeName != "" {
 		node, err = c.nodeLister.Get(nodeName) // We don't return on error immediately because we need to determine whether the machine has valid preservation state
 		if err == nil {
 			nodeFound = true
+		} else {
+			klog.Warningf("Error fetching node %q for machine %q: %v", nodeName, machine.Name, err)
 		}
 	}
 	preserveInfo := machineutils.GetPreserveStateInfo(node, machine)
@@ -802,22 +805,12 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
 		return
 	}
 
-	if err != nil {
-		if !apierrors.IsNotFound(err) {
-			return
-		}
-		err = nil
-		klog.Warningf("Couldn't find node %q for machine %q", nodeName, machine.Name)
-	}
-
-	// Note: when the backing node cannot be found, we assume the machine's annotation value needs to be enforced to enable
+	// Note: when the backing node cannot be fetched, we assume the machine's annotation value needs to be enforced to enable
 	// preservation of the machine object.
 	effectivePreserveValue := machineutils.GetEffectivePreservationAnnotations(&preserveInfo, nodeFound)
-
-	var removeAnnotations bool
 	clone := machine.DeepCopy()
 	switch effectivePreserveValue {
-	// effectivePreserveValue == "" implies the preservation annotation was deleted to indicate that
+	// effectivePreserveValue == "" implies the preservation annotation was deleted to express intent that
 	// preservation must be stopped
 	case "", machineutils.PreserveMachineAnnotationValueFalse:
 		clone, err = c.stopPreservationIfActive(ctx, clone, removeAnnotations)
--- a/pkg/util/provider/machinecontroller/machine.go
+++ b/pkg/util/provider/machinecontroller/machine.go
@@ -780,6 +780,9 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
 	nodeName := machine.Labels[v1alpha1.NodeLabelKey]
 	if nodeName != "" {
 		node, err = c.nodeLister.Get(nodeName) // We don't return on error immediately because we need to determine whether the machine has valid preservation state
+		if err == nil {
+			nodeFound = true
+		}
 	}
 	preserveInfo := machineutils.GetPreserveStateInfo(node, machine)
 	if preserveInfo.MachineAnnotated && !machineutils.AllowedPreserveAnnotationValues.Has(preserveInfo.MachineValue) {
@@ -805,8 +808,6 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
 		}
 		err = nil
 		klog.Warningf("Couldn't find node %q for machine %q", nodeName, machine.Name)
-	} else {
-		nodeFound = true
 	}
 
 	// Note: when the backing node cannot be found, we assume the machine's annotation value needs to be enforced to enable
@@ -822,7 +823,7 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
 		clone, err = c.stopPreservationIfActive(ctx, clone, removeAnnotations)
 	case machineutils.PreserveMachineAnnotationValueWhenFailed:
 		// on timing out, remove preserve annotation to prevent incorrect re-preservation
-		if machineutils.IsMachinePreservationExpired(clone) {
+		if machineutils.HasMachinePreservationExpired(clone) {
 			removeAnnotations = true
 			clone, err = c.stopPreservationIfActive(ctx, clone, removeAnnotations)
 		} else if !machineutils.IsMachineFailed(clone) {
@@ -831,7 +832,7 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
 			clone, err = c.preserveMachine(ctx, clone, effectivePreserveValue)
 		}
 	case machineutils.PreserveMachineAnnotationValueNow:
-		if machineutils.IsMachinePreservationExpired(clone) {
+		if machineutils.HasMachinePreservationExpired(clone) {
 			// on timing out, remove preserve annotation to prevent incorrect re-preservation
 			removeAnnotations = true
 			clone, err = c.stopPreservationIfActive(ctx, clone, removeAnnotations)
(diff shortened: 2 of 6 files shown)


thiyyakat:
Fix was verified by doing the following:
1. In both the master and fix versions, the code in the machine controller that handles auto-preservation was commented out. This way, PreserveExpiryTime is never set on an auto-preserved machine.
2. Set autoPreserveFailedMachineMax to 1 on the MCD.
3. Set machineCreationTimeout to a small value to force a new machine to go to Pending and then Failed due to creation timing out. No node has registered. Therefore, this would result in a NotFound error when the node is fetched from the lister.
4. When machine fails, the MCS annotates it with auto-preserved.
5. On master, this machine is deleted because there is no PreserveExpiryTime set, and the node could not be found.
6. With the fix, the machine stays preserved.

**Relevant logs** 

For the run with the fix:
```
W0821 15:29:55.836179   28925 machineset.go:922] node "ip-10-180-18-3.eu-west-1.compute.internal" not found for machine "shoot--xxx--ca-test1-34-sys-comp-z1-7f667-krs58".
I0821 15:29:55.836211   28925 machineset.go:935] Failed machine "shoot--xxx--ca-test1-34-sys-comp-z1-7f667-krs58" is either preserved or in the process of being preserved.
```
Machine stays in Failed phase.

For the run with master:
```
E0821 15:32:24.796181   31384 machineset.go:1031] error fetching node "ip-10-180-18-3.eu-west-1.compute.internal" for machine "shoot--xxx--ca-test1-34-sys-comp-z1-7f667-krs58": node "ip-10-180-18-3.eu-west-1.compute.internal" not found
E0821 15:32:24.796252   31384 machineset.go:919] error finding effective preserve value for machine "shoot--xxx--ca-test1-34-sys-comp-z1-7f667-krs58": node "ip-10-180-18-3.eu-west-1.compute.internal" not found. Proceeding with termination of the machine.
```
Machine moves to Terminating after this.

pkg/controller/machineset.go
```diff
@@ -904,29 +905,37 @@ func isMachineStatusEqual(s1, s2 v1alpha1.MachineStatus) bool {
...
+		err       error
+	)
+	nodeName := machine.Labels[v1alpha1.NodeLabelKey]
+	// We don't return on error until it is determined whether the machine has valid preservation state
+	if nodeName != "" {
+		node, err = c.nodeLister.Get(nodeName)
+		if err != nil {
+			if !apierrors.IsNotFound(err) {
+				klog.Errorf("error finding preservation state for machine %q: %v. Proceeding with termination of the machine.", machine.Name, err)
+				return true
```
r4mek: do we want to terminate the machine if we we get an error in `Get()` or just ignore it?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -785,35 +773,45 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
 		err = nil
 		return
-	} else if getErr != nil {
-		if !apierrors.IsNotFound(getErr) {
-			err = getErr
+	}
+
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
 			return
```
r4mek: should we return here or ignore this?

pkg/util/provider/machineutils/utils.go
```diff
@@ -211,3 +223,60 @@ func UpdateMachineWithRetries(ctx context.Context, machineClient v1alpha1client.
...
+//
+// lastAppliedNodeValue is required to handle the following scenario:
+//
+//	T1: Node and Machine both have the same annotation with the same value. (MCM is up and running).
+//	T2 (T2 > T1): MCM went down.
+//	T3 (T3 > T2): Node annotation was removed.
+//	T4 (T4 > T3): MCM came back up.
+//	At T4 it sees a Node with no preserve annotation but a Machine with a preserve annotation.
+//	It continues to preserve the machine.
+func GetEffectivePreservationAnnotations(info *PreserveStateInfo, nodeFound bool) string {
```
r4mek, line 266: it might be nice to have `GetPreserveStateInfo()` and `GetEffectivePreservationAnnotations()` be a single function.

pkg/util/provider/machineutils/utils.go
```diff
@@ -211,3 +223,60 @@ func UpdateMachineWithRetries(ctx context.Context, machineClient v1alpha1client.
...
+	if !machine.Status.CurrentStatus.PreserveExpiryTime.IsZero() {
+		info.PreserveExpiryTimeSet = true
+	}
+	return info
+}
+
+// IsPositivePreserveValue returns true when value is a preserve annotation value that requests
+// preservation (now/when-failed/auto-preserved). The value "false", empty, or any unrecognized
+// value is not a positive preserve value.
+func IsPositivePreserveValue(value string) bool {
```
r4mek: ideally we should have a function which gives us the preserve annotation value by checking node and machine annotation. We should only get annotation values which are valid, else error. So this function would not be needed.

pkg/controller/machineset.go
```diff
@@ -904,29 +905,37 @@ func isMachineStatusEqual(s1, s2 v1alpha1.MachineStatus) bool {
...
+		nodeFound bool
+		node      *corev1.Node
+		err       error
+	)
+	nodeName := machine.Labels[v1alpha1.NodeLabelKey]
+	// We don't return on error until it is determined whether the machine has valid preservation state
+	if nodeName != "" {
+		node, err = c.nodeLister.Get(nodeName)
+		if err != nil {
+			if !apierrors.IsNotFound(err) {
```
takoverflow: What if there was any other error for listing the node? Then we unconditionally don't honor preservation? Why is that, preservation state can still be inferred from the `Machine` object right?

pkg/controller/machineset.go
```diff
@@ -904,29 +905,37 @@ func isMachineStatusEqual(s1, s2 v1alpha1.MachineStatus) bool {
...
-		klog.Errorf("error finding effective preserve value for machine %q: %v. Proceeding with termination of the machine.", machine.Name, err)
+
+	if machineutils.IsMachinePreservationExpired(machine) {
+		klog.V(3).Infof("Preservation of failed machine %q has timed out at %v", machine.Name, machine.Status.CurrentStatus.PreserveExpiryTime)
 		return true
 	}
-	switch preserveValue {
-	case machineutils.PreserveMachineAnnotationValueWhenFailed, machineutils.PreserveMachineAnnotationValueNow, machineutils.PreserveMachineAnnotationValueAutoPreserved: // this is in case preservation process is not complete yet
+
+	preserveInfo := machineutils.GetPreserveStateInfo(node, machine)
```
takoverflow, line 928: In cases when lister returned `NotFound` error, the `node` object will be an empty one. I see that `GetPreserveStateInfo` then checks if `node != nil`, why can't the same be done in case of other errors when listing the node? Why take the destructive route of not honoring preservation?

pkg/util/provider/machineutils/utils.go
```diff
@@ -211,3 +223,60 @@ func UpdateMachineWithRetries(ctx context.Context, machineClient v1alpha1client.
...
+//
+// If there is no active node annotation AND no previously-applied node annotation,
+// enforce machine's preserve annotation.
+// Otherwise, the node annotation takes precedence (even if now empty/removed).
+//
+// lastAppliedNodeValue is required to handle the following scenario:
+//
+//	T1: Node and Machine both have the same annotation with the same value. (MCM is up and running).
+//	T2 (T2 > T1): MCM went down.
+//	T3 (T3 > T2): Node annotation was removed.
```
takoverflow: In this case, what was the reason the node annotation was removed?

If explicitly removed, isn't the intention to not preserve the machine anymore? I'm a bit confused by this example. Can you clarify?

pkg/controller/machineset.go
```diff
@@ -904,29 +905,37 @@ func isMachineStatusEqual(s1, s2 v1alpha1.MachineStatus) bool {
...
 		}
-		klog.V(3).Infof("Preservation of failed machine %q has timed out at %v", machine.Name, machine.Status.CurrentStatus.PreserveExpiryTime)
-		return true
 	}
-	preserveValue, err := c.findEffectivePreserveValue(machine)
-	if err != nil {
-		// in case of error fetching node or annotations, we don't want to block deletion of failed machines, so we return true
-		klog.Errorf("error finding effective preserve value for machine %q: %v. Proceeding with termination of the machine.", machine.Name, err)
+
+	if machineutils.IsMachinePreservationExpired(machine) {
```
takoverflow:
```suggestion
	if machineutils.HasMachinePreservationExpired(machine) {
```

Nit, shouldn't it be "has expired" instead?

pkg/util/provider/machineutils/utils.go
takoverflow: Can this be renamed to something other than positive?
Maybe `IsPreservationRequested` as per the doc-string? 

Also since it's only being used in one place right now, there also I can't understand just based on usage what is being checked, wouldn't it be better to just inline it where its being used? What was the reason behind extracting it out?

pkg/controller/machineset.go
takoverflow: The behavior isn't consistent across different usages.

As part of `manageMachinePreservation`, if there's an error getting the node and the error isn't `NotFound` error, we just log a warning but still compute `effectivePreserveValue` from the preserveInfo.

```
	if nodeName != "" {
		node, err = c.nodeLister.Get(nodeName) // We don't return on error immediately because we need to determine whether the machine has valid preservation state
	}
...
	if err != nil {
		if !apierrors.IsNotFound(err) {
			return
		}
		err = nil
		klog.Warningf("Couldn't find node %q for machine %q", nodeName, machine.Name)
	} else {
		nodeFound = true
	}

	// Note: when the backing node cannot be found, we assume the machine's annotation value needs to be enforced to enable
	// preservation of the machine object.
	effectivePreserveValue := machineutils.GetEffectivePreservationAnnotations(&preserveInfo, nodeFound)

```

But here we don't do the same.

takoverflow (changes requested):
Thanks for the changes.

Please don't make further changes, still in the process of reviewing the PR. (More changes will make it difficult to recall what I've already looked at)

thiyyakat:
> Please don't make further changes, still in the process of reviewing the PR. 

Do you mean I shouldn't make the changes you recommended yet? Or push any unrelated changes?

pkg/util/provider/machineutils/utils.go
thiyyakat: > IsPreservationRequested

I like this 👍 

> Also since it's only being used in one place right now, there also I can't understand just based on usage what is being checked, wouldn't it be better to just inline it where its being used? What was the reason behind extracting it out?

The name of the function explains the code without needing a comment. It will also improve testability (that reminds me, I need to add tests here!). Especially, if we are planning to change the annotation values.

I would like to retain this as a helper, if you don't feel very strongly about in-lining it.

pkg/util/provider/machineutils/utils.go
thiyyakat: > In this case, what was the reason the node annotation was removed?
To stop preservation.

> If explicitly removed, isn't the intention to not preserve the machine anymore? I'm a bit confused by this example. Can you clarify?

Yes. You're right. Without the LastAppliedNodeAnnotationValue annotation, we would end up continuing to preserve the machine instead of honouring the intent expressed by the removal of the annotation.

I think a better place for this docstring would be wherever LastAppliedNodeAnnotationValue is declared. The example explains why we need the annotation. Will rewrite to remove the ambiguity.

takoverflow:
I would prefer both for now, because it's a bit hard to track what change has what kind of impact without revisiting the entire call stack 😅

Just give me one day to go through the entire thing. In between changes are difficult to reason about is all.

Thanks!

thiyyakat:
> I would prefer both for now, because it's a bit hard to track what change has what kind of impact without revisiting the entire call stack 😅

Sure 👍 Will hold off on making the changes you requested until you give me the go-ahead.

pkg/controller/machineset.go
takoverflow: Nevermind, brain fog!
We're actually returning here as well.
But I'd still like to understand why.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -785,35 +773,45 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
 		}
-		klog.Warningf("Couldn't find node %q for machine %q", nodeName, machine.Name)
 		err = nil
+		klog.Warningf("Couldn't find node %q for machine %q", nodeName, machine.Name)
+	} else {
+		nodeFound = true
```
gagan16k: `nodefound = true` would be set when `nodeName` is "" as well due to `err` being `nil`, gated by the check at L781
```
	if nodeName != "" {
		node, err = c.nodeLister.Get(nodeName) // We don't return on error immediately because we need to determine whether the machine has valid preservation state
	}
```
Could move `nodefound = true` into this block

pkg/controller/machineset.go
gagan16k: A nit, but can we move this to above the node fetch as it is not using the node.

pkg/controller/machineset.go
```diff
@@ -904,29 +905,37 @@ func isMachineStatusEqual(s1, s2 v1alpha1.MachineStatus) bool {
...
-		if machine.Status.CurrentStatus.PreserveExpiryTime.After(time.Now()) {
-			klog.V(3).Infof("Failed machine %q is preserved until %v", machine.Name, machine.Status.CurrentStatus.PreserveExpiryTime)
-			return false
+	var (
+		nodeFound bool
+		node      *corev1.Node
+		err       error
+	)
+	nodeName := machine.Labels[v1alpha1.NodeLabelKey]
+	// We don't return on error until it is determined whether the machine has valid preservation state
```
takoverflow: This comment is not correct, when there's an error, we return early and never bother checking machine's preservation state.

pkg/util/provider/machinecontroller/machine.go
thiyyakat: Good catch. Thanks!

pkg/controller/machineset.go
thiyyakat: Just thinking aloud here:

If node is not found in the lister, it is usually not a transient error AFAIK. In that case, it makes sense to defer to the machine's value to manipulate preservation.

Trouble is, if it is a transient error (there's no way of knowing, I understand that), isn't it better to retry rather than change the current state? Or do we defer to the machine in that case too? 

Also, if you see the rest of the preservation code, unless the nodename is "" or the node is not found, we do a lot of node updates. All of those would fail and cause returns. However, if the machine was marked with when-failed or now, the machine object at least would get preserved, even if the latter steps result in errors. So there is some value in deferring to the machine.

If we are confident that operators wouldn't leave stale annotations on the machine object causing an early, unintentional end to preservation, then we could defer to the machine object on all errors.

Also, what do you think the behaviour should be if the machine object is found to be un-annotated on a lister error? To me it seems like we should return the error, rather than assume preservation is not desired.

LMK what you think.

thiyyakat: @aaronfern, please share your thoughts on this as well. Thanks.

takoverflow: > Trouble is, if it is a transient error (there's no way of knowing, I understand that), isn't it better to retry rather than change the current state?

But we aren't retrying here. `ShouldFailedMachineBeTerminated` returns `true`.
```
			if !apierrors.IsNotFound(err) {
				klog.Errorf("error finding preservation state for machine %q: %v. Proceeding with termination of the machine.", machine.Name, err)
				return true
			}
```

> However, if the machine was marked with when-failed or now, the machine object at least would get preserved, even if the latter steps result in errors. So there is some value in deferring to the machine.

Sure, that is being done in `ManageMachinePreservation` where we actually retry for non-`NotFound` errors. But the behavior is not the same for `ShouldFailedMachineBeTerminated`. There if its a transient error listing the node, the `machine`'s preservation annotation is not even being checked, is that desirable? That is what I'm asking.

> If we are confident that operators wouldn't leave stale annotations on the machine object causing an early, unintentional end to preservation, then we could defer to the machine object on all errors.

> Also, what do you think the behaviour should be if the machine object is found to be un-annotated on a lister error? To me it seems like we should return the error, rather than assume preservation is not desired.

I think you misunderstood my question, I want preservation to be honored as well. https://github.com/gardener/machine-controller-manager/pull/1135#discussion_r3842286128 https://github.com/gardener/machine-controller-manager/pull/1135#discussion_r3842292578

pkg/util/provider/machineutils/utils.go
takoverflow: > I would like to retain this as a helper, if you don't feel very strongly about in-lining it.

Alright.

pkg/controller/machineset.go
thiyyakat: After offline discussion with @takoverflow and @r4mek, it was decided to defer to the machine's annotation value on all errors to allow operators to preserve machines whose backing nodes cannot be fetched (in this case it is currently only NotFound error). If the machine object is not annotated, we return `true`.

pkg/controller/machineset.go
thiyyakat: We can't ignore it unconditionally because the error may not be transient and preservation may not be desired for the machine+node.  So, we can instead defer to the machine object like @takoverflow suggested in his [comment](https://github.com/gardener/machine-controller-manager/pull/1135#discussion_r3842286128). If the machine object carries no request for preservation then we terminate the machine.

pkg/controller/machineset.go
thiyyakat: Addressed in 111e65456c0d3fad6a462f0a6af20b94592decf1

pkg/util/provider/machineutils/utils.go
thiyyakat: Moved the example to the docstring for `LastAppliedNodeAnnotationValue`, and also simplified the example. Changes made in 111e65456c0d3fad6a462f0a6af20b94592decf1

pkg/controller/machineset.go
thiyyakat: Addressed in 111e65456c0d3fad6a462f0a6af20b94592decf1

pkg/util/provider/machinecontroller/machine.go
thiyyakat: Made the changes in commit 111e65456c0d3fad6a462f0a6af20b94592decf1.

pkg/controller/machineset.go
thiyyakat: Moved it in commit 111e65456c0d3fad6a462f0a6af20b94592decf1.

pkg/controller/machineset.go
thiyyakat: Removed in commit 111e65456c0d3fad6a462f0a6af20b94592decf1.

pkg/util/provider/machineutils/utils.go
```diff
@@ -211,3 +234,46 @@ func UpdateMachineWithRetries(ctx context.Context, machineClient v1alpha1client.
...
+	switch value {
+	case PreserveMachineAnnotationValueNow, PreserveMachineAnnotationValueWhenFailed, PreserveMachineAnnotationValueAutoPreserved:
+		return true
+	default:
+		return false
+	}
+}
+
+// GetEffectivePreservationAnnotations returns the effective preservation value.
+func GetEffectivePreservationAnnotations(info *PreserveStateInfo, nodeFound bool) string {
```
gagan16k, line 266: Can we fold `nodeFound` bool into `PreserveStateInfo`? Unless there is a use case where we will have to pass a different value for the same node, in which case it needs to be called something else.

pkg/controller/machineset.go
```diff
@@ -904,29 +905,32 @@ func isMachineStatusEqual(s1, s2 v1alpha1.MachineStatus) bool {
...
-			klog.V(3).Infof("Failed machine %q is preserved until %v", machine.Name, machine.Status.CurrentStatus.PreserveExpiryTime)
-			return false
-		}
-		klog.V(3).Infof("Preservation of failed machine %q has timed out at %v", machine.Name, machine.Status.CurrentStatus.PreserveExpiryTime)
+	var (
+		nodeFound bool
+		node      *corev1.Node
+		err       error
+	)
+	if machineutils.HasMachinePreservationExpired(machine) {
```
gagan16k, line 913: ~The whole condition is essentially inverted now. 
Would it be possible there would be a brief window where a machine that is being un-preserved due to preserve expiry can have the preserve annotation off the machine/node, but the expiry time has not been cleared yet?
Then potentially, a healthy machine could get marked for deletion.~

Edit: Healthy machines will not be part of this code path

pkg/util/provider/machineutils/utils.go
thiyyakat: I've been wondering if I should do this. However, is it really a part of preservation state though? Moreover, if say initially the node was not found because the cache was not updated and the node has only just registered. Towards the end of `manageMachinePreservation` we fetch the node again before removing the taint. If we instead use the value computed by `GetEffectivePreservationAnnotations`, then this value would be stale, and the taint may not get removed in that reconciliation. It is self-healing, but what I'm trying to ask is whether we can assume `nodeFound` to be a static value like the rest of `preserveStateInfo` within a reconciliation.

pkg/util/provider/machinecontroller/machine.go
thiyyakat: Ignoring error, as discussed offline. Addressed in 2f7ed546d5751eb8511f878e0e594c6d84de56ee. PTAL

pkg/util/provider/machineutils/utils.go
thiyyakat: Positive values are different from valid values. By positive value(which has been renamed now ref: https://github.com/gardener/machine-controller-manager/pull/1135#discussion_r3842649849), I meant the annotation requested for preservation to happen (when-failed,now,auto-preserved), rather than stopping preservation ("" and false).

pkg/util/provider/machineutils/utils.go
thiyyakat: There is some use to having them separate actually. If we only need to figure out if a machine has preservation state (I needed to do this for the preservation during updates issue), then we need only `GetPreserveStateInfo()` and `isMachinePreservationBound()`. We don't really care about what the effective value is in that case.

pkg/util/provider/machineutils/utils.go
gagan16k: Looking at the code, I was thinking if we even need this flag in the first place. The only place where it is consumed is by `getEffectivePreserveInfo()`and it's whole purpose is to return `info.MachineValue` if node is not found. As an add-on to @r4mek's comment [here](https://github.com/gardener/machine-controller-manager/pull/1135#discussion_r3841202728), we could have an `effectivePreservationValue` in `GetPreserveStateInfo()`, which can be used/discarded as required.
This might solve both cases, where we still re-fetch the node for untainting-purposes

thiyyakat: This is a good idea. @r4mek has made the change in his recent PR, which removes the need for `nodeFound` completely: https://github.com/gardener/machine-controller-manager/pull/1147/changes#diff-f3b8457c00ac7e0985403ca50f82c97e83e70d13431e96e03e23fa262290cc47R783-R785

thiyyakat:
/hold

PR:https://github.com/gardener/machine-controller-manager/pull/1147 needs to go in before this one.

PR #1136, opened 2026-08-25
Improve behaviour of `Preserved` Node Condition

thiyyakat:

Improves the behaviour of the `Preserved` node condition in the machine preservation flow:

- **Remove condition on preservation stop**: Instead of setting the condition to `ConditionFalse` when preservation ends, the condition is now removed entirely. 
- **Machine-code `Reason` field**: The `Reason` field now follows the Kubernetes convention of using a `CamelCase` machine-readable token (`PreservationInProgress`, `PreservationWithoutDrainCompleted`, `PreservationWithDrainCompleted`, `DrainFailed`) so it can be used for programmatic branching. Human-readable context is moved to `Message`.
- **Early condition initialisation**: The `Preserved` node condition is now set to `PreservationInProgress` before draining begins, so users can observe that MCM has acknowledged the preservation annotation and is acting on it. This also allows gardenlet health checks to exempt Failed preserved machines sooner thereby reducing alerting noise.
- **Message carries drain outcome and preserve-by info**: The `Message` field now includes whether the node was auto-preserved by MCM or by the user, the drain outcome, and the expiry time.

Fixes #

`computeNewNodePreservedCondition` is split into `recomputePreservedNodeCondition` (pure computation) and `needsPreservedNodeConditionUpdate` (dirty check), and a new `RemoveConditionFromNode` helper is added in `pkg/util/nodeops/conditions.go`.

Integration tests introduced in https://github.com/gardener/machine-controller-manager/pull/1124 was run against the changes. All the tests pass.

**Release note**:

```bugfix operator
The `Preserved` node condition is initialised early (before drain) so operators can observe preservation progress, and it is removed (rather than set to `False`) when preservation stops.
```

### DIFF
--- a/pkg/apis/machine/v1alpha1/machine_types.go
+++ b/pkg/apis/machine/v1alpha1/machine_types.go
@@ -244,24 +244,23 @@ const (
 	UpdateFailed string = "UpdateFailed"
 )
 
+// Constants used by the preservation flow.
 const (
-	// NodePreserved is a node condition type for preservation of machines to allow end-user to know that a node is preserved
+	// NodePreserved is a node condition type that surfaces preservation information to the end-users.
 	NodePreserved corev1.NodeConditionType = "Preserved"
 
-	// PreservedByMCM is a node condition reason for preservation of machines to indicate that the node is auto-preserved by MCM
-	PreservedByMCM string = "Preserved by MCM."
+	// PreservationInProgress is a node condition reason indicating preservation has started but is not yet complete.
+	PreservationInProgress string = "PreservationInProgress"
 
-	// PreservedByUser is a node condition reason to indicate that a machine/node has been preserved due to explicit annotation by user
-	PreservedByUser string = "Preserved by user."
+	// PreservationWithoutDrainCompleted is a node condition reason indicating the node has not been drained but is fully preserved. This Reason is used
+	// when machines are preserved in Running, and the node need not be drained.
+	PreservationWithoutDrainCompleted string = "PreservationWithoutDrainCompleted"
 
-	// PreservationStopped is a node condition reason to indicate that a machine/node preservation has been stopped due to annotation update or timeout
-	PreservationStopped string = "Preservation stopped."
+	// PreservationWithDrainCompleted is a node condition reason indicating the node has been drained and is fully preserved.
+	PreservationWithDrainCompleted string = "PreservationWithDrainCompleted"
 
-	// PreservedNodeDrainSuccessful is a constant for the message in condition that indicates that the preserved node's drain is successful
-	PreservedNodeDrainSuccessful string = "Preserved node drained successfully."
-
-	// PreservedNodeDrainUnsuccessful is a constant for the message in condition that indicates that the preserved node's drain was not successful
-	PreservedNodeDrainUnsuccessful string = "Preserved node could not be drained."
+	// DrainFailed is a node condition reason indicating the preserved node could not be drained.
+	DrainFailed string = "DrainFailed"
 )
 
 // CurrentStatus contains information about the current status of Machine.
--- a/pkg/util/nodeops/conditions.go
+++ b/pkg/util/nodeops/conditions.go
@@ -94,6 +94,37 @@ func AddOrUpdateConditionsOnNode(ctx context.Context, c clientset.Interface, nod
 	return updatedNode, err
 }
 
+// RemoveConditionFromNode removes the condition with the given type from the node's status.
+// If the condition is not present, it is a no-op.
+func RemoveConditionFromNode(ctx context.Context, c clientset.Interface, nodeName string, conditionType v1.NodeConditionType) (*v1.Node, error) {
+	firstTry := true
+	var updatedNode *v1.Node
+	err := clientretry.RetryOnConflict(Backoff, func() error {
+		var err error
+		var oldNode *v1.Node
+		if firstTry {
+			oldNode, err = c.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{ResourceVersion: "0"})
+			firstTry = false
+		} else {
+			oldNode, err = c.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
+		}
+		if err != nil {
+			return err
+		}
+		newNode := oldNode.DeepCopy()
+		conditions := make([]v1.NodeCondition, 0, len(newNode.Status.Conditions))
+		for _, cond := range newNode.Status.Conditions {
+			if cond.Type != conditionType {
+				conditions = append(conditions, cond)
+			}
+		}
+		newNode.Status.Conditions = conditions
+		updatedNode, err = UpdateNodeConditions(ctx, c, nodeName, oldNode, newNode)
+		return err
+	})
+	return updatedNode, err
+}
+
 // UpdateNodeConditions is for updating the node conditions from oldNode to the newNode
 // using the node's UpdateStatus() method
 func UpdateNodeConditions(ctx context.Context, c clientset.Interface, nodeName string, oldNode *v1.Node, newNode *v1.Node) (*v1.Node, error) {
--- a/pkg/util/provider/machinecontroller/machine_util.go
+++ b/pkg/util/provider/machinecontroller/machine_util.go
@@ -78,6 +78,13 @@ const (
 	cacheUpdateTimeout = 1 * time.Second
 )
 
+// constants used to build the Preserved node condition message/
+const (
+	preservedByUser             = "Preserved by user."
+	autoPreservedByMCM          = "Auto-preserved by MCM."
+	preserveExpiryMessagePrefix = "Machine preserved until"
+)
+
 // ValidateMachineClass validates the machine class.
 func (c *controller) ValidateMachineClass(_ context.Context, classSpec *v1alpha1.ClassSpec) (*v1alpha1.MachineClass, map[string][]byte, machineutils.RetryPeriod, error) {
 	var (
@@ -2389,30 +2396,41 @@ func (c *controller) preserveMachine(ctx context.Context, machine *v1alpha1.Mach
 
 	nodeName := machine.Labels[v1alpha1.NodeLabelKey]
 	if nodeName == "" {
-		// Machine has no backing node( such as in the case of self-hosted shoots), preservation is complete
+		// If machine has no backing node (or no targetCoreClient), such as in the case of self-hosted shoots,
+		// preservation is complete after setting preserveExpiryTime on the machine.
 		klog.V(2).Infof("Machine %q without backing node is preserved successfully till %v.", machine.Name, machine.Status.CurrentStatus.PreserveExpiryTime)
 		return machine, nil
 	}
 	// Machine has a backing node
 	node, err := c.nodeLister.Get(nodeName)
 	if err != nil {
-		klog.Errorf("error trying to get node %q of machine %q: %v. Retrying.", nodeName, machine.Name, err)
+		klog.Errorf("error trying to get node %q of machine %q: %v.", nodeName, machine.Name, err)
 		return machine, err
 	}
-	existingNodePreservedCondition := nodeops.GetCondition(node, v1alpha1.NodePreserved)
+	nodeClone := node.DeepCopy()
+	existingNodePreservedCondition := nodeops.GetCondition(nodeClone, v1alpha1.NodePreserved)
 	drainRequired := shouldPreservedNodeBeDrained(existingNodePreservedCondition, machine.Status.CurrentStatus.Phase)
-	// For a Running machine, preservation is complete when ConditionStatus is True. However, for a Failed machine,
-	// preservation is complete only once the node is drained and tainted.
-	// Edge-case: when a machine in Running phase is preserved with preserve=now, and
-	// the machine transitions to Failed.
-	// In such cases, even though ConditionStatus would be set to True, on transitioning to
-	// Failed, the preservation needs to be considered as incomplete.
-	if existingNodePreservedCondition != nil && existingNodePreservedCondition.Status == v1.ConditionTrue &&
-		!drainRequired {
+	// For a Running machine, preservation is complete once preserveExpiryTime is set on the machine and
+	// the CA annotations are set on the node. However, for a Failed machine, preservation is complete
+	// only after the node is tainted and drained. When a machine in Running phase is preserved with preserve=now, and
+	// the machine transitions to Failed, the node needs to be drained, even though preservation was complete earlier.
+	if existingNodePreservedCondition != nil &&
+		(existingNodePreservedCondition.Reason == v1alpha1.PreservationWithDrainCompleted ||
+			(existingNodePreservedCondition.Reason == v1alpha1.PreservationWithoutDrainCompleted && !drainRequired)) {
 		return machine, nil
 	}
+
+	if existingNodePreservedCondition == nil {
+		initialCond := getInitializedPreservedNodeCondition(preserveValue, machine.Status.CurrentStatus.PreserveExpiryTime)
+		if _, err = nodeops.AddOrUpdateConditionsOnNode(ctx, c.targetCoreClient, nodeName, initialCond); err != nil {
+			klog.Errorf("error setting initial Preserved node condition on node %q of machine %q: %v", nodeName, machine.Name, err)
+			return machine, err
+		}
+		existingNodePreservedCondition = &initialCond
+	}
+
 	// Step 2: Add annotations to prevent scale down of node by CA
-	updatedNode, err := c.addCAScaleDownDisabledAnnotationOnNode(ctx, node)
+	err = c.addCAScaleDownDisabledAnnotationOnNode(ctx, nodeName)
 	if err != nil {
 		return machine, err
 	}
@@ -2420,20 +2438,28 @@ func (c *controller) preserveMachine(ctx context.Context, machine *v1alpha1.Mach
 	if drainRequired {
 		// Step 3: If machine is in Failed Phase, drain the backing node
 		drainErr = c.drainPreservedNode(ctx, machine)
-	}
-	newCond, needsUpdate := computeNewNodePreservedCondition(machine.Status.CurrentStatus, preserveValue, drainErr, existingNodePreservedCondition)
-	if needsUpdate {
-		// Step 4: Update NodePreserved Condition on Node, with drain status
-		_, err = nodeops.AddOrUpdateConditionsOnNode(ctx, c.targetCoreClient, updatedNode.Name, *newCond)
 		if drainErr != nil {
 			klog.Errorf("error draining preserved node %q for machine %q : %v", nodeName, machine.Name, drainErr)
-			return machine, drainErr
-		}
-		if err != nil {
-			klog.Errorf("error trying to update node preserved condition for node %q of machine %q : %v", nodeName, machine.Name, err)
-			return machine, err
 		}
 	}
+
+	// Step 4: Update Preserved Node Condition with drain status if required
+	newCond := recomputePreservedNodeCondition(machine.Status.CurrentStatus, preserveValue, drainErr, existingNodePreservedCondition)
+	if needsPreservedNodeConditionUpdate(existingNodePreservedCondition, newCond) {
+		_, err = nodeops.AddOrUpdateConditionsOnNode(ctx, c.targetCoreClient, nodeName, *newCond)
+		if err != nil {
+			klog.Errorf("error trying to update node preserved condition for node %q of machine %q : %v", nodeName, machine.Name, err)
+		}
+	}
+
+	if drainErr != nil {
+		return machine, drainErr
+	}
+
+	if err != nil {
+		return machine, err
+	}
+
 	klog.V(2).Infof("Machine %q and backing node %q preserved successfully till %v.", machine.Name, nodeName, machine.Status.CurrentStatus.PreserveExpiryTime)
 	return machine, nil
 }
@@ -2487,15 +2513,8 @@ func (c *controller) stopPreservationIfActive(ctx context.Context, machine *v1al
 		klog.Errorf("error trying to get node %q of machine %q: %v. Retrying.", nodeName, machine.Name, err)
 		return nil, err
 	}
-	// prepare NodeCondition to set preservation as stopped
-	preservedConditionFalse := v1.NodeCondition{
-		Type:               v1alpha1.NodePreserved,
-		Status:             v1.ConditionFalse,
-		LastTransitionTime: metav1.Now(),
-		Reason:             v1alpha1.PreservationStopped,
-	}
-	// Step 1: change node condition to reflect that preservation has stopped
-	updatedNode, err := nodeops.AddOrUpdateConditionsOnNode(ctx, c.targetCoreClient, node.Name, preservedConditionFalse)
+	// Step 1: remove Preserved Node Condition.
+	updatedNode, err := nodeops.RemoveConditionFromNode(ctx, c.targetCoreClient, node.Name, v1alpha1.NodePreserved)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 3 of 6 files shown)


pkg/apis/machine/v1alpha1/machine_types.go
```diff
@@ -244,24 +244,23 @@ const (
...
-	// PreservedByUser is a node condition reason to indicate that a machine/node has been preserved due to explicit annotation by user
-	PreservedByUser string = "Preserved by user."
+	// PreservationWithoutDrainCompleted is a node condition reason indicating the node has not been drained but is fully preserved. This Reason is used
+	// when machines are preserved in Running, and the node need not be drained.
+	PreservationWithoutDrainCompleted string = "PreservationWithoutDrainCompleted"
 
-	// PreservationStopped is a node condition reason to indicate that a machine/node preservation has been stopped due to annotation update or timeout
-	PreservationStopped string = "Preservation stopped."
+	// PreservationWithDrainCompleted is a node condition reason indicating the node has been drained and is fully preserved.
+	PreservationWithDrainCompleted string = "PreservationWithDrainCompleted"
```
elankath, line 260: Change to adjective since condition reasons indicate past tense and use 'and' `PreservedAndDrained`

pkg/apis/machine/v1alpha1/machine_types.go
```diff
@@ -244,24 +244,23 @@ const (
...
+// Constants used by the preservation flow.
 const (
-	// NodePreserved is a node condition type for preservation of machines to allow end-user to know that a node is preserved
+	// NodePreserved is a node condition type that surfaces preservation information to the end-users.
 	NodePreserved corev1.NodeConditionType = "Preserved"
 
-	// PreservedByMCM is a node condition reason for preservation of machines to indicate that the node is auto-preserved by MCM
-	PreservedByMCM string = "Preserved by MCM."
+	// PreservationInProgress is a node condition reason indicating preservation has started but is not yet complete.
+	PreservationInProgress string = "PreservationInProgress"
```
elankath, line 253: Why do we need this  now ? It feels un-necessary with the other reasons.

pkg/apis/machine/v1alpha1/machine_types.go
```diff
@@ -244,24 +244,23 @@ const (
...
 
-	// PreservedByUser is a node condition reason to indicate that a machine/node has been preserved due to explicit annotation by user
-	PreservedByUser string = "Preserved by user."
+	// PreservationWithoutDrainCompleted is a node condition reason indicating the node has not been drained but is fully preserved. This Reason is used
+	// when machines are preserved in Running, and the node need not be drained.
+	PreservationWithoutDrainCompleted string = "PreservationWithoutDrainCompleted"
```
elankath, line 257: Since node need not be drained, this should be reflected in the reason appropriately using an adjective`PreservedWithoutDrain`

pkg/apis/machine/v1alpha1/machine_types.go
```diff
@@ -244,24 +244,23 @@ const (
...
+	// PreservationWithDrainCompleted is a node condition reason indicating the node has been drained and is fully preserved.
+	PreservationWithDrainCompleted string = "PreservationWithDrainCompleted"
 
-	// PreservedNodeDrainSuccessful is a constant for the message in condition that indicates that the preserved node's drain is successful
-	PreservedNodeDrainSuccessful string = "Preserved node drained successfully."
-
-	// PreservedNodeDrainUnsuccessful is a constant for the message in condition that indicates that the preserved node's drain was not successful
-	PreservedNodeDrainUnsuccessful string = "Preserved node could not be drained."
+	// DrainFailed is a node condition reason indicating the preserved node could not be drained.
+	DrainFailed string = "DrainFailed"
```
elankath, line 263: Why miss the `Preserve` prefix ? SHouldn't this be `PreservedDrainFailed` ?

pkg/util/nodeops/conditions.go
```diff
@@ -94,6 +94,37 @@ func AddOrUpdateConditionsOnNode(ctx context.Context, c clientset.Interface, nod
 	return updatedNode, err
 }
 
+// RemoveConditionFromNode removes the condition with the given type from the node's status.
+// If the condition is not present, it is a no-op.
+func RemoveConditionFromNode(ctx context.Context, c clientset.Interface, nodeName string, conditionType v1.NodeConditionType) (*v1.Node, error) {
+	firstTry := true
+	var updatedNode *v1.Node
+	err := clientretry.RetryOnConflict(Backoff, func() error {
```
elankath, line 102: we should avoid use of any new RetryOnConflict's introduced in the MCM codebase

pkg/apis/machine/v1alpha1/machine_types.go
thiyyakat: What do you suggest as a Reason for the initial condition? It is set before drain starts.

pkg/util/nodeops/conditions.go
thiyyakat: Okay. Return an error in the case of conflict then? Or do you recommend doing a fetch from the API server each time?

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2548,56 +2567,63 @@ func (c *controller) setPreserveExpiryTimeOnMachine(ctx context.Context, machine
...
+// In both cases the node needs to be updated with newCond
+func needsPreservedNodeConditionUpdate(oldCond, newCond *v1.NodeCondition) bool {
+	if oldCond == nil || newCond == nil {
+		return true
+	}
+	return oldCond.Status != newCond.Status || oldCond.Reason != newCond.Reason || oldCond.Message != newCond.Message
+}
+
+// getInitializedPreservedNodeCondition returns an initialized Node Condition of Type Preserved
+func getInitializedPreservedNodeCondition(value string, preserveExpiryTime *metav1.Time) v1.NodeCondition {
```
elankath, line 2608: This feels un-necessary as it it anyways overriden immediately in recomputePreservedNodeCondition

pkg/util/provider/machinecontroller/node.go
```diff
@@ -396,23 +398,38 @@ func (c *controller) removePreservationRelatedAnnotationsOnNode(ctx context.Cont
...
-	nodeCopy.Annotations[autoscaler.ClusterAutoscalerScaleDownDisabledAnnotationByMCMKey] = autoscaler.ClusterAutoscalerScaleDownDisabledAnnotationByMCMValue
-	updatedNode, err := c.targetCoreClient.CoreV1().Nodes().Update(ctx, nodeCopy, metav1.UpdateOptions{})
-	if err != nil {
-		klog.Errorf("error trying to update CA annotation on node %q: %v", node.Name, err)
-		return nil, err
-	}
-	return updatedNode, nil
+func (c *controller) addCAScaleDownDisabledAnnotationOnNode(ctx context.Context, nodeName string) error {
+	firstTry := true
+	return clientretry.RetryOnConflict(nodeops.Backoff, func() error {
```
elankath, line 403: We shouldn't add new `RetryOnConflict`s.

elankath (changes requested):

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: This will get overwritten only after drain completes, which can take a significant amount of time. The primary purpose of this PR was to add the node condition before drain starts so the gardenlet health checks exempt these nodes and don't unnecessarily mark the shoot as red, and the end user has an indication that their preservation request is being acted upon. Before this change, the user would receive an indication only once the drain finished.

pkg/apis/machine/v1alpha1/machine_types.go
elankath: I don't believe we need one since we override it anyways. Please correct me if I am wrong but we immediately call recomputePreservedNodeCondition which overwrites the same.

pkg/util/nodeops/conditions.go
elankath: Clarifying the point with team. I prefer just requeing similar to controller runtime but other ppl can have different opinions here.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2630,6 +2656,22 @@ func (c *controller) removePreserveAnnotationOnMachine(ctx context.Context, mach
...
 
+func preservationCause(preserveValue string) string {
+	why := preservedByUser
+	if preserveValue == machineutils.PreserveMachineAnnotationValueAutoPreserved {
+		why = autoPreservedByMCM
+	}
+	return why
+}
+
+func buildPreservedNodeConditionMessage(custom, preserveValue string, preserveExpiryTime *metav1.Time) string {
```
takoverflow, line 2667: Is this really required? Can't the message string be constructed at the call-site itself?

What's the advantage here?

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2389,51 +2396,70 @@ func (c *controller) preserveMachine(ctx context.Context, machine *v1alpha1.Mach
...
-	// Edge-case: when a machine in Running phase is preserved with preserve=now, and
-	// the machine transitions to Failed.
-	// In such cases, even though ConditionStatus would be set to True, on transitioning to
-	// Failed, the preservation needs to be considered as incomplete.
-	if existingNodePreservedCondition != nil && existingNodePreservedCondition.Status == v1.ConditionTrue &&
-		!drainRequired {
+	// For a Running machine, preservation is complete once preserveExpiryTime is set on the machine and
+	// the CA annotations are set on the node. However, for a Failed machine, preservation is complete
+	// only after the node is tainted and drained. When a machine in Running phase is preserved with preserve=now, and
+	// the machine transitions to Failed, the node needs to be drained, even though preservation was complete earlier.
```
takoverflow, line 2416:
```suggestion
	// For a Running machine, preservation is active once preserveExpiryTime is set on the machine and
	// the CA annotations are set on the node. However, for a Failed machine, preservation is active
	// only after the node is tainted and drained. When a machine in Running phase is preserved with preserve=now, and
	// the machine transitions to Failed, the node needs to be drained, even though preservation was active earlier.
```

I don't understand what completion in the context of preservation means, does `active` sound better than `complete`? WDYT

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2389,51 +2396,70 @@ func (c *controller) preserveMachine(ctx context.Context, machine *v1alpha1.Mach
...
 		}
 	}
+
+	if drainErr != nil {
+		return machine, drainErr
+	}
+
+	if err != nil {
+		return machine, err
+	}
```
takoverflow, line 2461:
```suggestion
	if drainErr != nil || err != nil {
		return machine, cmp.Or(drainErr, err)
	}
```

The two error checks can be combined, `cmp.Or` will go through the args and return the first non-zero/nil arg.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2389,51 +2396,70 @@ func (c *controller) preserveMachine(ctx context.Context, machine *v1alpha1.Mach
...
-		// Step 4: Update NodePreserved Condition on Node, with drain status
-		_, err = nodeops.AddOrUpdateConditionsOnNode(ctx, c.targetCoreClient, updatedNode.Name, *newCond)
 		if drainErr != nil {
 			klog.Errorf("error draining preserved node %q for machine %q : %v", nodeName, machine.Name, drainErr)
-			return machine, drainErr
 		}
+	}
+
+	// Step 4: Update Preserved Node Condition with drain status if required
+	newCond := recomputePreservedNodeCondition(machine.Status.CurrentStatus, preserveValue, drainErr, existingNodePreservedCondition)
```
takoverflow, line 2447: Just thinking out loud

Wouldn't it be beneficial to have a generic condition on the node that states that `DrainSuccessful`/`DrainComplete` or `DrainInProgress` as part of node drain. 

Then the preservation with/without drain logic can depend on that, it will also enhance the observability of nodes in general. Rather than having drain related conditions only active for preservation, WDYT?

takoverflow (commented):
Thanks for the changes, have added some queries/suggestions PTAL

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: This is used in multiple places. If I wish to change the format or contents of the message, I can do it in one place.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: That doesn't seem right. Preservation is _active_ once PreserveExpiryTime is set, but can be said to be _complete_ only after all the tasks such as node condition update, CA annotation update and drain are completed.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Oh neat. Thanks. Will make the change 👍

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: That's a good idea. But it should probably go into @r4mek's PR on drain and not in this. Along with those changes, we can change preservation logic to depend on the drain node condition instead.

pkg/util/provider/machinecontroller/machine_util.go
takoverflow: But I'm a bit confused, aren't the contents always determined by the arguments here. The function isn't adding anything additional right?

pkg/util/provider/machinecontroller/machine_util.go
takoverflow: So `Active` => `ExpiryTime` set
And `Complete` => `Active` and the `Preservation` node conditions (or anything else) are added?

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: > Is this really required? Can't the message string be constructed at the call-site itself?

It serves the purpose of reducing code duplication, and adding a layer of abstraction that improves readability. It also reduces the complexity in the caller. And there is no performance issue since the code will be inlined by the go compiler if it is small enough. 

What is the down side?

I have tried to use helpers whenever possible based on the review comments I received from @elankath in an earlier PR. Ref: https://github.com/gardener/machine-controller-manager/pull/1118#discussion_r3638940685, https://github.com/gardener/machine-controller-manager/pull/1118#discussion_r3639266523.

 If you see the comments, the recommendations he has made are for far simpler code blocks, but converting to a helper improves readability and testability.
 
 Will add tests for this function if you are convinced. :) 
 
 

> But I'm a bit confused, aren't the contents always determined by the arguments here. The function isn't adding anything additional right?

Of course.  We may want to introduce/remove some information in the future. In which case we would have to make the changes in multiple different places.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Yes!

pkg/util/provider/machinecontroller/machine_util.go
takoverflow: > What is the down side?

Needing to anyways having to check the definition at each occurrence to see how the description is being computed, rather than seeing an `Sprintf`. But it's fine I guess.

> We may want to introduce/remove some information in the future. In which case we would have to make the changes in multiple different places.

Okay.

pkg/util/provider/machinecontroller/machine_util.go
aaronfern: My reservations:
It might be a stretch to correlate a standalone generic drain condition with another condition that talks about preservation, so I'm not sure it would really help is observability any more than adding the drain status to the preserved node condition.
Additionally we would also now need logic somewhere to eventually remove this drain condition, else it will get confusing when the node finally recovers.

Imo, yes, it might help simplify preservation logic, but if not done well can impact observability

takoverflow: > Additionally we would also now need logic somewhere to eventually remove this drain condition, else it will get confusing when the node finally recovers.

I was thinking of a node condition that's only true when drain is in progress i.e. `DrainInProgress` with the state `true` only during `RunDrain` execution. Once drain finishes, it can either be removed or the state can transition to `false`.

The visibility I was referring to was **_also_** from an end-user PoV, regardless of drain being because of preservation or not. Where now they can easily look at a node to tell when it's being drained.

The other upside being that the preservation logic can also piggy-back on that.

> It might be a stretch to correlate a standalone generic drain condition with another condition that talks about preservation

Isn't that what's indirectly happening right now? :sweat_smile: From https://github.com/gardener/machine-controller-manager/pull/1136#issue-5244282133

> The Preserved node condition is now set to PreservationInProgress before draining begins, so users can observe that MCM has acknowledged the preservation annotation and is acting on it. This also allows gardenlet health checks to exempt Failed preserved machines sooner thereby reducing alerting noise.

<hr>

You're right, depends on how much work it ends up being, if it's a lot then can be deferred for later.

aaronfern: > I was thinking of a node condition that's only true when drain is in progress i.e. DrainInProgress with the state true only during RunDrain execution. Once drain finishes, it can either be removed or the state can transition to false.

Okay, fair. Earlier when you spoke about the drain condition having "`DrainSuccessful`/`DrainComplete`", it wasn't clear to me

> > It might be a stretch to correlate a standalone generic drain condition with another condition that talks about preservation
>
>Isn't that what's indirectly happening right now? 😅 From https://github.com/gardener/machine-controller-manager/pull/1136#issue-5244282133

Um, yes and no. Whatever happens does happen within the same condition so hard to miss and iiuc is set specifically for draining

---

On second thought though. A node condition for drain would be nice. Seems more idiomatic. This is of course provided that after drain is completed, MCM removes it as preservation progresses.

---

@thiyyakat 
Please correct me if I'm wrong, I did not thoroughly review the code.
Is the `PreservationInProgress` node condition reason only intended to be used when the node is draining? if yes then why not explicitly state that drain is in progress?

thiyyakat: Just one thing, we cannot remove the drain condition on drain completion if we wish to use it for the preservation flow. We wouldn't know if drain has completed or if it was never attempted. So the condition would have to persist beyond drain completion which complicates things as @aaronfern pointed out.

<hr>

> The visibility I was referring to was also from an end-user PoV

Is this required today though? Have you felt during your DoD shifts or RCA that you would have benefited from this being surfaced? Then we should consider it. 

<hr>

I didn't quite get how the drain node condition will simplify preservation logic. In fact, we would need to fetch this condition and based on the condition do nearly the same branching. We may reduce the reasons associated with the preserved node condition ( by one) , but we will still need to introduce new reasons in the new node condition.

So I don't see any reason for the dependency between the conditions. If we see value in the drain node condition, it should be introduced. But I don't know if it should be for the sake of simplifying existing code. If we introduce this condition, it's not like we will change the InPlace flow or triggerDeletionFlow to rely on the node condition, right?

<hr>

> Isn't that what's indirectly happening right now? 😅 From https://github.com/gardener/machine-controller-manager/pull/1136#issue-5244282133

I don't understand what you mean here. How are we relying on another condition indirectly?


<hr>

> You're right, depends on how much work it ends up being, if it's a lot then can be deferred for later.

That depends on whether RunDrain is the only function that needs to be modified to do this. Moreover, if @r4mek is changing RunDrain, then there is no point doing it now. Maybe we can take a call based on that.

thiyyakat: > Is the PreservationInProgress node condition reason only intended to be used when the node is draining? if yes then why not explicitly state that drain is in progress?

@aaronfern 

Good point. It is intended to be used even if there is no drain required. So maybe I can just set the condition reason to PreservedWithoutDrain directly, without initializing it, when drain is not required. When drain is required, I can set it to DrainInProgress before drain starts. We can do away with PreservationInProgress.

takoverflow: > So the condition would have to persist beyond drain completion which complicates things

I'm sorry, what complications arise from persisting a `DrainInProgress: false` condition? Can you elaborate I'm not clear as to where the problem lies.

> but we will still need to introduce new reasons in the new node condition.

What reasons are required? If its drain due to preservation, there are two conditions `Preservation` and `DrainInProgress` indicating that (with the latter not really needing any reason). If it's drain due to termination, the node only has one condition.

> I don't understand what you mean here. How are we relying on another condition indirectly?

So IIUC, the entire reason for existence of `PreservationInProgress` as per https://github.com/gardener/machine-controller-manager/pull/1136#discussion_r3852574221 is for the gardenlet to know that preservation is active and for user to be aware that because of node failure, drain is triggered.

Now, based on what kind of preservation annotation was added, what was the phase of the machine when preservation activated, the `NodePreserved` condition is being modified. This should IMO be independent of whether drain happened or not. Once `preserveExpiryTime` is set, the Node is essentially preserved, no matter what the condition `Reason` is, I would expect the condition to be added and not used as a state. 

To elaborate a bit, the below checks would then be done based on the `Drain` condition, and not the `NodePreserved` condition `Reason`.
```go
	if existingNodePreservedCondition != nil &&
		(existingNodePreservedCondition.Reason == v1alpha1.PreservationWithDrainCompleted ||
			(existingNodePreservedCondition.Reason == v1alpha1.PreservationWithoutDrainCompleted && !drainRequired)) {
		return machine, nil
	}
	...
	// Step 4: Update Preserved Node Condition with drain status if required
	newCond := recomputePreservedNodeCondition(machine.Status.CurrentStatus, preserveValue, drainErr, existingNodePreservedCondition)
	if needsPreservedNodeConditionUpdate(existingNodePreservedCondition, newCond) {
		_, err = nodeops.AddOrUpdateConditionsOnNode(ctx, c.targetCoreClient, nodeName, *newCond)
		if err != nil {
			klog.Errorf("error trying to update node preserved condition for node %q of machine %q : %v", nodeName, machine.Name, err)
		}
	}
```
Also `Step 4` is not needed then as well.

Why is this condition being used as a preservation drain indicator? This muddling of concerns is something which I disliked.

That's why I proposed a separate indicator for any actor, be it MCM or gardenlet or the end-user to be aware when there's drain activity on a node.

If you still feel this is out of the scope of the `Preserved` node condition change, then it can be left for later whenever more `Drain` related changes are done. The only concern is, once a behavior is introduced there's a tendency to not change things a lot for fear of additional things breaking but it's fine.

thiyyakat: > what complications arise from persisting a DrainInProgress: false condition? Can you elaborate I'm not clear as to where the problem lies.
 
Say a machine has been preserved in Failed with `preserve=when-failed`, the node was drained successfully, then from what I understand, you wish to set node condition DrainInProgress to False. These are the extra things that need to be taken care of:
1. DrainInProgress being set to False implies drain has completed, not that it has never commenced. This would be a behaviour specific to this particular condition only and it will be a cognitive overhead since it cannot be inferred from the condition itself.
2. If say you accept the above semantics, then what would happen if the node got back to Ready, machine went to Running and the node failed again? You cannot interpret DrainInProgress at this point as an indicator for drain having completed.

I may not have fully understood your proposal, I apologise if the questions are silly 😅 

> What reasons are required?

Would you not want the new node condition to have any Reasons to enable branching on the state of drain? What is the use of the node condition for drain then? Just to indicate to the enduser that drain is ongoing? It also gives us a chance to branch in existing flows in a more kubernetes idiomatic way, right?

takoverflow: Out of curiosity, I checked if `cluster-api` has some indicator as well for draining and stumbled across this [cluster-api/drain#observability](https://cluster-api.sigs.k8s.io/tasks/automated-machine-management/machine_deletions.html?highlight=drain#observability)

> The drain process can be observed through the `DrainingSucceeded` condition on the Machine and various logs.

<hr>

Regarding the questions:

> DrainInProgress being set to False implies drain has completed, not that it has never commenced.

> what would happen if the node got back to Ready, machine went to Running and the node failed again? You cannot interpret DrainInProgress at this point as an indicator for drain having completed.

1. If the node recovers, the condition can be removed as @aaronfern pointed out.
2. If the node never undergoes drain, the condition would not be present on it.
3. When the drain commences, its added with the state `True` and transitions to `False` upon completion.

As for how the preservation condition uses it, there's no need for it to be concerned about the `Drain` condition right? Once the `PreserveExpiryTime` is set, then the `NodePreserved` condition is also added. Now there can be two cases:
1. `when-failed` annotation: This implies that the condition is added only when the machine goes to failed state. The `NodePreserved` condition is added, `drainPreservation` is called and the `Drain` condition is added on the node as well.
2. `now` annotation: The `NodePreserved` condition is added and the `Drain` condition is only added when the machine actually goes to the `Failed` phase and the node drain is attempted.

In case of recovery post drain, the `NodePreserved` condition is still present since `PreserveExpiryTime` isn't breached but the `Drain` condition is removed since the node is no longer unschedulable.

This keeps the entire drain process and its progress independent of the `NodePreservation` conditions.

> Just to indicate to the enduser that drain is ongoing?

Isn't that what the different reasons for the `NodePreserved` condition are achieving right now? Ideally, the condition should just be set to `true` once with the reason maybe stating why its preserved rather than saying `PreservedAndDrained`/`PreservedWithoutDrain` or `PreservedDrainFailed`.

The latter (drain) has nothing to do with Preservation.

> This would be a behaviour specific to this particular condition only and it will be a cognitive overhead since it cannot be inferred from the condition itself.

Could you explain what cognitive overhead are you talking about and what do you wish to infer?

> I apologise if the questions are silly

They aren't, it allows for clarification of the proposal, please don't hesitate to ask.

thiyyakat:
/hold

The new node condition for drain needs to be introduced, and the code refactored.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Hi @takoverflow. I would like to use this thread to continue discussion on the topic though the PR is in draft at the moment.

A summary of what I've understood to be the expected behaviour, please correct me if I've got something wrong:

   - A new node drain condition should exist for **all** drain scenarios (preservation, in-place and regular termination). This condition tracks the drain operation's state independently of the flow that triggered it.
   
  - Preservation flow now becomes:
	  - Running + preserve=now: Don't drain initially
	  - Failed + preserve=now/when-failed/auto-preserved: Check drain condition:
	    - Not present: Perform drain (call `drainPreservedNode`)
	    - Present + unsuccessful: Retry drain
	    - Present + successful: Skip drain, proceed
	  - Remove drain and preserved node conditions when stopping preservation (since the machine isn't necessarily terminating so the cleanup has to be done)
	  
 - Regular termination:
	- Leave drain condition on the node since the machine is terminating anyway
	
- In-place update:
	- Remove the node condition once drain is successful or towards the end of machineutils.inPlaceUpdate(), since this should not be carried forward across the update. 

This removal will differ across different flows – that's what I was referring to when I said there may be a cognitive burden on the developer, but its acceptable. We're doing drain in different phases already, this just becomes one other thing that is not uniform.

- As is the norm, the drain condition is not to be used by drain logic itself for branching. It is to be used for other flows that depend on the drain status and for observability.
 - **For in-place updates:** The existing `v1alpha1.DrainSuccessful` reason is used by the in-place logic to branch. If we're introducing a new drain condition, should in-place:
  - Option A: Keep using its existing reason in the in-place condition (status quo)
  - Option B: Align with the new drain condition for consistency
    If we decide to go with Option A, we reduce chances of causing issues in the in-place flow. In that case though, we might as well continue using the preserved node condition as it is today if we are okay with the overlap in in-place.
    
    
  Please LMK if I've missed anything or if you can think of any edge-cases that I've not accounted for. I will create an issue once you feel the points above are okay.

PR #1138, opened 2026-08-25
Fix: Align auto-preservation eviction order with scale-down deletion priority

aaronfern:
We noticed an inconsistency between how auto-preserved failed machines are chosen for termination due to `mcs`'s `spec.replica` reduction and how auto-preserve failed machines are chosen for termination due to `mcs`'s `spec.autoPreserveFailedMachineMax` reduction.

Currently when `spec.autoPreserveFailedMachineMax` is reduced, auto-preserved failed machines that have a nearer `PreserveExpiryTime` are chosen for deletion, whereas when `spec.replica` is reduced auto-preserved failed machines that have an earlier `CreationTimestamp` are chosen for deletion.

This PR aligns both these orderings to ensure that now auto-preserved failed machines are only ordered based on their `PreserveExpiryTime`s

Fixes #

**Release note**:

```bugfix operator
Ensure auto-preserve failed machines are ordered by their `PreserveExpiryTime` in case of deletion due to reduction is mcs.spec.Replica or mcs.spec.autoPreserveFailedMachineMax
```

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -361,11 +361,13 @@ The following can be the reason:
 
 ### Why are some preserved failed machines deleted even though they haven't passed their preserve expiry time yet?
 
-This can happen if either of the following values have been reduced
-1. `spec.autoPreserveFailedMachineMax`
-2. `spec.Replicas`
+The most likely reason would be due to a rolling or inPlace upgrade that happened, since preservation is not honoured during upgrades.
 
-When `spec.autoPreserveFailedMachineMax` is reduced and there are more auto-preserved machines than the max count, MCM will sort the machines according to their `PreserveExpiryTime`s and remove machines that have the nearest expiry time. This is to done to keep machines who have the potential for longer preservation and hence additional time for debugging.
+This can also happen if either of the following values have been reduced
+1. `machineDeployment.spec.autoPreserveFailedMachineMax`
+2. `machineDeployment.spec.Replicas`
+
+When `spec.autoPreserveFailedMachineMax` is reduced and there are more auto-preserved machines than the max count, MCM will sort the machines according to their `PreserveExpiryTime`s and remove machines that have the nearest expiry time. This is done to keep machines who have the potential for longer preservation and hence additional time for debugging.
 In the case where `spec.Replicas` has been reduced, MCM will always prioritize keeping preserved machines. If this is not possible due to all non-preserved machines being deleted already, then MCM uses the same mechanism as above. Additionally, MCM will also prioritize keeping manually preserved failed machines over auto-preserved failed machines. This might lead to a case where MCM deleted all auto preserved failed machines if `spec.Replicas` is reduced accordingly.
 
 # Developer
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -9,7 +9,7 @@ persona: Operators
 
 The answers in this FAQ apply to the newest (HEAD) version of Machine Controller Manager. If
 you're using an older version of MCM please refer to corresponding version of
-this document. Few of the answers assume that the MCM being used is in conjuction with [cluster-autoscaler](https://github.com/gardener/autoscaler):
+this document. Few of the answers assume that the MCM being used is in conjunction with [cluster-autoscaler](https://github.com/gardener/autoscaler):
 
 # Table of Contents:
 <!--- TOC BEGIN -->
@@ -48,6 +48,7 @@ this document. Few of the answers assume that the MCM being used is in conjuctio
     - [My machine is stuck in deletion for 1 hr, why?](#my-machine-is-stuck-in-deletion-for-1-hr-why)
     - [My machine is not joining the cluster, why?](#my-machine-is-not-joining-the-cluster-why)
     - [My rolling update is stuck, why?](#my-rolling-update-is-stuck-why)
+    - [Why are some preserved failed machines deleted even though they haven't passed their preserve expiry time yet?](#why-are-some-preserved-failed-machines-deleted-even-though-they-havent-passed-their-preserve-expiry-time-yet)
 - [Developer](#developer)
     - [How should I test my code before submitting a PR?](#how-should-i-test-my-code-before-submitting-a-pr)
     - [I need to change the APIs, what are the recommended steps?](#i-need-to-change-the-apis-what-are-the-recommended-steps)
@@ -239,7 +240,7 @@ It is recommended to only set `MachineDrainTimeout`. It satisfies the related re
   - MCM auto calculates the `maxEvictRetries` based on the `drainTimeout`.
 - If `drainTimeout` isn't set and only `maxEvictRetries` is set:
   - Default `drainTimeout` and user provided `maxEvictRetries` for each pod is considered.
-- If both `maxEvictRetries` and `drainTimoeut` are set:
+- If both `maxEvictRetries` and `drainTimeout` are set:
   - Then both will be respected.
 - If none are set:
   - Defaults are respected.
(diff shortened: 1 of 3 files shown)


docs/faq.md
```diff
@@ -358,6 +359,15 @@ The following can be the reason:
 - [Old machines are stuck in deletion](#my-machine-is-stuck-in-deletion-for-1-hr-why)
 - If you are using Gardener for setting up kubernetes cluster, then machine object won't turn to `Running` state until `node-critical-components` are ready. Refer [this](https://github.com/gardener/gardener/blob/master/docs/usage/advanced/node-readiness.md) for more details.
 
+### Why are some preserved failed machines deleted even though they haven't passed their preserve expiry time yet?
+
+This can happen if either of the following values have been reduced
+1. `spec.autoPreserveFailedMachineMax`
+2. `spec.Replicas`
+
+When `spec.autoPreserveFailedMachineMax` is reduced and there are more auto-preserved machines than the max count, MCM will sort the machines according to their `PreserveExpiryTime`s and remove machines that have the nearest expiry time. This is to done to keep machines who have the potential for longer preservation and hence additional time for debugging.
```
thiyyakat:
```suggestion
When `spec.autoPreserveFailedMachineMax` is reduced and there are more auto-preserved machines than the max count, MCM will sort the machines according to their `PreserveExpiryTime`s and remove machines that have the nearest expiry time. This is done to keep machines who have the potential for longer preservation and hence additional time for debugging.
```

docs/faq.md
```diff
@@ -358,6 +359,15 @@ The following can be the reason:
 - [Old machines are stuck in deletion](#my-machine-is-stuck-in-deletion-for-1-hr-why)
 - If you are using Gardener for setting up kubernetes cluster, then machine object won't turn to `Running` state until `node-critical-components` are ready. Refer [this](https://github.com/gardener/gardener/blob/master/docs/usage/advanced/node-readiness.md) for more details.
 
+### Why are some preserved failed machines deleted even though they haven't passed their preserve expiry time yet?
+
+This can happen if either of the following values have been reduced
+1. `spec.autoPreserveFailedMachineMax`
```
thiyyakat: Do we need to mention which object's Spec we are referring to here?

docs/faq.md
```diff
@@ -358,6 +359,15 @@ The following can be the reason:
 - [Old machines are stuck in deletion](#my-machine-is-stuck-in-deletion-for-1-hr-why)
 - If you are using Gardener for setting up kubernetes cluster, then machine object won't turn to `Running` state until `node-critical-components` are ready. Refer [this](https://github.com/gardener/gardener/blob/master/docs/usage/advanced/node-readiness.md) for more details.
 
+### Why are some preserved failed machines deleted even though they haven't passed their preserve expiry time yet?
```
thiyyakat, line 362: One more reason could be "due to an ongoing rolling or inplace update, since preservation is not honoured during updates."

thiyyakat (commented):
Thanks for the quick fix, @aaronfern, and for updating the FAQ! Just a few nits from me.

Also, if it's not too much trouble, could you please add a similar test for the sorting performed in the case of mcd replica reduction as well?

aaronfern:
Thanks for reviewing @thiyyakat 
I've addressed your comments. PTAL

pkg/controller/machineset_test.go
```diff
@@ -1623,6 +1623,26 @@ var _ = Describe("machineset", func() {
 			// expect machinesToDelete to contain testPreservedFailedMachine
 			Expect(machinesToDelete).To(ContainElement(testPreservedFailedMachine))
 		})
+		It("should remove auto-preservation from machine with earlier PreserveExpiryTime first, not oldest by creation time", func() {
```
thiyyakat: Thanks for adding the test! The description in `It()` may need to be modified. Since this test is not specific to auto-preservation.

aaronfern: This specific one is related to preservation though. The machines used in the test have preservation related fields and we're trying to test sorting wrt preserved machines

thiyyakat: What I meant was, maybe the description should say "should prioritize preserved machine with earlier PreserveExpiryTime first, not oldest by creation time". Since the function being tested does not have logic to stop preservation, it only returns machines to be deleted.

aaronfern: Understood. Updated, ptal

thiyyakat:
Thanks! The changes look good to me. Just wanted to add a comment here for completeness:

When autoPreserveFailedMachineMax is reduced, we stop auto-preservation based solely on PreserveExpiryTime (removing the machine whose preservation expires soonest first), unlike the more general ActiveMachine sort. There's no value in considering other criteria such as MachinePriority here because when PreserveExpiryTime values differ by only milliseconds, and we choose the machine with the earlier PreserveExpiryTime the choice is effectively arbitrary. 
More criteria would imply a meaningful preference when PreserveExpiryTime values are equal, where none really exists. 😅 

/lgtm

pkg/controller/machineset.go
```diff
@@ -988,7 +988,7 @@ func (c *controller) stopAutoPreservationForMachines(ctx context.Context, machin
 	}
 	if numOfAutoPreservedMachines > numToStop {
 		sort.Slice(autoPreservedMachines, func(i, j int) bool {
-			return autoPreservedMachines[i].CreationTimestamp.Before(&autoPreservedMachines[j].CreationTimestamp)
+			return autoPreservedMachines[i].Status.CurrentStatus.PreserveExpiryTime.Before(autoPreservedMachines[j].Status.CurrentStatus.PreserveExpiryTime)
```
gagan16k, line 991: docstring on `manageAutoPreservationOfFailedMachines()` might need to be updated

Also, we could potentially reuse the 'ActiveMachines' type we already have to sort by these conditions, so that the logic can stay in one place.

pkg/controller/machineset_test.go
```diff
@@ -1623,6 +1623,26 @@ var _ = Describe("machineset", func() {
 			// expect machinesToDelete to contain testPreservedFailedMachine
 			Expect(machinesToDelete).To(ContainElement(testPreservedFailedMachine))
 		})
+		It("should prioritize preserved machine with earlier PreserveExpiryTime first, not oldest by creation time", func() {
```
gagan16k, line 1626:
```suggestion
		It("should prioritize preserved machine with earlier PreserveExpiryTime first for deletion", func() {
```

pkg/controller/machineset_test.go
```diff
@@ -2254,6 +2274,83 @@ var _ = Describe("machineset", func() {
 				},
 			}),
 		)
+
+		It("should remove auto-preservation from machine with earlier PreserveExpiryTime first, not oldest by creation time", func() {
```
gagan16k, line 2278:
```suggestion
		It("should remove auto-preservation from machine with earlier PreserveExpiryTime first", func() {
```

pkg/controller/machineset.go
thiyyakat: While I said this earlier: [comment](https://github.com/gardener/machine-controller-manager/pull/1138#issuecomment-5493213641), @gagan16k's  suggestion sounds good to me, not from the value it adds to the sorting itself, but from the POV of preventing errors. In the original PR sorting as ActiveMachines was considered an overkill because we only sort auto-preserved machines here. But seeing how its already led to an inconsistency, this may be a good idea. Moreover, once the maintenance/reboot stories come in, this sorting is likely to further change . If we have multiple sorting logics we have to take extra care to ensure semantic consistency.

pkg/controller/machineset_test.go
```diff
@@ -1623,6 +1623,26 @@ var _ = Describe("machineset", func() {
 			// expect machinesToDelete to contain testPreservedFailedMachine
 			Expect(machinesToDelete).To(ContainElement(testPreservedFailedMachine))
 		})
+		It("should prioritize preserved machine with earlier PreserveExpiryTime first for deletion", func() {
```
gagan16k, line 1626: Do we need this test here? Both the tests added by the PR do the same thing.

thiyyakat: The second test checks if the auto-preservation annotation is removed as well. Also, different functions are being unit-tested  in the two test cases😅

gagan16k: Makes sense

thiyyakat (approved):

PR #1142, opened 2026-08-27
Fix for Machine (sometimes) getting stuck with empty phase when finalizer addition exhausts retry budget

### LINKED ISSUE #1141: Machine can get permanently stuck with empty phase when finalizer addition exhausts retry budget
## Problem

A new `Machine` object can (occasionally) get  permanently stuck with an empty phase and never progress to `Pending`.

## Root Cause

`addMachineFinalizers` has two behaviours that can perform badly under pressure:

1. On **success**, it deliberately returns a non-nil error (`"Machine finalizers are UPDATED"`) to force a re-queue. This consumes one retry slot.
2. On a **409 conflict** (informer cache not yet reflecting the write), it returns the conflict error. This also consumes a retry slot.

Under load, the informer cache can lag long enough that every retry fires against a stale object, producing repeated 409s. Once the worker exhausts `DefaultMaxRetries = 15`, it calls `queue.Forget(key)` and permanently drops the machine.

Main problem is that there is unfortunately no recovery path: the `updateMachine` informer handler skips re-enqueuing when `Generation` is unchanged, and adding a finalizer only increments `resourceVersion`, not `Generation`. The machine is never reconciled again.

## Observed Symptom

```
machine_util.go:1204  Added finalizer to machine "…" …          <- success
machine_util.go:1201  Failed to add finalizers for machine "…"  <- 409 conflict on next retry
machine.go:194        reconcileClusterMachine: Start … phase:""
```
After the retry budget is exhausted, `reconcileClusterMachine` is never called again and the machine stays with `phase: ""` indefinitely.

elankath:

Fixes #1141

**Release note**:

```bugfix operator
Fix for Machine getting stuck with empty phase when finalizer addition exhausts retry budget
```


elankath:
NOTE: We have a flaky test "set LDRCBST annotation on the machineSet and TriggerDeletionByMCM annotation is not set on the machineSet" which is timing dependent and tends to fail. The fix for this test should be addressed in some other PR.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1197,6 +1197,12 @@ func (c *controller) addMachineFinalizers(ctx context.Context, machine *v1alpha1
 		clone.Finalizers = finalizers.List()
 		_, err := c.controlMachineClient.Machines(clone.Namespace).Update(ctx, clone, metav1.UpdateOptions{})
 		if err != nil {
+			if apierrors.IsConflict(err) {
+				// Informer cache is stale; the watch event will re-enqueue the machine once the
+				// cache catches up, so this is not a retriable error from the worker's perspective.
+				// Returning nil avoids burning a retry slot from the fixed budget (DefaultMaxRetries).
+				return machineutils.ShortRetry, nil
```
aaronfern, line 1204: There is a `ConflictRetry` timeout defined in utils. Would you instead want to use that is the 5s `ShortRetry` enough?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -76,7 +76,13 @@ func (c *controller) updateMachine(oldObj, newObj any) {
 	}
 
 	if oldMachine.Generation == newMachine.Generation {
-		klog.V(3).Infof("Skipping other non-spec updates for machine %q", oldMachine.Name)
+		// Finalizer changes increment resourceVersion but not Generation, so a machine that just
+		// had the MCM finalizer added would be silently dropped here. Re-enqueue it so
+		// reconcileClusterMachine is reached and the machine advances past the empty phase.
+		if !sets.NewString(oldMachine.Finalizers...).HasAll(newMachine.Finalizers...) {
+			klog.V(4).Infof("updateMachine: machine %q gained finalizer — re-enqueuing despite unchanged Generation", newMachine.Name)
```
aaronfern, line 83: Would come handy when debugging
```suggestion
			klog.V(3).Infof("updateMachine: machine %q gained finalizer — re-enqueuing despite unchanged Generation", newMachine.Name)
```

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: I don't think it matters, the caller discards the retry period. Can we change the signature to return just an error then? The function is only used in one place which discards the retry period returned.

pkg/util/provider/machinecontroller/machine.go
r4mek, line 176: I was wondering if we can just requeue the machine here like:

```go
retryPeriod, err := c.addMachineFinalizers(ctx, machine)
if err != nil {
    c.enqueueMachineAfter(machine, time.Duration(retryPeriod), "failed to add MachineFinalizer")
	return nil
}
```

instead of tracking the update event  and requeuing in `updateMachine()`.
And I am not sure if we should proceed if the finalizer is not added on the machine (since we return `nil` for `ConflictError` in `c.addMachineFinalizers()`).

r4mek: BTW I was wondering if `MachineFinalizers` should be added by machine controller or machineset controller. And if machine controller does add it, should we be requequing the machine or continue?

pkg/util/provider/machinecontroller/machine_util.go
elankath: @thiyyakat  I will follow @aaronfern  and @r4mek  suggestion of reque with the `retryPeriod` and then returning `nil` from caller.

pkg/util/provider/machinecontroller/machine.go
elankath: @r4mek It is OK for `MachineFinalizers` to be added by `MachineController` if they are missing.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -77,7 +76,13 @@ func (c *controller) updateMachine(oldObj, newObj any) {
 	}
 
 	if oldMachine.Generation == newMachine.Generation {
```
thiyyakat, line 79: Shouldn't we have this log outside the inner `if`? It helps us know that `updateMachine` was called for the machine and the machine was not reconciled because the change was not significant from MCM's POV.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -77,7 +76,13 @@ func (c *controller) updateMachine(oldObj, newObj any) {
...
-		klog.V(3).Infof("Skipping other non-spec updates for machine %q", oldMachine.Name)
+		// Finalizer changes increment resourceVersion but not Generation, so a machine that just
+		// had the MCM finalizer added would be silently dropped here. Re-enqueue it so
+		// reconcileClusterMachine is reached and the machine advances past the empty phase.
+		if !sets.NewString(oldMachine.Finalizers...).HasAll(newMachine.Finalizers...) {
+			klog.V(3).Infof("updateMachine: machine %q gained finalizer — re-enqueuing despite unchanged Generation", newMachine.Name)
+			c.enqueueMachine(newObj, "handling machine finalizer UPDATE event")
```
r4mek: Do we need to enqueue the machine when the finalizer is added? Because right now when we add the finalizer in `reconcileClusterMachineKey()` we continue performing other steps so not sure how requeue helps here. And as mentioned in my other comment, depending on how we solve it, we might not need this requeue. WDYT?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -209,9 +173,10 @@ func (c *controller) reconcileClusterMachineKey(key string) error {
 	}
 
 	// Add finalizers if not present on machine object
-	_, err = c.addMachineFinalizers(ctx, machine)
+	retryPeriod, err := c.addMachineFinalizers(ctx, machine)
```
r4mek, line 178: Right now after adding the finalizer, we are still working with the outdated machine object which doesn't have the finalizer set. We have two options:
- overwrite the machine object so that it has the updated machine.
- requeue the machine after the finalizer is added on the machine so that in the next reconciliation it has the updated machine.

pkg/util/provider/machinecontroller/machine.go
r4mek, line 84: Do we need to enqueue the machine when the finalizer is added? Because right now when we add the finalizer in `reconcileClusterMachineKey()` we continue performing other steps so not sure how requeue helps here.

And as mentioned in my other comment, depending on how we solve it, we might not need this requeue. WDYT?

elankath: The existing code was re-queing the Machine when the finalizer was added to avoid using an old Machine object. Not in favor of changing existing code to adopt general purpose improvement for a bug-fix PR.

pkg/util/provider/machinecontroller/machine.go
elankath: I don't get it - we should only emit the log if the finalizers don't match indicating the necessity for the re-queue. Otherwise if old Machine generation is same as new machine generation and there is no change (including any finalizer), then we return without any log.

pkg/util/provider/machinecontroller/machine.go
elankath: I am not in favour of adding a third return parameter - the `Machine` object for this one method. The thing is that your suggestion is valid, but we should restrict the scope to fixing the bug and not general improvement across MCM code base to always take new Machine objects after update and proceeding with reconciliation instead of re-queueing.  The latter can be done as part of porting the MCM V2 or a separate refactoring feature PR.

pkg/util/provider/machinecontroller/machine.go
thiyyakat: I think it will be helpful to have this log to know that the watches are not stale ( seeing the use for this in an issue we are debugging at the moment). Knowing that MCM picked up an update event and chose not to act on it will help us distinguish between MCM not seeing the update versus not acting on it.

pkg/util/provider/machinecontroller/machine.go
takoverflow: > we are still working with the outdated machine object

We aren't, since when the finalizer addition is successful we return a dummy error that requeues the machine and exits the current reconcilition.

takoverflow: To avoid this and to not have a third parameter, one can do this if the finalizer update is successfully applied, since `addMachineFinalizers` takes `*machine` as the parameter
```
		machine = clone
```
And the fake error can be removed, but I leave it upto @elankath, there's no harm in retrying 5s later as well.

pkg/util/provider/machinecontroller/machine.go
takoverflow: ~~But this will lead to a lot of spam @thiyyakat 
Since every status update for every machine would result in a log being printed!~~

Nvm, misunderstood. We're already doing it.

pkg/util/provider/machinecontroller/machine.go
r4mek: My bad, missed that err being sent.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -167,9 +175,10 @@ func (c *controller) reconcileClusterMachineKey(key string) error {
...
 	if err != nil {
-		return err
+		c.enqueueMachineAfter(machine, time.Duration(retryPeriod), err.Error())
+		return nil
```
gagan16k, line 181: Should we be doing this for every error? Maybe we can look for `409` here instead?

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1190,25 +1190,25 @@ func getFormattedNodeConditions(conditions []v1.NodeCondition) string {
...
 		if err != nil {
-			// Keep retrying until update goes through
 			klog.Errorf("Failed to add finalizers for machine %q: %s", machine.Name, err)
+			if apierrors.IsConflict(err) {
+				retryPeriod = machineutils.ConflictRetry
+			}
 		} else {
 			// Return error even when machine object is updated
 			klog.V(2).Infof("Added finalizer to machine %q with providerID %q and backing node %q", machine.Name, getProviderID(machine), getNodeName(machine))
 			err = fmt.Errorf("Machine creation in process. Machine finalizers are UPDATED")
```
gagan16k, line 1207: We should probably return `retryPeriod, nil` here instead, on the success path, to avoid re-enqueing unnecessarily now.

pkg/util/provider/machinecontroller/machine.go
elankath: I added a `V4` log in the inner else to aid diagnosis if this problem comes again. Debugger should increase log verbosity since otherwise this will bomb the logs.

pkg/util/provider/machinecontroller/machine.go
elankath: I would prefer not doing such terrible hacks like changing the finalizer list on the OLDER machine object and then continuing with reconciliation - that will cause lovely number of new bugs 😀. 

This PR should be restricted to JUST the issue at hand and not general code improvement. Proceeding with consuming new `Machine` object and using it for reconciliation should be taken up as a general-purpose activity across MCM code-base as the current "i love to re-enqueue" idiom is a pattern applied ALL over the code-base and changing in this one specific case makes no sense to me as part of a very targeted & specific bug-fix PR. We should do it as general refactoring PR or in MCM V2.

pkg/util/provider/machinecontroller/machine.go
r4mek: If we are already re-queuing the Machine after the finalizer is added, do we need enqueue here?

pkg/util/provider/machinecontroller/machine.go
takoverflow: Makes sense.

pkg/util/provider/machinecontroller/machine_test.go
```diff
@@ -5049,4 +5054,70 @@ var _ = Describe("machine", func() {
...
+			// Sync the lister so it reflects what the fake API server now holds.
+			waitForCacheSync(stop, c)
+
+			// The machine must have the finalizer; without the fix it would not (the key would have
+			// been dropped from the queue before the Update could succeed).
+			updatedMachine, err := c.controlMachineClient.Machines(testNamespace).Get(
+				context.TODO(), machine.Name, metav1.GetOptions{},
+			)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(updatedMachine.Finalizers).To(
```
gagan16k, line 5117: Just for curiosity, should we also test for phase not to be empty? Because that was the primary indication of this bug.

pkg/util/provider/machinecontroller/machine_test.go
```diff
@@ -5049,4 +5054,70 @@ var _ = Describe("machine", func() {
...
+				}
+				return false, nil, nil
+			})
+
+			key := cache.MetaObjectToName(machine).String()
+
+			// Run DefaultMaxRetries+5 iterations — without the fix this would exhaust the worker's
+			// retry budget on the first DefaultMaxRetries calls and permanently strand the machine.
+			for i := 0; i < worker.DefaultMaxRetries+5; i++ {
+				_ = c.reconcileClusterMachineKey(key)
```
gagan16k, line 5105: Another question, should we not count the errors returned by the reconcile to see if it crosses the budget for the test? 
Currently the change is so that we do not return any on the simulated `409` errors, but a regression would do so and re-introduce the bug.

pkg/util/provider/machinecontroller/machine.go
elankath: Yes this is the way it was designed. The 409 check is done inside the `addMachineFinalizers`.

pkg/util/provider/machinecontroller/machine_util.go
elankath: No, we cannot. We need to re-enqueue even on success path if finalizers have been updated.  Only if there is NO change do we not re-enqueue.

pkg/util/provider/machinecontroller/machine_test.go
elankath: Yes, but that is a specific rare edge-case in a race condition. In the general case this is not true and the test is valid or maybe I am misunderstanding what you are asking

pkg/util/provider/machinecontroller/machine_test.go
elankath: Yes, will do this. Not sure why I had a brainfart here and missed this in the test.

elankath: Actually, now with the revised code, the test is currently correct in both cases. All we need to check is that the finalizer is definitely added - which the test checks at the end. If you have a better way to write this test, please let me know.

gagan16k: Makes sense for this particular bugfix to check if the finalizer is deleted eventually.

pkg/util/provider/machinecontroller/machine.go
elankath: Yes, that is why the comment and log was added for clarify, in the earlier code we we just returning instead of re-queuing.

elankath: anyways, I get what you mean. Since the earlier review feedback had added the re-enqueue at the point where the finalizer was added, we don't need this one now.

thiyyakat:
Thanks for the fix! 
/lgtm

aaronfern (approved):

PR #1144, opened 2026-08-31
Add integration tests for inplace machine upgrade

aaronfern:
This PR adds integration tests for machine inplace upgrade

The following scenarios have been added
- [x] Happy path - move machine to new mcs after successful upgrade
- [x] Machine transition to InPlaceUpdateFailed phase when node update fails and update-result=failed annotation is set
- [x] Machine transitions to InPlaceUpdateFailed when update result is not set within timeout
- [x] Manual inPlace upgrade

Fixes partially #1123 

**Release note**:

```other developer
Added integration tests for inPlace machine upgrade
```


pkg/test/integration/common/helpers/machine_resources.go
```diff
@@ -111,3 +130,134 @@ func (c *Cluster) IsTestMachineDeleted() bool {
...
+					},
+				},
+			},
+		},
+	}
+
+	return mcd
+}
+
+// NewMachineDeploymentWithName returns a MachineDeployment with a custom name and a unique label derived from that name,
```
takoverflow, line 190: The canonical MCM way of identifying MCD for a machine with the `name` label can be used here rather than creating a different key.

pkg/test/integration/common/helpers/machine_resources.go
```diff
@@ -1,21 +1,40 @@
...
+	// InPlaceMcdNameFailure is the name of the MCD used for the inplace update failure test
+	InPlaceMcdNameFailure = "test-mcd-inplace-failure"
+	// InPlaceMcdNameTimeout is the name of the MCD used for the inplace update timeout test
+	InPlaceMcdNameTimeout = "test-mcd-inplace-timeout"
+	// InPlaceMcdNameManual is the name of the MCD used for the manual orchestration inplace update test
+	InPlaceMcdNameManual = "test-mcd-inplace-manual"
+)
+
+var (
+	testLabels = map[string]string{"test-label": "test-label"}
```
takoverflow, line 36: What is the use-case of adding this?

pkg/test/integration/common/helpers/machine_resources.go
```diff
@@ -111,3 +130,134 @@ func (c *Cluster) IsTestMachineDeleted() bool {
...
+	for _, mc := range machineList.Items {
+		if mc.Status.CurrentStatus.Phase == v1alpha1.MachineRunning {
+			runningMachines = append(runningMachines, mc)
+		}
+	}
+	return runningMachines, nil
+}
+
+// InPlaceMcdLabels returns the unique label map used for a given inplace test MCD name
+func InPlaceMcdLabels(mcdName string) map[string]string {
```
takoverflow, line 222: If the only purpose of this is to identify machines belonging to a specific MCD, consider using the `name` label key.

pkg/test/integration/common/helpers/machine_resources.go
```diff
@@ -111,3 +130,134 @@ func (c *Cluster) IsTestMachineDeleted() bool {
...
+// CreateOrUpdateMcd creates or updates the given MachineDeployment in the specified namespace
+func (c *Cluster) CreateOrUpdateMcd(ctx context.Context, mcd v1alpha1.MachineDeployment, namespace string) error {
+	_, err := c.McmClient.MachineV1alpha1().MachineDeployments(namespace).Create(ctx, &mcd, metav1.CreateOptions{})
+	if errors.IsAlreadyExists(err) {
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			existingMCD, err := c.McmClient.MachineV1alpha1().MachineDeployments(namespace).Get(ctx, mcd.Name, metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+			mcd.ResourceVersion = existingMCD.ResourceVersion
```
takoverflow, line 235: What is this update call achieving when the MCD already exists?

pkg/test/integration/common/helpers/machine_resources.go
aaronfern: Use case would be to be able to list machines created be tests easily without having to filter out machines that were already present or created by some other cluster operation

pkg/test/integration/common/helpers/machine_resources.go
aaronfern: This is used mainly for the mc preservation tests where we reuse the same mcd but update a few fields such as `AutoPreserveFailedMachineMax` or some timeouts. Those values need to be updated and thats what this does

pkg/test/integration/common/framework.go
```diff
@@ -1658,7 +1658,435 @@ func (c *IntegrationTestFramework) ControllerTests() {
...
+				var runningMachines []v1alpha1.Machine
+				runningMachines, err = c.ControlCluster.GetRunningMachineListByLabel(ctx, controlClusterNamespace, helpers.InPlaceMcdLabels(helpers.InPlaceMcdNameHappyPath))
+				gomega.Expect(err).To(gomega.BeNil())
+				gomega.Expect(runningMachines).To(gomega.HaveLen(1))
+
+				ginkgo.By("Record original MachineSet name")
+				originalOwnerRefMcsName := ""
+				for _, b := range runningMachines[0].ObjectMeta.OwnerReferences {
+					if b.Kind == "MachineSet" {
+						originalOwnerRefMcsName = b.Name
```
takoverflow, line 1768: We can break out of the loop once we've found the `MCS` ownerRef.

pkg/test/integration/common/framework.go
```diff
@@ -1658,7 +1658,435 @@ func (c *IntegrationTestFramework) ControllerTests() {
...
+					}
+				}
+				gomega.Expect(originalOwnerRefMcsName).ToNot(gomega.Equal(""))
+
+				ginkgo.By("Trigger inPlace rollout by updating MachineClass")
+				retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+					md, _ := c.ControlCluster.McmClient.MachineV1alpha1().MachineDeployments(controlClusterNamespace).Get(ctx, helpers.InPlaceMcdNameHappyPath, metav1.GetOptions{})
```
takoverflow, line 1775: Shouldn't this `Get` error be checked for?

pkg/test/integration/common/framework.go
```diff
@@ -1658,7 +1658,435 @@ func (c *IntegrationTestFramework) ControllerTests() {
...
+				gomega.Eventually(
+					c.ControlCluster.AreMachinesInPhase,
+					c.timeout,
+					c.pollingInterval).
+					WithArguments(ctx, []string{runningMachines[0].Name}, controlClusterNamespace, v1alpha1.MachineInPlaceUpdating).
+					Should(gomega.BeTrue())
+
+				ginkgo.By("Add node.machine.sapcloud.io/update-result:successful label to node")
+				retryErr = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+					node, _ := c.TargetCluster.Clientset.CoreV1().Nodes().Get(ctx, runningMachines[0].ObjectMeta.Labels[v1alpha1.NodeLabelKey], metav1.GetOptions{})
```
takoverflow, line 1792: `Get` call's error is not being handled

pkg/test/integration/common/framework.go
```diff
@@ -1658,7 +1658,435 @@ func (c *IntegrationTestFramework) ControllerTests() {
...
+						_, hasCandidate := node.Labels[v1alpha1.LabelKeyNodeCandidateForUpdate]
+						_, hasSelected := node.Labels[v1alpha1.LabelKeyNodeSelectedForUpdate]
+						_, hasResult := node.Labels[v1alpha1.LabelKeyNodeUpdateResult]
+						return !node.Spec.Unschedulable && !hasCandidate && !hasSelected && !hasResult
+					},
+					c.timeout,
+					c.pollingInterval).Should(gomega.BeTrue())
+
+				ginkgo.By("Verify that old mcs has been scaled to zero")
+				// Old mcs is not removed because mcd.Spec.RevisionHistoryLimit is set to it's default value nil.
```
takoverflow, line 1841:
```suggestion
				// Old mcs is not removed because mcd.Spec.RevisionHistoryLimit is set to its default value nil.
```

pkg/test/integration/common/framework.go
```diff
@@ -1658,7 +1658,435 @@ func (c *IntegrationTestFramework) ControllerTests() {
 		})
 	})
 
-	// Testcase #04 | Orphaned Resources
+	// Testcase #04 | InPlace Updates
+	ginkgo.Describe("InPlace machine update", ginkgo.Ordered, func() {
```
takoverflow, line 1662: Can these tests not be run in parallel? Asking since they're using independent machines belonging to different MCDs.

takoverflow (commented):
Thanks for the tests!

Just a general question, is there a reason why a lot of `Get` calls' errors are being ignored?

I've only explicitly added comments for a few but was wondering since it happened across tests.

PR #1145, opened 2026-09-01
Fix incorrect drain error drop in preservation flow

thiyyakat:

There is a bug in `preserveMachine` where if draining a preserved node fails twice such that the node condition remains unchanged, no error is returned from the function, leading to a `LongRetry` rather than the desired `ShortRetry`, and the success log is incorrectly printed.

 The PR fixes the bug by moving the error return outside the block guarded by `needsUpdate`.

Fixes #

A log has been removed from `preserveMachine` when `drainPreservedNode` returns an error because we are already logging the error within  `drainPreservedNode`.

**Release note**:

```bugfix operator
Fix drain error drop in preservation flow where a failed drain on a previously-visited preserved node would not cause the reconciler to retry fast enough.
```

### DIFF
--- a/pkg/util/provider/machinecontroller/machine_util.go
+++ b/pkg/util/provider/machinecontroller/machine_util.go
@@ -24,6 +24,7 @@ package controller
 
 import (
 	"bytes"
+	"cmp"
 	"context"
 	"encoding/json"
 	"errors"
@@ -2425,14 +2426,12 @@ func (c *controller) preserveMachine(ctx context.Context, machine *v1alpha1.Mach
 	if needsUpdate {
 		// Step 4: Update NodePreserved Condition on Node, with drain status
 		_, err = nodeops.AddOrUpdateConditionsOnNode(ctx, c.targetCoreClient, updatedNode.Name, *newCond)
-		if drainErr != nil {
-			klog.Errorf("error draining preserved node %q for machine %q : %v", nodeName, machine.Name, drainErr)
-			return machine, drainErr
-		}
 		if err != nil {
 			klog.Errorf("error trying to update node preserved condition for node %q of machine %q : %v", nodeName, machine.Name, err)
-			return machine, err
 		}
+	}
+	if drainErr != nil || err != nil {
+		return machine, cmp.Or(drainErr, err)
 	}
 	klog.V(2).Infof("Machine %q and backing node %q preserved successfully till %v.", machine.Name, nodeName, machine.Status.CurrentStatus.PreserveExpiryTime)
 	return machine, nil


pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2425,15 +2426,13 @@ func (c *controller) preserveMachine(ctx context.Context, machine *v1alpha1.Mach
 	if needsUpdate {
 		// Step 4: Update NodePreserved Condition on Node, with drain status
 		_, err = nodeops.AddOrUpdateConditionsOnNode(ctx, c.targetCoreClient, updatedNode.Name, *newCond)
-		if drainErr != nil {
-			klog.Errorf("error draining preserved node %q for machine %q : %v", nodeName, machine.Name, drainErr)
-			return machine, drainErr
-		}
 		if err != nil {
 			klog.Errorf("error trying to update node preserved condition for node %q of machine %q : %v", nodeName, machine.Name, err)
-			return machine, err
```
gagan16k, line 2434: Just wondering if we could keep this line, and only check `drainErr` in the outer scope. Might be better for readability, but I am not sure, just thinking.

thiyyakat: Sorry, I'm not quite sure what you mean. 😅 Could you please explain a bit more?

gagan16k: Something like this

```
	if needsUpdate {
		// Step 4: Update NodePreserved Condition on Node, with drain status
		_, err = nodeops.AddOrUpdateConditionsOnNode(ctx, c.targetCoreClient, updatedNode.Name, *newCond)
		if err != nil {
			klog.Errorf("error trying to update node preserved condition for node %q of machine %q : %v", nodeName, machine.Name, err)
			return machine, err
		}
	}
	if drainErr != nil {
		return machine, drainErr
	}
```

thiyyakat: Ah okay! So the reason I didn't do it this way is because the drain error here is more important and occurred first. The second error occurred because we tried to update the node condition with the drain error.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2425,15 +2426,13 @@ func (c *controller) preserveMachine(ctx context.Context, machine *v1alpha1.Mach
...
 		}
 	}
+	if drainErr != nil || err != nil {
+		return machine, cmp.Or(drainErr, err)
```
aaronfern, line 2434: If both errors are present, why not just append the errors and return an aggregated error? With the current `cmp.Or` drain errors is preferred but it would be nice to have both errors returned

pkg/util/provider/machinecontroller/machine_util.go
aaronfern, line 2423: Can you print a log here if there is drain error? Like how we do when `AddOrUpdateConditionsOnNode` has an error

thiyyakat: `drainPreservedNode` has a lot of logging in it. Do you want logs to be printed in the caller as well?

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Agreed. Made the change: 6f1b5e4c7c08f0e83b9b0939297f9ad22449c7fd. PTAL.

pkg/util/provider/machinecontroller/machine_util.go
aaronfern: Alright, logs within `drainPreservedNode` should be sufficient

aaronfern:
Thanks for the changes!
/lgtm

PR #1146, opened 2026-09-01
Exclude cloud provider VMs of preserved machines from being cleaned up 

gagan16k:

  When a machine fails to register within the `MachineCreationTimeout` and is preserved (auto or manual), the orphan-VM safety controller (`checkMachineClass`) could delete the machine's backing VM. This could happen because at the point of preservation the machine's `spec.providerID` may not yet be persisted, so the safety controller's primary guard (`spec.ProviderID == machineID`) does not fire.

Changes:
  - `machine_safety.go` adds a guard in `checkMachineClass` that skips any VM whose machine object carries an unexpired `PreserveExpiryTime`. 
  - `machineutils/utils.go`: adds `IsPreserved(m)`, to check if a machine has a `PreserveExpiryTime`
  - `controller_utils.go`: replaces `isPreserved` in `ActiveMachines.Less` with a call to the new `IsPreserved()`. Also renamed all util methods to drop `Machine` prefix.
  - `machine_safety_test.go`: adds table entries covering the new guard
    1.  preserved (unexpired `PreserveExpiryTime`) - VM retained
    2. preservation expired - VM deleted; 
  - `docs/faq.md`: adds a Troubleshooting entry explaining why a preserved machine and its VM remain after `MachineCreationTimeout`, including the bootstrap-token expiry note.

**Release note**:

```other operator
The orphan-VM safety controller no longer deletes the backing VM of a preserved machine.
```

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -47,6 +47,7 @@ this document. Few of the answers assume that the MCM being used is in conjuctio
 - [Troubleshooting](#troubleshooting)
     - [My machine is stuck in deletion for 1 hr, why?](#my-machine-is-stuck-in-deletion-for-1-hr-why)
     - [My machine is not joining the cluster, why?](#my-machine-is-not-joining-the-cluster-why)
+    - [My machine failed to register within the MachineCreationTimeout, what can I expect?](#my-machine-failed-to-register-within-the-machinecreationtimeout-what-can-i-expect)
     - [My rolling update is stuck, why?](#my-rolling-update-is-stuck-why)
 - [Developer](#developer)
     - [How should I test my code before submitting a PR?](#how-should-i-test-my-code-before-submitting-a-pr)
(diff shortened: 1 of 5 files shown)


docs/faq.md
```diff
@@ -350,6 +351,26 @@ It could possibly be debugged with following steps:
...
+  `node.machine.sapcloud.io/preserve: auto-preserved`) instead of being terminated and replaced.
+
+- **The machine is not re-created while preserved.** A preserved machine that failed during creation is not re-created by the
+  `MachineSet` controller. The operator is expected to inspect and recover it (or explicitly stop preservation) before the
+  effective `machinePreserveTimeout` elapses.
+
+- **The backing VM is not deleted as an orphan.** The safety controller skips orphan-VM collection for machines that are
+  actually preserved (i.e. those carrying a `PreserveExpiryTime` on their status), so it will not delete the backing VM of a
+  machine that was preserved during creation, even before the machine's `providerID` has been persisted onto its spec.
+
```
thiyyakat: How does this sound?

> The backing VM of a preserved machine (PreserveExpiryTime set) is not treated as an orphaned resource and is therefore not deleted while preservation is in effect. As a result, even if a machine transitions to Failed before its providerID has been persisted to the machine spec, its backing VM will not be deleted as long as the machine is preserved.

docs/faq.md
```diff
@@ -350,6 +351,26 @@ It could possibly be debugged with following steps:
...
+> Note: The bootstrap token for a machine is valid only for the `MachineCreationTimeout`. If a node fails to register within that window and the machine is preserved, its bootstrap token will already have expired, so the node will not join the cluster on its own. This is expected behaviour and the preserved machine and its backing VM are retained for inspection and recovery rather than for the node to join later.
```
thiyyakat: > the preserved machine and its backing VM are retained for inspection and recovery rather than for the node to join later.

This may need some disambiguation. If Im not wrong, the machine cannot recover to Running. That doesn't come through from the sentence

thiyyakat (commented):
Thanks for the code and doc changes! Could you please add some test cases for the helper function you've introduced, please?

The code looks good to me 👍 . I just had a few suggestions about the doc. PTAL

docs/faq.md
gagan16k: Thanks! Made this change

docs/faq.md
gagan16k: Made some changes, PTAL

pkg/util/provider/machineutils/utils.go
```diff
@@ -175,6 +175,13 @@ func IsMachinePreservationExpired(m *v1alpha1.Machine) bool {
 	return t != nil && !t.After(time.Now())
 }
 
+// IsMachinePreserved returns true if the machine is currently being preserved, i.e. it has a
+// PreserveExpiryTime set on its status that lies in the future.
+func IsMachinePreserved(m *v1alpha1.Machine) bool {
```
aaronfern, line 180: Rename this to `IsPreserved`. Since the package name specifies machine utils, having machine here is redundant and it will read better

gagan16k: The other functions above this one have a `IsMachine.....` pattern to them. If I am to apply this change, should I rename all of them?

aaronfern: Yes, would be nice to have them all with the updated pattern

thiyyakat: ~@aaronfern , the default package name for this would be `utils` and not `machineutils`. So should we still rename it without the `Machine`?~

Apologies. I made a mistake while deciphering the path. `Machine` is redundant in the function name 👍

gagan16k: Made these changes

aaronfern:
Thanks for the changes! This looks fine
/lgtm

thiyyakat:
@gagan16k , there's one file: https://github.com/gardener/machine-controller-manager/blob/master/docs/todo/outline.md which still references `IsMachineActive`, could you please make the change here too?

thiyyakat:
/lgtm

Thanks for the fix!

thiyyakat (approved):

PR #1147, opened 2026-09-01
Preservation bugfix to put PreserveExpiryTime whenever machine transitions to Failed phase

r4mek:
- Now we add the `PreserveExpiryTime` on the machine if the machine has annotation: `node.machine.sapcloud.io/preserve: "when-failed"` and it transitions to `Failed` phase.
- For auto-preservation flow, we add the `PreserveExpiryTime` on the machine if the failed machine gets the annotation: `node.machine.sapcloud.io/preserve: "auto-preserved"`
- Miscellaneous code fix:
  - Ensure objects retrieved from the lister cache are never mutated directly.
  - Machine fetch in `UpdateMachineWithRetries()` from the lister cache, which may return a stale object.
  - On update failure, requeue the object for the next reconciliation rather than retrying with a potentially stale cached value.

Fixes #

All the miscellaneous code fix are much needed to fix the aforementioned bug without surfacing/introducing new bugs.

**Release note**:

```other operator
Fixed the scenario where a machine to be preserved can for brief moment not have PreserveExpiryTime set.
```


pkg/controller/deployment_test.go
```diff
@@ -2356,7 +2357,10 @@ var _ = Describe("machineDeployment", func() {
 
 				defer trackers.Stop()
 				waitForCacheSync(stop, c)
-				err := c.updateMachineAndMachineDeploymentDeletionAnnotations(context.TODO(), testMachineDeployment)
+				err := func() error {
```
thiyyakat, line 2360: This may be a stupid question, but what is the point of the closure here? Can you not simply have just L2361?

pkg/util/provider/machineutils/utils.go
```diff
@@ -180,34 +173,24 @@ func GetMachineDeploymentName(machine *v1alpha1.Machine) string {
...
-		if err != nil {
-			return err
+// GetPreserveAnnotationValue returns the preserve annotation value for the given node and machine.
+// Invalid annotation values are treated as absent.
+func GetPreserveAnnotationValue(node *corev1.Node, machine *v1alpha1.Machine) (string, bool) {
+	if node != nil {
+		if val, ok :=
+			node.Annotations[PreserveMachineAnnotationKey]; ok &&
+			AllowedPreserveAnnotationValues.Has(val) {
+			return val, true
```
thiyyakat, line 228: Please print a warning here if the value is not valid.

pkg/util/provider/machineutils/utils.go
```diff
@@ -180,34 +173,24 @@ func GetMachineDeploymentName(machine *v1alpha1.Machine) string {
...
-	if retryErr == errorsutil.ErrPreconditionViolated {
-		klog.V(4).Infof("Machine %s precondition doesn't hold, skip updating it.", name)
-		retryErr = nil
 	}
-
-	return machine, retryErr
+	if val, ok :=
+		machine.Annotations[PreserveMachineAnnotationKey]; ok &&
+		AllowedPreserveAnnotationValues.Has(val) {
+		return val, true
```
thiyyakat, line 238: Please log a warning if the value is not valid.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -786,38 +779,15 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
+	if !exists {
 		return
-	} else if getErr != nil {
-		if !apierrors.IsNotFound(getErr) {
-			err = getErr
-			return
-		}
-		klog.Warningf("Couldn't find node %q for machine %q", nodeName, machine.Name)
-		err = nil
 	}
```
thiyyakat: Please correct me if I'm wrong, but what about the case where preservation was being manipulated by only the machine object's annotations? If a machine was preserved, and the annotation was deleted to indicate that preservation should be stopped, you will end up returning early here instead of stopping preservation.

thiyyakat (commented):
Thanks for the changes and the refactoring! I haven't looked at the full PR yet. I had one case in mind for which I thought the new code may not behave as expected. Once you address that, I will look at the rest as well.

pkg/util/provider/machinecontroller/machine.go
r4mek: Good catch! Addressed in the latest commit https://github.com/gardener/machine-controller-manager/pull/1147/commits/ab6bf4420e1e3d56f4e1d798658db512a2f69c9e

pkg/controller/deployment.go
```diff
@@ -674,36 +674,47 @@ func (dc *controller) updateMachineAndMachineDeploymentDeletionAnnotations(ctx c
...
-		machineDeepCopy := machine.DeepCopy()
-		if machineDeepCopy.Annotations == nil {
-			machineDeepCopy.Annotations = make(map[string]string)
-		}
-		machineDeepCopy.Annotations[machineutils.MachinePriority] = "1"
-		if machineDeepCopy.Annotations[machineutils.MarkedForDeletionTime] == "" {
-			machineDeepCopy.Annotations[machineutils.MarkedForDeletionTime] = tgd.markedMachineDeletionTimes[i]
-		}
-		_, err = dc.controlMachineClient.Machines(machine.Namespace).Update(ctx, machineDeepCopy, metav1.UpdateOptions{})
+		deletionTime := tgd.markedMachineDeletionTimes[i]
```
gagan16k, line 691: Do we need this variable `deletionTime`?

pkg/controller/deployment.go
```diff
@@ -674,36 +674,47 @@ func (dc *controller) updateMachineAndMachineDeploymentDeletionAnnotations(ctx c
...
+			}
+			return nil
+		}, true)
 		if err != nil {
 			klog.Errorf("failed to set MachinePriority=1 annotation on Machine %q of MachineDeployment %q: %v", machine.Name, mcd.Name, err)
-			return
+			return mcd, err
+		}
+		// TODO: not neat. refactor later.
+		for _, machineList := range machineMap {
```
gagan16k, line 707: Can we key the `machineMap` by the UID of the `machineSet` here, might save a loop
Something like
```
if controllerRef := metav1.GetControllerOf(updatedMachine); controllerRef != nil {
	if machineList, ok := machineMap[controllerRef.UID]; ok {
		for i := range machineList.Items {
```

pkg/controller/machineset.go
```diff
@@ -955,24 +965,39 @@ func (c *controller) manageAutoPreservationOfFailedMachines(ctx context.Context,
...
+				m.Annotations = make(map[string]string)
+			}
+			m.Annotations[machineutils.PreserveMachineAnnotationKey] = machineutils.PreserveMachineAnnotationValueAutoPreserved
+			if m.Spec.MachineConfiguration != nil && m.Spec.MachineConfiguration.MachinePreserveTimeout != nil {
+				m.Status.CurrentStatus.PreserveExpiryTime = &metav1.Time{Time: metav1.Now().Add(m.Spec.MachineConfiguration.MachinePreserveTimeout.Duration)}
+			} else {
+				m.Status.CurrentStatus.PreserveExpiryTime = &metav1.Time{Time: metav1.Now().Add(c.safetyOptions.MachinePreserveTimeout.Duration)}
+			}
+			return nil
+		}, true)
```
gagan16k: As per [docs](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#spec-and-status)
```
The PUT and POST verbs on objects MUST ignore the status values, to avoid accidentally overwriting the status in read-modify-write scenarios. A /status subresource MUST be provided to enable system components to update statuses of resources they manage.
```

On annotating machine (with mcm scaled to zero to avoid it reconciling) with these keys and dummy values
```
(⎈|garden--aws-ha-external:garden shoot--xxx--demo)➜  machine-controller-manager git:(pr/r4mek/1147) ✗ k patch mc shoot--xxx--demo-worker-cpu-z1-58c68-qjrl2 --type=merge -p '{
  "metadata":{"annotations":{"node.machine.sapcloud.io/preserve":"auto-preserved"}},
  "status":{"currentStatus":{"preserveExpiryTime":"2030-01-01T00:00:00Z"}}
}'
machine.machine.sapcloud.io/shoot--xxx--demo-worker-cpu-z1-58c68-qjrl2 patched

(⎈|garden--aws-ha-external:garden shoot--xxx--demo)➜  machine-controller-manager git:(pr/r4mek/1147) ✗ k get mc shoot--xxx--demo-worker-cpu-z1-58c68-qjrl2 -oyaml | grep "annotations" -A 2
  annotations:
    machinepriority.machine.sapcloud.io: "3"
    node.machine.sapcloud.io/preserve: auto-preserved

(⎈|garden--aws-ha-external:garden shoot--xxx--demo)➜  machine-controller-manager git:(pr/r4mek/1147) ✗ k get mc shoot--xxx--demo-worker-cpu-z1-58c68-qjrl2 -oyaml | grep "preserveExpiryTime"

(⎈|garden--aws-ha-external:garden shoot--xxx--demo)➜  machine-controller-manager git:(pr/r4mek/1147) ✗
```

As the "/status" subresource is not specified here, it drops the change on it. Maybe we would need two different patches for each change as they target different resource endpoints?.

pkg/controller/deployment_util.go
```diff
@@ -963,26 +962,22 @@ func WaitForMachinesHashPopulated(ctx context.Context, c v1alpha1listers.Machine
...
-					// Precondition: the machine doesn't contain the new hash in its label.
-					if machineToUpdate.Labels[v1alpha1.DefaultMachineDeploymentUniqueLabelKey] == hash {
-						return errors.ErrPreconditionViolated
-					}
-					machineToUpdate.Labels = labelsutil.AddLabel(machineToUpdate.Labels, v1alpha1.DefaultMachineDeploymentUniqueLabelKey, hash)
-					return nil
-				})
+			updatedMachine, err := machineutils.PatchMachine(ctx, c.Machines(machine.Namespace), &machine, func(m *v1alpha1.Machine) error {
+				m.Labels = labelsutil.AddLabel(m.Labels, v1alpha1.DefaultMachineDeploymentUniqueLabelKey, hash)
+				return nil
```
gagan16k, line 975: This will force a reconcile instead of retry on returning `409` errors, so the unit test for `LDRCBST` could flake even more now. Might need to find another way to correct that.

pkg/controller/machineset.go
```diff
@@ -534,19 +534,25 @@ func (c *controller) reconcileClusterMachineSet(key string) error {
...
+	filteredMachines, err := c.claimMachines(ctx, machineSet, selector, machineList)
 	if err != nil {
 		return err
 	}
 
+	// Deep-copy to avoid mutating cache objects downstream.
+	for i, m := range filteredMachines {
+		filteredMachines[i] = m.DeepCopy()
+	}
+
```
gagan16k, line 553: Great change, but I was thinking if we could apply this while creating the `filteredMachines` slices, in `claimMachines()`.
Argument for: Saves extra work on the slice, and does it in the same loop while the slice is getting constructed.
Argument against: The other callers for the function, do not seem to write to the slice like this reconciler does, and thus do not really need a deep copy. WDYT?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -786,69 +780,42 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
-	}
-	if preserveInfo.nodeAnnotated && !machineutils.AllowedPreserveAnnotationValues.Has(preserveInfo.nodeValue) {
-		klog.Warningf("Preserve annotation %q=%q on node %q backing machine %q is invalid", machineutils.PreserveMachineAnnotationKey, preserveInfo.nodeValue, nodeName, machine.Name)
-		return
-	}
-	preservationBound := isMachinePreservationBound(&preserveInfo)
-	if !preservationBound {
-		// We clear the error here to prevent preservation logic from interfering with non-preservation-bound machines.
-		err = nil
+	node, _ := c.nodeLister.Get(nodeName)
```
gagan16k, line 783: Should we handle non `notFound` errors as well, similar to the other places in the codebase

pkg/util/provider/machineutils/utils.go
```diff
@@ -180,34 +179,66 @@ func GetMachineDeploymentName(machine *v1alpha1.Machine) string {
...
 
-	return machine, retryErr
+// GetPreserveAnnotationValue returns the preserve annotation value for the given node and machine
+// and a boolean informing whether we need to do any work or skip.
+// Invalid annotation values are treated as absent.
+func GetPreserveAnnotationValue(node *corev1.Node, machine *v1alpha1.Machine) (annotationValue string, shouldHandlePreservation bool) {
+	if node != nil {
+		if val, ok :=
+			node.Annotations[PreserveMachineAnnotationKey]; ok &&
+			AllowedPreserveAnnotationValues.Has(val) {
```
gagan16k, line 227: If there is a garbage value for preservation, it is possible we return `"", true"` from this function. This can cause a preserved machine to stop preservation. 
Seems like acceptable behaviour, but wanted to confirm if this is part of the feature.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -871,97 +838,58 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
 		}
 	}
 
-	if shouldAnnotationsBeUpdatedOnMachine(removeAnnotations, &preserveInfo) {
-		err = c.updatePreserveAnnotationOnMachine(ctx, preserveInfo.nodeValue, clone)
+	if node != nil {
+		updatedMachine, err = c.updatePreserveAnnotationOnMachine(ctx, node.Annotations[machineutils.PreserveMachineAnnotationKey], clone)
```
gagan16k, line 842: If the patch fails here with a `NotFound` error, `updatedMachine` would be nil. The err would be cleared by the defer, so the caller reconcile function proceeds with the `nil` machine instead of re-queueing.
This will cause a nil pointer deref panic at L258

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -2143,6 +2147,7 @@ func (c *controller) UpdateNodeTerminationCondition(ctx context.Context, machine
 func (c *controller) updateMachineToFailedState(ctx context.Context, description string, machine, clone *v1alpha1.Machine) (bool, error) {
 	// Log the error message for machine failure
 	klog.Error(description)
+	node, _ := c.nodeLister.Get(machine.Annotations[v1alpha1.NodeLabelKey])
```
gagan16k, line 2150:
```suggestion
	node, _ := c.nodeLister.Get(machine.Labels[v1alpha1.NodeLabelKey])
```

pkg/controller/machineset.go
```diff
@@ -955,24 +965,39 @@ func (c *controller) manageAutoPreservationOfFailedMachines(ctx context.Context,
...
+		updatedMachine, err := machineutils.PatchMachine(ctx, c.controlMachineClient.Machines(machine.Namespace), machine, func(m *v1alpha1.Machine) error {
+			if m.Annotations == nil {
+				m.Annotations = make(map[string]string)
+			}
+			m.Annotations[machineutils.PreserveMachineAnnotationKey] = machineutils.PreserveMachineAnnotationValueAutoPreserved
+			if m.Spec.MachineConfiguration != nil && m.Spec.MachineConfiguration.MachinePreserveTimeout != nil {
+				m.Status.CurrentStatus.PreserveExpiryTime = &metav1.Time{Time: metav1.Now().Add(m.Spec.MachineConfiguration.MachinePreserveTimeout.Duration)}
+			} else {
+				m.Status.CurrentStatus.PreserveExpiryTime = &metav1.Time{Time: metav1.Now().Add(c.safetyOptions.MachinePreserveTimeout.Duration)}
+			}
```
gagan16k, line 988: Same logic as `getEffectiveMachinePreserveTimeout()`, maybe that can be used here as well?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -871,97 +838,58 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
-	if clone.Annotations == nil {
-		clone.Annotations = make(map[string]string)
+func (c *controller) updatePreserveAnnotationOnMachine(ctx context.Context, nodeValue string, machine *v1alpha1.Machine) (*v1alpha1.Machine, error) {
+	if nodeValue == "" {
+		if _, exists := machine.Annotations[machineutils.LastAppliedNodePreserveValueAnnotationKey]; !exists {
+			return machine, nil
+		}
+		klog.V(3).Infof(
+			"Since node %q 's annotation:%q was removed, removing machine %q 's annotation:%q",
+			machine.Annotations[v1alpha1.NodeLabelKey],
```
gagan16k, line 858:
```suggestion
			machine.Labels[v1alpha1.NodeLabelKey],
```

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -871,97 +838,58 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
+			machineutils.LastAppliedNodePreserveValueAnnotationKey,
+		)
 	} else {
-		delete(clone.Annotations, machineutils.PreserveMachineAnnotationKey)
+		klog.V(3).Infof(
+			"Syncing machine %q 's annotation:%q=%q to its node %q 's annotation:%q=%q",
+			machine.Name,
+			machineutils.LastAppliedNodePreserveValueAnnotationKey,
+			machine.Annotations[machineutils.LastAppliedNodePreserveValueAnnotationKey],
+			machine.Annotations[v1alpha1.NodeLabelKey],
```
gagan16k, line 869:
```suggestion
			machine.Labels[v1alpha1.NodeLabelKey],
```

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -871,97 +838,58 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
 	}
-	clone.Annotations[machineutils.LastAppliedNodePreserveValueAnnotationKey] = nodeValue
-	_, err := c.controlMachineClient.Machines(clone.Namespace).Update(ctx, clone, metav1.UpdateOptions{})
-	return err
+	klog.V(3).Infof(
+		"Removing machine %q 's annotation:%q=%q as node %q 's has annotation:%q=%q",
+		machine.Name,
+		machineutils.PreserveMachineAnnotationKey,
+		machine.Annotations[machineutils.PreserveMachineAnnotationKey],
+		machine.Annotations[v1alpha1.NodeLabelKey],
```
gagan16k, line 879:
```suggestion
		machine.Labels[v1alpha1.NodeLabelKey],
```

pkg/controller/machineset.go
```diff
@@ -1007,19 +1036,6 @@ func (c *controller) stopAutoPreservationForMachines(ctx context.Context, machin
...
-	machineToUpdate.Annotations[machineutils.PreserveMachineAnnotationKey] = machineutils.PreserveMachineAnnotationValueAutoPreserved
-	return nil
-}
-
-func removeAutoPreserveAnnotationFromMachine(machineToUpdate *v1alpha1.Machine) error {
-	delete(machineToUpdate.Annotations, machineutils.PreserveMachineAnnotationKey)
-	return nil
-}
-
 func (c *controller) findEffectivePreserveValue(machine *v1alpha1.Machine) (string, error) {
```
gagan16k, line 1039: This function is redundant with the introduction of `GetPreserveAnnotationValue()`

pkg/util/provider/machineutils/utils.go
gagan16k, line 227:
```suggestion
		if val, ok := node.Annotations[PreserveMachineAnnotationKey]; ok && AllowedPreserveAnnotationValues.Has(val) {
```

Not a strong opinion, but I found this kind of confusing to read the first time. But that could be just me, feel free to ignore this comment.

pkg/util/provider/machineutils/utils.go
```diff
@@ -180,34 +179,66 @@ func GetMachineDeploymentName(machine *v1alpha1.Machine) string {
...
-		return err
-	})
-
-	// Ignore the precondition violated error, this machine is already updated
-	// with the desired label.
-	if retryErr == errorsutil.ErrPreconditionViolated {
-		klog.V(4).Infof("Machine %s precondition doesn't hold, skip updating it.", name)
-		retryErr = nil
 	}
+	return machineClient.Patch(ctx, machine.Name, types.MergePatchType, patch, metav1.PatchOptions{}, subresources...)
```
gagan16k, line 217:
```suggestion
	return machineClient.Patch(ctx, machine.Name, types.StrategicMergePatchType, patch, metav1.PatchOptions{}, subresources...)
```

pkg/util/provider/machineutils/utils.go
```diff
@@ -180,34 +179,66 @@ func GetMachineDeploymentName(machine *v1alpha1.Machine) string {
...
+		return nil, err
+	}
+	modifiedJSON, err := json.Marshal(modified)
+	if err != nil {
+		return nil, err
+	}
+	patch, err := strategicpatch.CreateTwoWayMergePatch(base, modifiedJSON, v1alpha1.Machine{})
+	if err != nil {
+		return nil, err
+	}
```
gagan16k, line 200:
```suggestion
	}
	if string(patch) == "{}" {
		return machine, nil
	}
```

We could short-circuit if the patch is empty for whatever reason, as the request will still be considered valid by the API server

gagan16k (commented):
Thanks for the changes! Have some comments, PTAL

cmd/machine-controller-manager/app/options/options.go
```diff
@@ -98,6 +99,7 @@ func (s *MCMServer) AddFlags(fs *pflag.FlagSet) {
 	fs.Int32Var(&s.SafetyOptions.SafetyDown, "safety-down", s.SafetyOptions.SafetyDown, "Upper-limit minus safety-down value gives the lower-limit. This is the limits below which any temporarily frozen machineSet/machineDeployment object is unfrozen. lower-limit = desired + maxSurge (if applicable) + safetyUp - safetyDown.")
 
 	fs.DurationVar(&s.SafetyOptions.MachineSafetyOvershootingPeriod.Duration, "machine-safety-overshooting-period", s.SafetyOptions.MachineSafetyOvershootingPeriod.Duration, "Time period (in duration) used to poll for overshooting of machine objects backing a machineSet by safety controller.")
+	fs.DurationVar(&s.SafetyOptions.MachinePreserveTimeout.Duration, "machine-preserve-timeout", s.SafetyOptions.MachinePreserveTimeout.Duration, "Duration for which a failed machine should be preserved if it has the appropriate preserve annotation set.")
```
aaronfern, line 102: The issue description does not mention any new flag. A new flag like this is important enough to be mentioned as a release note

pkg/controller/machineset.go
```diff
@@ -955,24 +965,39 @@ func (c *controller) manageAutoPreservationOfFailedMachines(ctx context.Context,
...
+				m.Status.CurrentStatus.PreserveExpiryTime = &metav1.Time{Time: metav1.Now().Add(c.safetyOptions.MachinePreserveTimeout.Duration)}
+			}
+			return nil
+		}, true)
 		if err != nil {
-			klog.V(2).Infof("Error annotating machine %q for auto-preservation: %v", m.Name, err)
-			// since addAutoPreserveAnnotation uses retries internally, on error we can continue with other machines
-			continue
+			klog.Errorf("Error annotating and setting PreserveExpiryTime on machine %q for auto-preservation: %v", machine.Name, err)
+			return nil, err
```
aaronfern: Returning here stops patching of further machines. Consider pooling all error messages and returning an aggregate error at the end of the loop

pkg/controller/machineset_util.go
```diff
@@ -111,96 +107,66 @@ func (c *controller) syncMachinesNodeTemplates(ctx context.Context, machineList
...
 }
 
-// copyMachineSetNodeTemplatesToMachines copies machineset's nodeTemplate to machine's nodeTemplate,
-// and returns true if machine's nodeTemplate is changed.
-// Note that apply and revision nodeTemplates are not copied.
-func copyMachineSetNodeTemplatesToMachines(machineset *v1alpha1.MachineSet, machine *v1alpha1.Machine) bool {
-	machineSetNodeTemplateCopy := machineset.Spec.Template.Spec.NodeTemplateSpec.DeepCopy()
-	machineNodeTemplateCopy := machine.Spec.NodeTemplateSpec.DeepCopy()
+// nodeTemplateChanged returns true if machine's nodeTemplate is changed.
+func nodeTemplateChanged(machineset *v1alpha1.MachineSet, machine *v1alpha1.Machine) bool {
```
aaronfern: This func name is semantically incorrect. Generally a machine's nodeTemplate won't change but would rather get outdated and needs to be re-synced.
Consider names like `nodeTemplateOutOfSync` or `machineNodeTemplateMismatch` or feel free to choose another name that you feel is more accutare

pkg/controller/machineset_util.go
```diff
@@ -111,96 +107,66 @@ func (c *controller) syncMachinesNodeTemplates(ctx context.Context, machineList
...
-// Note that apply and revision config are not copied.
-func copyMachineSetConfigToMachines(machineset *v1alpha1.MachineSet, machine *v1alpha1.Machine) bool {
-	isConfigChanged := false
-
-	machineSetConfigCopy := machineset.Spec.Template.Spec.MachineConfiguration.DeepCopy()
-	machineConfigCopy := machine.Spec.MachineConfiguration.DeepCopy()
-
-	isConfigChanged = !(apiequality.Semantic.DeepEqual(machineSetConfigCopy, machineConfigCopy))
+// configChanged returns true if machine's config is changed.
+func configChanged(machineset *v1alpha1.MachineSet, machine *v1alpha1.Machine) bool {
```
aaronfern: Same as above, name is not accurate

pkg/controller/machineset_util.go
```diff
@@ -111,96 +107,66 @@ func (c *controller) syncMachinesNodeTemplates(ctx context.Context, machineList
...
-// and returns true if machine's class.Kind is changed.
-func copyMachineSetClassKindToMachines(machineset *v1alpha1.MachineSet, machine *v1alpha1.Machine) bool {
-	if machineset.Spec.Template.Spec.Class.Kind != machine.Spec.Class.Kind {
-		machine.Spec.Class.Kind = machineset.Spec.Template.Spec.Class.Kind
-		return true
-	}
-
-	return false
+// classKindChanged returns true if machine's class.Kind is changed.
+func classKindChanged(machineset *v1alpha1.MachineSet, machine *v1alpha1.Machine) bool {
```
aaronfern: This naming is not accurate too

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -428,6 +428,14 @@ func (c *controller) triggerCreationFlow(ctx context.Context, createMachineReque
 						} else {
 							klog.V(2).Infof("VM successfully deleted in context of stale node obj for machine %q", machine.Name)
 						}
+						machineCurrentStatus := v1alpha1.CurrentStatus{
+							Phase:          v1alpha1.MachineFailed,
+							LastUpdateTime: metav1.Now(),
+						}
```
aaronfern, line 470: Does having this out and assigned to another var help much?

pkg/util/provider/machinecontroller/machine.go
aaronfern: In any case, returned err should not be ignored

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -871,97 +838,58 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
+			"Since node %q 's annotation:%q was removed, removing machine %q 's annotation:%q",
+			machine.Annotations[v1alpha1.NodeLabelKey],
+			machineutils.PreserveMachineAnnotationKey,
+			machine.Name,
+			machineutils.LastAppliedNodePreserveValueAnnotationKey,
+		)
 	} else {
-		delete(clone.Annotations, machineutils.PreserveMachineAnnotationKey)
+		klog.V(3).Infof(
+			"Syncing machine %q 's annotation:%q=%q to its node %q 's annotation:%q=%q",
```
aaronfern: This is worded incorrectly. We sync the node's annotation to the machine

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1141,6 +1141,10 @@ func (c *controller) reconcileMachineHealth(ctx context.Context, machine *v1alph
 					LastUpdateTime:     metav1.Now(),
 					PreserveExpiryTime: machine.Status.CurrentStatus.PreserveExpiryTime,
 				}
+				// check if preservation is needed for the failed machine
+				if val, shouldHandlePreservation := machineutils.GetPreserveAnnotationValue(node, machine); shouldHandlePreservation && val == machineutils.PreserveMachineAnnotationValueWhenFailed {
+					clone.Status.CurrentStatus.PreserveExpiryTime = &metav1.Time{Time: metav1.Now().Add(c.getEffectiveMachinePreserveTimeout(machine).Duration)}
+				}
```
aaronfern, line 1148: This snippet could potentially overwrite existing `PreserveExpiryTime` on the machine that is set a line above

Although, @thiyyakat do you know why we might already have a `PreserveExpiryTime` on the machine? We enter this block only when the machine is in the `Pending` phase so I wonder if line 1142 is ever relevant

pkg/util/provider/machinecontroller/machine_util.go
aaronfern: Please don't ignore a potential error

aaronfern (commented):
General feedback
1. Newly introduced flag should be mentioned in description and release notes
2. Error strings should not be capitalised
3. In general it's not nice that this huge bunch of miscellaneous code fixes are part of this bug fix. They are good pieces, but not required to fix the bug this PR talks about. It's fine for now as the work is done already, but keep this in mind for future reference

pkg/util/provider/machinecontroller/machine.go
r4mek: Yes. I update the `PreserveExpiryTime` below, based on certain criteria. If there is a better way, I can change it.

pkg/util/provider/machinecontroller/machine_util.go
r4mek: After offline discussions, we decided to log the error.

pkg/util/provider/machinecontroller/machine_util.go
r4mek: Correct me if I am wrong @thiyyakat 

Only for `preserve=now` `PreserveExpiryTime` is set else it is nil. I am setting `PreserveExpiryTime` only for `preserve=when-failed`, so they should not conflict.

pkg/util/provider/machinecontroller/machine.go
r4mek: After offline discussions, we decided to log the error.

pkg/util/provider/machineutils/utils.go
r4mek: This was intended behaviour and was mentioned as well
```go
// Invalid annotation values are treated as absent.
func GetPreserveAnnotationValue(node *corev1.Node, machine *v1alpha1.Machine) (annotationValue string, shouldHandlePreservation bool) {
```

If there is a better approach, I can change it.

pkg/util/provider/machinecontroller/machine.go
r4mek: Will change it like this

```go
if node != nil {
	updatedMachine, err = c.updatePreserveAnnotationOnMachine(ctx, node.Annotations[machineutils.PreserveMachineAnnotationKey], clone)
	if err != nil {
		updatedMachine = clone
	}
}
```

pkg/controller/machineset.go
r4mek: `getEffectiveMachinePreserveTimeout()` can't be used directly. @thiyyakat had an idea to extract the logic to a common function so that both controllers can use it.

I didn't implement it as I thought refactoring was not worth since we are only using it once here. If there is further reuse of the logic, we can think of refactoring it.

pkg/controller/machineset.go
r4mek: I did not implement it in `claimMachines()` for the reason mentioned in `Argument against`. But if the team sees differently, I can implement it.

pkg/controller/machineset.go
r4mek: Nice catch! I missed it in my testing as `manageMachinePreservation()` was also setting `PreserveExpiryTime` at the same time.

Will split it into two calls - one for annotation the other for status.

pkg/controller/deployment.go
r4mek: Not anymore. Will remove it.

pkg/controller/deployment_util.go
r4mek: After our offline discussions, the error is very low and honestly I am not really sure why the `LDRCBST` test is failing. More investigation would be required for fixing it and not sure whether we want to tackle it now.

@aaronfern

pkg/util/provider/machineutils/utils.go
r4mek: changed the code to use `jsonpatch.CreateMergePatch()`

pkg/apis/machine/v1alpha1/constants.go
```diff
@@ -26,3 +32,9 @@ const (
 	// LabelValueNodeUpdateFailed is the label value that indicates the update on the node has failed.
 	LabelValueNodeUpdateFailed = "failed"
 )
+
+var (
+	// DefaultMachinePreserveTimeout is the default time for which the machine is preserved
+	// when [MachineConfiguration.MachinePreserveTimeout] is not specified.
+	DefaultMachinePreserveTimeout = metav1.Duration{Duration: 96 * time.Hour}
```
thiyyakat, line 39: Why not change this to a constant?

```
DefaultMachinePreserveTimeout =  96 * time.Hour
```
And, in `manageAutoPreservationOfFailedMachines`
```
m.Status.CurrentStatus.PreserveExpiryTime = &metav1.Time{Time: metav1.Now().Add(v1alpha1.DefaultMachinePreserveTimeout)}
```

pkg/controller/machineset.go
```diff
@@ -563,7 +569,10 @@ func (c *controller) reconcileClusterMachineSet(key string) error {
 		return err
 	}
 
-	filteredMachines = c.manageAutoPreservationOfFailedMachines(ctx, filteredMachines, machineSet)
+	filteredMachines, err = c.manageAutoPreservationOfFailedMachines(ctx, filteredMachines, machineSet)
+	if err != nil {
```
thiyyakat, line 573: Since we don't update `AutoPreserveFailedMachineCount` until `calculateMachineSetStatus` is called in `reconcileClusterMachineSet`, there is a chance that a few machines would be successfully annotated with the auto-preserve annotation, but in the next reconciliation, since `AutoPreserveFailedMachineCount` is stale, `manageAutoPreservationOfFailedMachines` will end up preserving additional machines above `AutoPreserveFailedMachineMax`.

However, this will self-heal because the annotation update on the machine object will requeue the machineset object as well. 


@gagan16k , I think your fix will reduce the likelihood of this happening, right? If we do not rely on `AutoPreserveFailedMachineCount` set on the status? We need to make sure both fixes go in together.

pkg/controller/machineset_test.go
```diff
@@ -2330,60 +2331,6 @@ var _ = Describe("machineset", func() {
 					result: false,
 				},
 			}),
-			Entry("should return true if machine is annotated with preserve=false", testCase{
```
thiyyakat, line 2333: Could you please add a test case for when preserveExpiryTime is not set on a Failed machine?

pkg/controller/machineset.go
```diff
@@ -955,24 +952,50 @@ func (c *controller) manageAutoPreservationOfFailedMachines(ctx context.Context,
...
+		}, true)
+		if err != nil {
+			klog.Errorf("could not annotate machine %q for auto-preservation: %v", machine.Name, err)
+			errs = append(errs, err)
+			continue
+		}
+
+		preservedMachine, err := machineutils.PatchMachine(ctx, c.controlMachineClient.Machines(annotatedMachine.Namespace), annotatedMachine, func(m *v1alpha1.Machine) error {
+			if annotatedMachine.Spec.MachineConfiguration != nil && annotatedMachine.Spec.MachineConfiguration.MachinePreserveTimeout != nil {
+				m.Status.CurrentStatus.PreserveExpiryTime = &metav1.Time{Time: metav1.Now().Add(annotatedMachine.Spec.MachineConfiguration.MachinePreserveTimeout.Duration)}
```
thiyyakat, line 982: In the current state of the code in your PR, is there a possibility that a machine is annotated with `auto-preserved`, causing the machine and its owner object (the machineset) to get reconciled again, but the PreserveExpiryTime is not set, so in the latest reconciliation, the machine is found to be without PreserveExpiryTime and is thus deleted?


I was wondering if we could instead set the PreserveExpiryTime first and then add the auto-preserved annotation(just for tracking). But if there is an error during the status update, it would be difficult to add this annotation correctly to the same machine. Plus semantically it seems odd because we are using the annotation as an intent here.

Instead, in the `shouldFailedMachineBeTerminated`, what if we check for either preserve expiry time or auto-preserved annotation, and not just preserve expiry time?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -823,69 +816,46 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
-		klog.Warningf("Preserve annotation %q=%q on machine %q is invalid", machineutils.PreserveMachineAnnotationKey, preserveInfo.machineValue, machine.Name)
-		preserveInfo.machineAnnotated = false
-		preserveInfo.machineValue = ""
-	}
-	if preserveInfo.nodeAnnotated && !machineutils.AllowedPreserveAnnotationValues.Has(preserveInfo.nodeValue) {
-		klog.Warningf("Preserve annotation %q=%q on node %q backing machine %q is invalid", machineutils.PreserveMachineAnnotationKey, preserveInfo.nodeValue, nodeName, machine.Name)
-		return
+	node, err := c.nodeLister.Get(nodeName)
+	if err != nil {
+		klog.V(3).Infof("Error fetching node %q . Will check the machine %q for annotation:%q", nodeName, machine.Name, machineutils.PreserveMachineAnnotationKey)
```
thiyyakat, line 821: This log will get printed every reconciliation until a node object registers and the lister cache gets updated. Even for non-preservation-related machines. I would suggest doing away with the log or increasing the level.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -823,69 +816,46 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
-		}
-		klog.Warningf("Couldn't find node %q for machine %q", nodeName, machine.Name)
-		err = nil
 	}
-	// Note: when the backing node cannot be found, we assume the machine's annotation value needs to be enforced to enable
-	// preservation of the machine object.
-	effectivePreserveValue := getEffectivePreservationAnnotations(&preserveInfo, getErr)
 
-	var removeAnnotations bool
 	clone := machine.DeepCopy()
```
thiyyakat, line 829: Why not use `updatedMachine` here? Is it because you don't want to return `nil` on error? Can that not be taken care of in the `defer` ?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -908,97 +878,61 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
+			machine.Name,
+			machineutils.LastAppliedNodePreserveValueAnnotationKey,
+			nodeValue,
+		)
 	}
-	clone.Annotations[machineutils.LastAppliedNodePreserveValueAnnotationKey] = nodeValue
-	_, err := c.controlMachineClient.Machines(clone.Namespace).Update(ctx, clone, metav1.UpdateOptions{})
-	return err
+	klog.V(3).Infof(
+		"Removing machine %q 's annotation:%q=%q as node %q 's has annotation:%q=%q",
```
thiyyakat, line 918:
```suggestion
		"Removing machine %q 's annotation:%q=%q as node %q has annotation:%q=%q",
```
Nit.

pkg/util/provider/machinecontroller/machine_util.go
thiyyakat: Yes. As long as we're checking for the annotation value == `when-failed`, it should be fine.

pkg/util/provider/machineutils/utils.go
```diff
@@ -180,34 +181,90 @@ func GetMachineDeploymentName(machine *v1alpha1.Machine) string {
...
+	node *corev1.Node,
+	machine *v1alpha1.Machine,
+) (annotationValue string, shouldHandlePreservation bool) {
+	if node != nil {
+		if val, ok :=
+			node.Annotations[PreserveMachineAnnotationKey]; ok &&
+			AllowedPreserveAnnotationValues.Has(val) {
+			return val, true
+		}
+		klog.Warningf(
```
thiyyakat, line 246: This would get printed for every un-annotated node, every reconciliation, right? 😅  I think you should only print if it is an invalid value. No point printing otherwise.

pkg/util/provider/machineutils/utils.go
```diff
@@ -180,34 +181,90 @@ func GetMachineDeploymentName(machine *v1alpha1.Machine) string {
...
+			machine.Annotations[LastAppliedNodePreserveValueAnnotationKey]; ok {
+			return "", true
+		}
+	}
+	if val, ok :=
+		machine.Annotations[PreserveMachineAnnotationKey]; ok &&
+		AllowedPreserveAnnotationValues.Has(val) {
+		return val, true
+	}
+	klog.Warningf(
```
thiyyakat, line 261: This too would get printed for every un-annotated machine, every reconciliation, until a node joins?

thiyyakat (commented):
Hi Vivek. Few more comments from me. PTAL. Thanks

pkg/controller/machineset.go
r4mek: The fix for this will come in #1154

pkg/controller/machineset.go
r4mek: I have changed the code such that a machine with only `auto-preserved` annotation is not considered as `preserved`. It should also have `PreserveExpiryTime` set.

pkg/util/provider/machinecontroller/machine.go
r4mek: After discussing with @aaronfern , printing this log every reconciliation seems fine. Will keep the existing code.

pkg/util/provider/machinecontroller/machine.go
r4mek: Will keep the existing code for now. If the team can find a better approach, will change the code.

pkg/controller/machineset.go
thiyyakat: As discussed offline, in combination with @gagan16k's bug fixes, where we recalculate the count in `manageAutoPreservationOfFailedMachines`, and by also considering only machines with auto-preserved annotation +PreserveExpiryTime when counting, this issue will get solved since the `PreserveExpiryTime` will get updated on the machine before `manageReplicas` is called.

pkg/util/provider/machineutils/utils_test.go
```diff
@@ -0,0 +1,214 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestGetPreserveAnnotationValue(t *testing.T) {
+	tests := []struct {
+		name           string
+		node           *corev1.Node
+		machine        *v1alpha1.Machine
+		expectedValue  string
+		expectedExists bool
```
thiyyakat, line 21: Please rename this variable. It can be misleading since the boolean returned is not solely based on the existence of key.

pkg/util/provider/machineutils/utils.go
thiyyakat: If someone has made a typo, it should not result in preservation ending. The case maybe unlikely, but say MCM has annotated a failed machine for auto-preservation and an enduser annotates the node with preserve=now. If there is a typo, it stops preservation. Instead the effect should be for MCM to log a warning and do nothing - neither start nor stop preservation. WDYT?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -911,97 +882,61 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
-		info.preserveExpiryTimeSet = true
-	}
-	nodeName := machine.Labels[v1alpha1.NodeLabelKey]
-	if nodeName != "" {
-		node, err := c.nodeLister.Get(nodeName)
+	if node != nil {
+		updatedMachine, err = c.updatePreserveAnnotationOnMachine(ctx, node.Annotations[machineutils.PreserveMachineAnnotationKey], clone)
 		if err != nil {
-			return info, err
+			updatedMachine = clone
```
thiyyakat, line 888: In L881, shouldn't you do `updatedMachine=clone` before returning? Otherwise original machine object will be returned. 

For that matter, before L873 as well, though in this case it won't cause any bugs since if node is not found, it is unlikely machine is Running. But for consistency this can be done.

pkg/util/provider/machineutils/utils.go
```diff
@@ -187,34 +188,92 @@ func GetMachineDeploymentName(machine *v1alpha1.Machine) string {
...
+		klog.Warningf(
+			"Machine %q doesn't have a valid annotation:%q=%q",
+			machine.Name,
+			PreserveMachineAnnotationKey,
+			machine.Annotations[PreserveMachineAnnotationKey],
+		)
+	}
+	if machine.Status.CurrentStatus.PreserveExpiryTime != nil {
+		return "", true
+	}
```
thiyyakat, line 277: If someone has made a typo, it should be a no-op. The case maybe unlikely, but say MCM has annotated a failed machine for auto-preservation and an enduser annotates the node with preserve=nw to extend the time or some other hacky stuff is done . If there is a typo, it stops preservation. 


In my opinion, the effect should be for MCM to log a warning and do nothing - neither start nor stop preservation. And it is easy enough to achieve by just setting bool to false. WDYT?

thiyyakat (commented):
Few more comments. PTAL. Thanks

pkg/util/provider/machinecontroller/machine.go
r4mek: For error cases, we enqueue the machine for next reconciliation. So, I don't really need to do `updatedMachine=clone`.

I could change L888 to directly return. The `updatedMachine=clone` is not really required.
Thanks for pointing it out.

thiyyakat: > For error cases, we enqueue the machine for next reconciliation

Not always. In the case of NotFound errors, we return nil (check defer) and don't return from the caller.

pkg/util/provider/app/options/options.go
takoverflow:
```suggestion
				MachinePreserveTimeout:                   metav1.Duration{DefaultMachinePreserveTimeout},
```

Now that we have a well-defined constant, the flag should re-use the same.

pkg/util/provider/machineutils/utils.go
r4mek: IIUC if the annotation key is present but the annotation value is not valid we should treat it as no-op?
What happens if the node has an invalid value but the machine has a valid value? Should that also be treated as no-op?

pkg/util/provider/machinecontroller/machine.go
r4mek: Right. Will make the change.

pkg/util/provider/machineutils/utils.go
thiyyakat: This is how I see it – if a node is annotated, we always honour the value over machine object. If the user has made a mistake while annotating the node, then we should log and wait for it to be corrected. Do you see a use case where a node is annotated with an invalid annotation key and it can't be rectified, requiring someone to annotate the machine object to achieve the intended effect? 

The only reason we supported annotations on the machine object is for cases where the node has not registered yet, or for some reason the node cannot be found. Using the machine object's annotation value as a fallback for when node value is incorrect doesn't seem useful to me.

@takoverflow , WDYT?

pkg/controller/machineset.go
```diff
@@ -904,75 +913,96 @@ func isMachineStatusEqual(s1, s2 v1alpha1.MachineStatus) bool {
 // or if it is a candidate for auto-preservation. If none of these conditions are met, it returns true indicating
 // that the failed machine should be terminated.
 func (c *controller) shouldFailedMachineBeTerminated(machine *v1alpha1.Machine) bool {
```
takoverflow: Please update the docstring, now it only checks if the machine is `preserved` by checking whether the expiry time is set and is in the future.

No other intermediate stage checks are done.

pkg/controller/machineset.go
```diff
@@ -904,75 +913,96 @@ func isMachineStatusEqual(s1, s2 v1alpha1.MachineStatus) bool {
...
 	var autoPreservationCandidates []*v1alpha1.Machine
 	var others []*v1alpha1.Machine
 	for _, m := range machines {
 		// check if machine is already annotated for preservation, if yes, skip. Machine controller will take care of the rest.
-		if machineutils.IsFailed(m) && !machineutils.AllowedPreserveAnnotationValues.Has(m.Annotations[machineutils.PreserveMachineAnnotationKey]) {
-			autoPreservationCandidates = append(autoPreservationCandidates, m)
+		// machine is considered as preserved when it has `PreserveExpiryTime` set.
+		if machineutils.IsFailed(m) {
+			if !machineutils.AllowedPreserveAnnotationValues.Has(m.Annotations[machineutils.PreserveMachineAnnotationKey]) ||
+				(m.Annotations[machineutils.PreserveMachineAnnotationKey] == machineutils.PreserveMachineAnnotationValueAutoPreserved && m.Status.CurrentStatus.PreserveExpiryTime == nil) {
```
takoverflow: Please extract this into a helper, this condition is getting unwieldy and hard to read. Consider
```go
func isMachineAutoPreservationCandidate(mc *v1alpha1.Machine) bool {
	// Only failed machines are candidates for auto-preservation
	if !machineutils.IsFailed(mc) {
		return false
	}

	// If machine is not already marked with valid preservation annotations, it is a candidate
	if !machineutils.AllowedPreserveAnnotationValues.Has(mc.Annotations[machineutils.PreserveMachineAnnotationKey]) {
		return true
	}

	// If machine is annotated with auto-preserve annotation but doesn't have its
	// PreserveExpiryTime set, its added to the list of candidates to complete its
	// preservation process.
	return mc.Annotations[machineutils.PreserveMachineAnnotationKey] == machineutils.PreserveMachineAnnotationValueAutoPreserved &&
		mc.Status.CurrentStatus.PreserveExpiryTime == nil
}
```
Clarifies the reasoning as well why the machine is considered or skipped.

pkg/controller/machineset.go
```diff
@@ -904,75 +913,96 @@ func isMachineStatusEqual(s1, s2 v1alpha1.MachineStatus) bool {
...
+	var errs []error
+	for index, machine := range autoPreservationCandidates {
 		if autoPreservationCapacityRemaining == 0 {
 			break
 		}
-		klog.V(2).Infof("Annotating failed machine %q for auto-preservation as part of machine set %q", m.Name, machineSet.Name)
-		updatedMachine, err := machineutils.UpdateMachineWithRetries(ctx, c.controlMachineClient.Machines(m.Namespace), c.machineLister, m.Namespace, m.Name, addAutoPreserveAnnotationOnMachine)
+
+		klog.V(2).Infof("Annotating failed machine %q for auto-preservation as part of machine set %q", machine.Name, machineSet.Name)
+		annotatedMachine, err := machineutils.PatchMachine(ctx, c.controlMachineClient.Machines(machine.Namespace), machine, func(m *v1alpha1.Machine) error {
```
takoverflow: The function is getting large, please try to extract the two patch operations in their own functions `annotateMachineForAutoPreservation` and `setAutoPreservationExpiryTime`.

These can take the machineclient `c.controlMachineClient.Machines(machine.Namespace)` which can be defined as a variable outside this loop and the machine object.

pkg/controller/machineset_util.go
```diff
@@ -111,96 +107,66 @@ func (c *controller) syncMachinesNodeTemplates(ctx context.Context, machineList
...
 	return nil
 }
 
-// copyMachineSetNodeTemplatesToMachines copies machineset's nodeTemplate to machine's nodeTemplate,
-// and returns true if machine's nodeTemplate is changed.
-// Note that apply and revision nodeTemplates are not copied.
-func copyMachineSetNodeTemplatesToMachines(machineset *v1alpha1.MachineSet, machine *v1alpha1.Machine) bool {
-	machineSetNodeTemplateCopy := machineset.Spec.Template.Spec.NodeTemplateSpec.DeepCopy()
-	machineNodeTemplateCopy := machine.Spec.NodeTemplateSpec.DeepCopy()
+// nodeTemplateOutOfSync returns true if machine's nodeTemplate is changed.
```
takoverflow:
```suggestion
// nodeTemplateOutOfSync returns true if Machine's NodeTemplate is different from the MachineSet it belongs to.
```

pkg/util/provider/machineutils/utils.go
```diff
@@ -187,34 +188,92 @@ func GetMachineDeploymentName(machine *v1alpha1.Machine) string {
...
+				return val, true
+			}
+			klog.Warningf(
+				"Node %q doesn't have a valid annotation:%q=%q",
+				machine.Labels[v1alpha1.NodeLabelKey],
+				PreserveMachineAnnotationKey,
+				node.Annotations[PreserveMachineAnnotationKey],
+			)
+		}
+		if _, ok :=
```
takoverflow: What if node is `nil` for some reason but machine has `LastAppliedNodePreserveValueAnnotation`?

In this case shouldn't we return `val, true`?

Currently this case is silently missed right?

pkg/util/provider/machineutils/utils.go
```diff
@@ -187,34 +188,92 @@ func GetMachineDeploymentName(machine *v1alpha1.Machine) string {
...
-		retryErr = nil
 	}
+	return machineClient.Patch(ctx, machine.Name, types.MergePatchType, patch, metav1.PatchOptions{}, subresources...)
+}
 
-	return machine, retryErr
+// GetPreserveAnnotationValue returns the preserve annotation value for the given node and machine
+// and a boolean informing whether we need to do any work or skip.
+// Invalid annotation values are treated as absent.
+func GetPreserveAnnotationValue(
```
takoverflow: Consider renaming it to something like `GetPreserveAnnotationState` since the name makes it seem like we're only returning the annotation. Feel free to take some liberties, this was just the first thing that popped in my head.

pkg/util/provider/machinecontroller/machine_util.go
takoverflow: This can make a machine transition to `Failed` phase, however we're not checking if the `PreserveExpiryTime` is non-nil and just setting it.

If `PreserveExpiryTime` for this machine is `nil` (which is going to be the case since it's a machine that's being created, shouldn't it be computed if the machine is being marked as `Failed`?

Is it that this case is being handled by MCS `autoPreservation`? What if the machine was already marked with some preservation annotation?

Am I missing something?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -796,20 +801,9 @@ func (c *controller) isCreationProcessing(machine *v1alpha1.Machine) bool {
...
-	nodeValue             string
-	machineValue          string
-	lastAppliedNodeValue  string
-	preserveExpiryTimeSet bool
-}
-
 // manageMachinePreservation manages machine preservation based on the preserve annotation values on the node and machine objects.
-func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1alpha1.Machine) (retry machineutils.RetryPeriod, err error) {
+func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1alpha1.Machine) (updatedMachine *v1alpha1.Machine, retry machineutils.RetryPeriod, err error) {
+	updatedMachine = machine
```
takoverflow: Can we avoid having three different references to an object in one function: `machine`, `updatedMachine` and `clone`.

What's the harm in doing `updatedMachine = machine.DeepCopy()` and getting rid of clone entirely.

We can also move the `GetPreserveAnnotationValue` check to this function's caller and keep this object modification completely gated. Then this gets the machine object and the computed annotation value.

Actually, I prefer the separation, it keeps things nice and tidy for this function WDYT?

pkg/util/provider/machinecontroller/machine.go
takoverflow: Why are we fetching the node again now? Haven't we already done it at L823, start of this function?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -911,97 +882,61 @@ func (c *controller) manageMachinePreservation(ctx context.Context, machine *v1a
...
+	return
 }
 
 // updatePreserveAnnotationOnMachine clears the machine's PreserveMachineAnnotationKey and sets
 // [machineutils.LastAppliedNodePreserveValueAnnotationKey] to nodeValue.
-func (c *controller) updatePreserveAnnotationOnMachine(ctx context.Context, nodeValue string, machine *v1alpha1.Machine) error {
-	clone := machine.DeepCopy()
-	if clone.Annotations == nil {
-		clone.Annotations = make(map[string]string)
+func (c *controller) updatePreserveAnnotationOnMachine(ctx context.Context, nodeValue string, machine *v1alpha1.Machine) (*v1alpha1.Machine, error) {
```
takoverflow: TODO: @takoverflow need to revisit this function

pkg/util/provider/machineutils/utils.go
```diff
@@ -187,34 +188,92 @@ func GetMachineDeploymentName(machine *v1alpha1.Machine) string {
...
-
-	retryErr := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-		var err error
-		machine, err = machineLister.Machines(namespace).Get(name)
-		if err != nil {
-			return err
+// PatchMachine patches a machine using a merge patch derived from mutateFn applied to the given machine object.
+// If optimisticLock is true, the patch includes the current resourceVersion to detect concurrent updates.
+// subresources optionally targets a subresource (e.g. "status"); omit it to patch the main resource.
+func PatchMachine(
```
takoverflow: TODO: @takoverflow need to check this

takoverflow (commented):
Thanks for taking up the task of cleaning up the machine failure status updates!

Was tricky to review, I'm sure it was even trickier to make the changes.

(I've added two TODOs for myself for functions that I've to check once, will update once I've taken a look at those as well)

pkg/util/provider/machinecontroller/machine.go
takoverflow: IMO, we should just stop using deleted annotation as a way to stop preservation. It makes MCM's job of interpreting state needlessly complex, annotation removal should be `no-op` and maintain status quo. User/operator should explicitly stop it by using the `false` value. MCM can then stop preservation and once its done remove the annotation from both objects.

All this mess because either node or machine can have `""` value to indicate preservation halting is causing a lot of mess WDYT?

pkg/util/provider/machinecontroller/machine.go
takoverflow: PTAL at https://github.com/gardener/machine-controller-manager/pull/1147#discussion_r4099355168

pkg/util/provider/machineutils/utils.go
takoverflow: > IIUC if the annotation key is present but the annotation value is not valid we should treat it as no-op?
What happens if the node has an invalid value but the machine has a valid value?

It should be no-op, preservation status should be maintained, if the node has invalid value, anyways you're warning and checking the machine value right?

An accidental typo resulting in preservation stopping is not nice IMO. It would've been good if stopping was always gated by value being `false` but we didn't go that route unfortunately.

takoverflow: > Using the machine object's annotation value as a fallback for when node value is incorrect doesn't seem useful to me.

Consider the case would the machine be marked `preserve=false` and node is marked with `preserve=now` which by some accident then becomes `preserve=nw`:
Function checks node value, warns and then checks the `lastnodeannotation` on the machine -> preservation continues.

> but say MCM has annotated a failed machine for auto-preservation and an enduser annotates the node with preserve=nw

If the initial node value was problematic, then there would be nothing on the `lastnodeannotation` for the machine, so whatever is there on the machine would be checked (auto-preserve) and honored, what's the harm there?

Am I missing something??

PR #1154, opened 2026-09-16
Bug fixes and address integration test flakes

gagan16k:
/kind flake

This PR addresses some bugs in MCM surfaced by integration tests on kwok:

1. `AutoPreserveFailedMachineMax` reduction tests result in deletion of both machines
    - Instances: [Run 1](https://github.com/gardener/machine-controller-manager/actions/runs/35057739887/job/104671338880), [Run 2](https://github.com/gardener/machine-controller-manager/actions/runs/35016237108/job/104540250253), [Run 3](https://github.com/gardener/machine-controller-manager/actions/runs/34961574062/job/104356146863)
    - Observed Behaviour: When `AutoPreserveFailedMachineMax` was reduced (2 -> 1), the controller unpreserved **both** machines instead of one, and both got deleted.
    - Root Cause: Two reconciles run: the first un-preserved one machine, updated `Status.AutoPreserveFailedMachineCount` and wrote it to the API server but it hadn't yet been persisted in the lister cache . The second still read the stale count (2), thought it needed to remove one more, and unpreserved the second machine too.
    - Fix: Count the currently auto-preserved machines from their annotations instead of the stale `Status.AutoPreserveFailedMachineCount`. A second reconcile then sees the reduced count and does nothing. The oldest machine is un-preserved first. Added unit tests in `machineset_test.go` to cover these cases.

Other changes to address a set of flakes in the existing integration tests.

1. Reducing MCD replicas to 2 in `Manual Preservation` tests (from 3 in `auto-preservation` tests) terminated a machine that got picked up by a test spec.
    - Instances: [Run](https://github.com/gardener/machine-controller-manager/actions/runs/35064107801/job/104690505417)
    - Fix: Keep the replica count consistent throughout the test body.
2. `BeforeEach` node readiness check stalls on a preserved node
    - Files changed: `nodes.go`, `kwok-config.yaml`
    - Root Cause: The BeforeEach check waits until every node is Ready (GetNumberOfReadyNodes==GetNumberOfNodes). Kwok sometimes does not update the leases until a node update event forces it to do so. Additionally the Get->Update way can sometimes not persist due to conflict errors. Nodes recovering from an earlier spec go through the kwok node-recover stage, which kept them NotReady for a long time. The next spec hangs waiting for nodes to be ready.
    - Fix: Removed this check as we wait for running machines in every preservation test, and added these checks where necessary in the other tests. Wherever the spec waits for preserved nodes to come back alive, add the kwok `"kwok/fail-condition":"Recover","kwok/recovery-attempts":"%d"` annotations on every poll interval to trigger an update edge on the node for the kwok controller. Also tighten the kwok `node-recover` stage (`immediateNextStage: true`) so recovering nodes come back faster. Renamed `addNodeRecoverAnnotation` to `attemptNodeRecovery`, and changed it to use a strategic-merge `Patch` to avoid update conflicts while the node is recovering.
3. Preservation specs capture a recovering or just-recovered machine as their target
    - Root Cause: Sometimes, a machine used by a previous spec and still running `DeferCleanup` gets picked up. This could have a `preserve=false` annotation and can get deleted as a consequence.
    - Fix: Sort `GetRunningMachineList` by `LastUpdateTime` ascending, so specs pick the machine that has been running the longest instead of a recently churned one. The tightened kwok `node-recover` stage also shrinks this window.
4. Freeze/Unfreeze test captures more events than assertion requires when run against a real cluster multiple times.
    - Instances: [Run](https://github.com/gardener/machine-controller-manager/actions/runs/34578005442/job/103194817544)
    - Root Cause: Across multiple runs, the machineset name is the same and the events can persist in the cluster across runs.
    - Fix: Track events by machineset UID instead of name

Other changes/improvements:
    - Added more detailed logging in `framework.go` which now tells exactly which machine a test is being executed on.
    - Extracted functions with reuse value as helpers from changes done for `Bug #1`
    - Make target to easily export kubeconfig of a simulated cluster.
    - Changed constants in `kwok-config.yaml` to more tighter values to make the average test runtime faster.
    - Requeue preserved machines immediately after their PreserveExpiryTime to avoid them potentially waiting a full reconcile (10mins).
    - Delete MCDs before running the `orphaned resources` test, as a test failure can leave the mcd still on the cluster causing this test to hang waiting for non-terminating VMs.
    - Upload mc and mcm process logs to the actions run in the event of a failure, for easy diagnosis.

**Special notes for Reviewers**
Current runtime statistics

```log
Date         : Fri Sep 18 06:28:35 IST 2026
Executed     : 20 run(s)
Passed       : 20
Failed       : 0
Pass rate    : 100% (20/20)

Runtime per run (all executed):
  best       : 11m 08s
  worst      : 16m 33s
  average    : 13m 56s
  median     : 13m 40s
```

**Release note**:

```bugfix developer
Fix bugs and integration test flakes
```


pkg/controller/deployment_machineset_util.go
```diff
@@ -138,7 +136,7 @@ func calculateMachineSetStatus(is *v1alpha1.MachineSet, filteredMachines []*v1al
 		}
 		// Count number of failed machines annotated with PreserveAnnotationValueAutoPreserved
 		// Cannot combine with above if block in case auto-preservation is not complete yet
-		if machine.Annotations[machineutils.PreserveMachineAnnotationKey] == machineutils.PreserveMachineAnnotationValueAutoPreserved {
+		if isAutoPreserved(machine) {
```
elankath: Please move this function `machineutils` package (`pkg/util/provider/machineutils`) and give it Godoc so it can be used in other places.

gagan16k: Moved

pkg/util/provider/machineutils/utils.go
```diff
@@ -169,6 +169,11 @@ func IsTriggeredForDeletion(m *v1alpha1.Machine) bool {
 	return m.Annotations[MachinePriority] == "1"
 }
 
+// IsAutoPreserved checks whether the machine currently carries the auto-preserved annotation.
+func IsAutoPreserved(m *v1alpha1.Machine) bool {
+	return m.Annotations[PreserveMachineAnnotationKey] == PreserveMachineAnnotationValueAutoPreserved
```
r4mek: You need to check for `machine.Status.CurrentStatus.PreserveExpiryTime != nil`

This is so that a machine which just has the `auto-preserved` annotation set but not the `PreserveExpiryTime` can again be considered for `auto-preservation`.

Check #1147

pkg/controller/machineset_test.go
```diff
@@ -2220,13 +2220,49 @@ var _ = Describe("machineset", func() {
 					preservedMachineCount: 0,
 				},
 			}),
-			Entry("should not trigger auto preservation of failed machines if AutoPreserveFailedMachineCount has reached AutoPreserveFailedMachineMax", testCase{
+			Entry("should not trigger auto preservation of additional failed machines if AutoPreserveFailedMachineCount has reached AutoPreserveFailedMachineMax", testCase{
 				setup: setup{
 					autoPreserveFailedMachineCount: 2,
 					autoPreserveFailedMachineMax:   2,
+					// Two machines are already auto-preserved, so the machine set is at capacity.
+					// machine-1 and machine-2 must therefore NOT be newly preserved.
```
takoverflow:
```suggestion
					// Two machines below are already auto-preserved, so the machine set is at capacity.
					// machine-1 and machine-2 must therefore NOT get preserved now.
```

pkg/simulatedprovider/cluster/kwok-config.yaml
```diff
@@ -210,8 +210,8 @@ metadata:
   name: node-heartbeat-with-lease
 spec:
   delay:
-    durationMilliseconds: 60000
-    jitterDurationMilliseconds: 61000
+    durationMilliseconds: 2500
```
takoverflow, line 213: Why has this been reduced to such a small value?
Was the 60s delay too long for some operation?

pkg/test/integration/common/helpers/machine_resources.go
```diff
@@ -133,11 +137,19 @@ func (c *Cluster) AreMachinesRunning(ctx context.Context, machineNames []string,
...
+			if !isSimulatedProvider {
+				return false
+			} else {
+				// Nudge kwok's node-recover stage to retry if it fired while the VAP was still active.
+				if nodeName := mc.Labels[v1alpha1.NodeLabelKey]; nodeName != "" {
+					c.attemptNodeRecovery(ctx, nodeName, int(time.Now().UnixMilli()))
+				}
+				allRunning = false
+			}
 		}
```
takoverflow: Simplification by inversion of conditions:

```go
		if mc.Status.CurrentStatus.Phase == v1alpha1.MachineRunning {
			continue
		}

		if !isSimulatedProvider {
			return false
		}
		// Nudge kwok's node-recover stage to retry if it fired while the VAP was still active.
		if nodeName := mc.Labels[v1alpha1.NodeLabelKey]; nodeName != "" {
			c.attemptNodeRecovery(ctx, nodeName, int(time.Now().UnixMilli()))
		}
		allRunning = false
```

pkg/simulatedprovider/cluster/kwok-config.yaml
takoverflow, line 320: As discussed offline, the trigger for running this stage can now be changed where the annotation key can be used to denote `recovery` and the value can be the timestamp when the annotation is added.

This is to allow for subsequent update events to be triggered via modification of a single annotation.

pkg/test/integration/common/helpers/nodes.go
```diff
@@ -7,11 +7,26 @@ package helpers
...
+	"github.com/gardener/machine-controller-manager/pkg/util/provider/machineutils"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 )
 
+// isNodePreserved reports whether the node is preserved, by checking if it carries the
+// preserve taint that MCM applies when a Failed machine's node is preserved.
+func isNodePreserved(node *corev1.Node) bool {
```
takoverflow, line 21: There's a `NodeCondition` with type `NodePreserved` that's added. Checking for its presence and state would be better IMO compared to looking for the taint.

```go
nodeops.GetCondition(node, v1alpha1.NodePreserved)
```

pkg/test/integration/common/helpers/nodes.go
```diff
@@ -24,10 +39,14 @@ func (c *Cluster) getNodes() (*corev1.NodeList, error) {
...
 // GetNumberOfReadyNodes tries to retrieve the list of node objects in the cluster.
+// Preserved nodes are excluded, since they are NotReady.
 func (c *Cluster) GetNumberOfReadyNodes() int16 {
 	nodes, _ := c.getNodes()
 	count := 0
 	for _, n := range nodes.Items {
+		if isNodePreserved(&n) {
+			continue
+		}
 		for _, nodeCondition := range n.Status.Conditions {
```
takoverflow, line 50: Since you're anyways iterating over node conditions, consider checking for the preservation info from the condition itself rather that using a different helper above

Ref: https://github.com/gardener/machine-controller-manager/pull/1154/changes#r4080837182

Makefile
```diff
@@ -15,6 +15,10 @@ COVERPROFILE       := test/output/coverprofile.out
 
 LEADER_ELECT 	   ?= "true" # If LEADER_ELECT is not set in the environment, use the default value "true"
 MACHINE_SAFETY_OVERSHOOTING_PERIOD:=1m
+# KUBE_API_QPS/KUBE_API_BURST default to the client-side rate limits.
+# The simulated-provider integration tests override them to cut client-side throttling against the fast kwok cluster.
+KUBE_API_QPS       ?= 20
```
takoverflow, line 20: Are these the default values?

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -309,6 +309,13 @@ func (c *controller) reconcileClusterMachine(ctx context.Context, machine *v1alp
 		)
 	}
 
+	// If the machine is preserved, requeue it at its PreserveExpiryTime with
+	// /a delay to ensure it is processed after the preservation period ends.
```
takoverflow:
```suggestion
	// a delay to ensure it is processed right after the preservation period ends.
	// This allows for quicker processing of preservation expiry for a machine since it doesn't wait
	// the entire `LongRetry` period.
```

pkg/test/integration/common/framework.go
```diff
@@ -144,7 +144,8 @@ type IntegrationTestFramework struct {
 // initializing resource tracker implementation.
 // Optially the timeout and polling interval are configurable as optional arguments
 // The default values used for Eventually to probe kubernetes cluster resources is
-// 300 seconds for timeout and  2 seconds for polling interval
+// 300 seconds for timeout and  500 milliseconds for polling interval for the simulated provider,
```
takoverflow:
```suggestion
// 300 seconds for timeout and 500 milliseconds polling interval for the simulated provider,
```

pkg/test/integration/common/framework.go
```diff
@@ -159,6 +160,9 @@ func NewIntegrationTestFramework(
 	}
 
 	pollingInterval := 2 * time.Second
+	if isSimulatedProvider {
```
takoverflow, line 163: This value is set in `SetupBeforeSuite`, which is invoked in `simulatedprovider/test/integration/controller/controller_test.go` after the call to `NewIntegrationTestFramework` so I'm not sure if its even set at this point, can you check locally if things are working as expected. I'm guessing even with the simulated provider it might still be using `2s` polling period.

pkg/test/integration/common/framework.go
```diff
@@ -652,12 +662,6 @@ func (c *IntegrationTestFramework) BeforeEachCheck() {
 		gomega.Expect(mcsession.ExitCode()).Should(gomega.Equal(-1))
 		ginkgo.By("Checking machineControllerManager process is running")
 		gomega.Expect(mcmsession.ExitCode()).Should(gomega.Equal(-1))
-		ginkgo.By("Checking nodes in target cluster are healthy")
```
takoverflow, line 655: Are you sure getting rid of this check cannot cause problems when running IT with infra?

pkg/test/integration/common/helpers/machine_resources.go
```diff
@@ -119,8 +122,9 @@ func (c *Cluster) IsMachineDeploymentDeleted(ctx context.Context, machineDeploym
 	return errors.IsNotFound(err)
 }
 
-// AreMachinesRunning returns boolean value indicating whether all the machines names passed to it are in the running state or not
-func (c *Cluster) AreMachinesRunning(ctx context.Context, machineNames []string, namespace string) bool {
```
takoverflow, line 130: There's a PR now that relies on `AreMachinesRunning` without any preservation in context.
https://github.com/gardener/machine-controller-manager/pull/1158
It might be worthwhile to have two methods now.

takoverflow (commented):
Thanks for looking in detail at the IT test failures and improving the flakiness.

I've added some comments, PTAL

Makefile
gagan16k: Yes, these are defaults

pkg/simulatedprovider/cluster/kwok-config.yaml
gagan16k: I had it for testing purposes, and pushed by mistake. I did test to see if they change performance, but it does not seem to matter. So, reverted in [10fc8f2](https://github.com/gardener/machine-controller-manager/pull/1154/commits/10fc8f2de49f1fdd54b2dfcf5baeb9b65daed321)

pkg/test/integration/common/helpers/nodes.go
gagan16k: Removed this function entirely as was not being used in any preservation tests, so was not serving any purpose

pkg/test/integration/common/helpers/nodes.go
gagan16k: Same as in https://github.com/gardener/machine-controller-manager/pull/1154#discussion_r4094373875

pkg/test/integration/common/framework.go
gagan16k: You were right, great find! Thanks for the comment, fixed

pkg/test/integration/common/framework.go
gagan16k: Yes, I ran IT multiple times to check on real providers as well, and saw no issues. I've also added some steps to check if machine is running in the `machine` and `machineDeployment` tests for some hardening of the testcases.

pkg/test/integration/common/helpers/machine_resources.go
gagan16k: Added back and removed `ArePreservedMachinesRunning`, now all tests use this function and do not annotate the nodes as a closure does this.

pkg/simulatedprovider/cluster/kwok-config.yaml
gagan16k: Added as part of [0c07b88](https://github.com/gardener/machine-controller-manager/pull/1154/commits/0c07b888c7abf88437bfa6b93a146b22cef5d465)

pkg/test/integration/common/framework.go
```diff
@@ -751,6 +768,14 @@ func (c *IntegrationTestFramework) ControllerTests() {
 						return exists
 					}).Should(gomega.BeTrue())
 
+					ginkgo.By("Ensuring machine is running before deleting its node")
```
takoverflow: Why is this necessary?

gagan16k: Its a replacement to the `ReadyNodes` check we had before. 
Probably not absolutely necessary, but I think there is value in waiting for the machine to be `Running` before we delete it. But I can remove it if it does not make sense

takoverflow: > Its a replacement to the ReadyNodes check we had before.

That used to run for each test, why is it only added here then?

> but I think there is value in waiting for the machine to be Running before we delete it. 

Deletion flow has no such limitation right? A machine in any phase can be terminated, is this to avoid orphan resources?

PR #1157, opened 2026-09-21
Revert machine termination flow enqueue changes

gagan16k:
Machines were observed stuck in `Terminating` phase
This issue surfaced on gardener tests on prow ([run](https://prow.gardener.cloud/view/gs/gardener-prow/pr-logs/pull/gardener_gardener/15724/pull-gardener-e2e-kind/2100455432869908480)) with master branch MCM as well, tests run by @shafeeqes.
- Root Cause: Some code paths (eg. [delNodeVolAttachments()](https://github.com/gardener/machine-controller-manager/blob/master/pkg/util/provider/machinecontroller/machine_util.go#L1800-L1806)) do not return a error when returning. This, due with changes in PR https://github.com/gardener/machine-controller-manager/pull/1102 cause it to not reconcile this machine again in the event there are no machine updates.
- Fix: Add back the unconditional re-enqueue on not having any error from reconcileClusterMachineTermination().

/kind regression

**Release note**:

```bugfix developer
Revert changes on machineTerminationFlow which caused a machine to be stuck in `Terminating` phase
```


takoverflow (commented):
/lgtm
Thanks for the fix!

thiyyakat:
/lgtm
Thanks for the quick fix!

PR #1158, opened 2026-09-22
Add support for hook to suspend VM deletion

### LINKED ISSUE #1131: Provide the ability to perform actions before termination of the VM

/area control-plane
/kind enhancement

As an operator, I would like to have the ability to perform some actions after the draining of the node and before the deletion of the VM. For example, in a self-hosted control plane whose VMs are managed by MCM, there should be a mechanism to remove the etcd member associated with the VM from the etcd cluster before MCM deletes the VM as part of replacement/rolling updates.

Proposed changes:
- A phase in the Machine lifecycle to signify to external actors that the drain is complete and that it is safe to perform the out-of-band operations necessary before termination of the VM. Suggestion: a `WaitingForPreTerminationHooks` phase.
- An annotation, whose absence could be used by MCM as a signal to proceed with the termination of the VM. Suggestion: `pre-terminate.delete.hook.machine.sapcloud.io/<name>=true` where `<name>` is set by the external actor for identification. So, once the external actor is done with the actions (like removal of the etcd member), it will delete the annotation from the Machine object, after which MCM can continue with the deletion flow. This annotation shall be set by the external actor as part of the `MachineTemplate` while creating the `MachineDeployment`.

Tasks:
- [ ] Confirm the naming of the phase and annotation
- [ ] Implement the handling of the introduced phase and annotation

**Why is this needed**:

Part of https://github.com/gardener/gardener/issues/15298

CaptainIRS:

Adds support for suspending VM deletion after node draining, allowing external controllers to perform pre-termination actions such as removing an etcd member before the VM is deleted.

When a terminating Machine reaches the VM deletion stage and has a suspension annotation, MCM:

- skips the provider VM deletion;
- sets the `InstanceDeletionSuspended=True` condition;
- reports the responsible owner and purpose in the condition message;
- retries until the annotation is removed.

The annotation format is:

```text
suspend-instance-deletion.node.machine.sapcloud.io/<purpose>: <owner>
```

For example:

```text
suspend-instance-deletion.node.machine.sapcloud.io/etcd-member-removal: my-controller
```

Once the external controller removes the annotation, MCM clears the condition and resumes VM deletion.

Fixes #1131

- No new `MachinePhase` is introduced; the existing `Terminating` phase and deletion flow are preserved.
- The annotation intentionally differs from the format suggested in the issue comments. `node.machine.sapcloud.io/suspend-instance-deletion/<purpose>` is not a valid Kubernetes annotation key because it contains two `/` separators. The implemented format uses `suspend-instance-deletion.node.machine.sapcloud.io` as the DNS prefix.
- Multiple suspension annotations are supported, with deterministic condition messages.
- Tests cover annotation parsing, suspension, condition updates, and resuming deletion.
- Verified with:
  - `go test ./pkg/util/annotations`
  - `go test ./pkg/util/provider/machinecontroller`

**Testing with g/g**:

<details>
<summary> This patch can be utilized. </summary>

```diff
diff --git a/example/provider-local/shoot.yaml b/example/provider-local/shoot.yaml
index e5731ed90a..f105440a09 100644
--- a/example/provider-local/shoot.yaml
+++ b/example/provider-local/shoot.yaml
@@ -25,12 +25,14 @@ spec:
     type: local
     workers:
     - name: local
+      annotations:
+        suspend-instance-deletion.node.machine.sapcloud.io/etcd-member-removal: my-controller
       machine:
         type: local
       cri:
         name: containerd
-      minimum: 1
-      maximum: 2
+      minimum: 3
+      maximum: 4
       maxSurge: 1
       maxUnavailable: 0
   kubernetes:
diff --git a/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go b/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
index a9051893f7..6dca6f6973 100644
--- a/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
+++ b/extensions/pkg/controller/worker/genericactuator/actuator_reconcile.go
@@ -304,7 +304,8 @@ func deployMachineDeployment(
 		machineDeployment.Spec.AutoPreserveFailedMachineMax = deployment.AutoPreserveFailedMachineMax
 		machineDeployment.Spec.Template = machinev1alpha1.MachineTemplateSpec{
 			ObjectMeta: metav1.ObjectMeta{
-				Labels: getMachineLabels(deployment.Strategy, labels, worker.Name),
+				Labels:      getMachineLabels(deployment.Strategy, labels, worker.Name),
+				Annotations: deployment.Annotations,
 			},
 			Spec: machinev1alpha1.MachineSpec{
 				Class: machinev1alpha1.ClassSpec{
diff --git a/go.mod b/go.mod
index 76f81bb4c2..9c215dd007 100644
--- a/go.mod
+++ b/go.mod
@@ -373,3 +373,5 @@ require (
 	sigs.k8s.io/randfill v1.0.0 // indirect
 	sigs.k8s.io/structured-merge-diff/v6 v6.4.2 // indirect
 )
+
+replace github.com/gardener/machine-controller-manager => ../machine-controller-manager
diff --git a/go.sum b/go.sum
index bd6f6f4334..57cb66fee4 100644
--- a/go.sum
+++ b/go.sum
@@ -270,8 +270,6 @@ github.com/gardener/dependency-watchdog v1.8.0 h1:+3FE8sR1V6YM9JsGqX7m9oUjhr1wXh
 github.com/gardener/dependency-watchdog v1.8.0/go.mod h1:jRIyBZ4ySWy+EdQjfLm/N+90vBPdWJ7khiBN7CGQs18=
 github.com/gardener/etcd-druid/api v0.38.2 h1:ltaqC9PV5rd39kj766UW+ZsN7LputyF1JplgcohsRHE=
 github.com/gardener/etcd-druid/api v0.38.2/go.mod h1:HsqEM/h4XuxPC4cFDm20V9TPAUqv0tjtnzds/nZ5p7Y=
-github.com/gardener/machine-controller-manager v0.62.1 h1:TQ4Qs9aaYTqSkqFownIaEAvnrYoNpBMcp7ZMYVEgUWM=
-github.com/gardener/machine-controller-manager v0.62.1/go.mod h1:ICsymXK4VHm7NVr8Mpl6USvYcTZjUNSk6+XaVu/dd74=
 github.com/gardener/pvc-autoscaler v0.3.0 h1:vneqFW7WESecbucng0Ravuz825BzKLI3jSKmiopccrs=
 github.com/gardener/pvc-autoscaler v0.3.0/go.mod h1:GVA7q8qiMi2UD/vcCBVbHEUijZvxVqZ1SGqiwKGHTE4=
 github.com/gardener/terminal-controller-manager v0.39.0 h1:w+aKFvM5k5EobILX9vrxgs6VbA5gd88Gyx/3a5m7nVc=
```

</details>

In local gardener setup:

```sh
SKAFFOLD_CACHE_ARTIFACTS=false make kind-up && SKAFFOLD_CACHE_ARTIFACTS=false make gardener-up && KUBECONFIG=dev-setup/kubeconfigs/virtual-garden/kubeconfig kubectl apply -f example/provider-local/shoot.yaml
```

(This uses MCM checked out to this branch. Make sure skaffold is not using the cached MCM local provider image)

```sh
KUBECONFIG=dev-setup/kubeconfigs/virtual-garden/kubeconfig kubectl annotate shoot/local -n garden-local gardener.cloud/operation=rollout-workers=local
# Do the following for the machines in the MCD to let termination proceed.
k annotate --overwrite -n shoot--local--local machine/shoot--local--local-local-z1-<machine> suspend-instance-deletion.node.machine.sapcloud.io/etcd-member-removal-
```

**Release note**:

```other operator
Add support for suspending VM deletion using suspend-instance-deletion.node.machine.sapcloud.io annotations.
```


pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1856,6 +1856,28 @@ func (c *controller) deleteVM(ctx context.Context, deleteMachineRequest *driver.
...
 
+	if annotationsutils.IsInstanceDeletionSuspended(machine) {
+		suspensionMessage := annotationsutils.GetInstanceDeletionSuspensionMessage(machine)
+		retryPeriod, err := c.updateInstanceDeletionSuspensionCondition(ctx, machine, v1.ConditionTrue, suspensionMessage)
+		if err != nil {
+			return retryPeriod, err
+		}
+
+		// Return an error to requeue while instance deletion is suspended.
+		return machineutils.ShortRetry, fmt.Errorf("instance deletion suspended: %s", suspensionMessage)
```
takoverflow: If instance deletion is suspended, does it make sense to re-queue again after only `5s`? Isn't that too short of an interval for any external actor to perform its activities, IMO it should be atleast `MediumRetry` (3m).

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1922,6 +1944,67 @@ func (c *controller) deleteVM(ctx context.Context, deleteMachineRequest *driver.
...
+	if err != nil {
+		klog.Warningf("Machine/status UPDATE for instance deletion suspension failed for machine %q. Retrying, error: %s", machine.Name, err)
+	}
+	if apierrors.IsConflict(err) {
+		return machineutils.ConflictRetry, err
+	}
+	return machineutils.ShortRetry, err
+}
+
+func getInstanceDeletionSuspensionCondition(conditions []v1.NodeCondition) *v1.NodeCondition {
```
takoverflow: Currently `pkg/controller/deployment.go` has an unexported `getMachineCondition`, can you move it to `pkg/util/provider/machineutils`, make it exported and re-use it here rather than creating another function just for getting a specific condition.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1856,6 +1856,28 @@ func (c *controller) deleteVM(ctx context.Context, deleteMachineRequest *driver.
 		deleteDuration time.Duration
 	)
 
+	if annotationsutils.IsInstanceDeletionSuspended(machine) {
+		suspensionMessage := annotationsutils.GetInstanceDeletionSuspensionMessage(machine)
```
takoverflow: `IsInstanceDeletionSuspended` calls `getInstanceDeletionSuspensions` but discards its output, only checks the length. If its `>0`, we again call the same thing, again check the length, this time actually using the output. Can this be improved?

```go
if msg, suspended := isInstanceDeletionSuspended(machine); suspended {
   // updateInstanceDeletionSuspension
   return retry, msg
}
```

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1922,6 +1944,67 @@ func (c *controller) deleteVM(ctx context.Context, deleteMachineRequest *driver.
 	return retryRequired, err
 }
 
+func (c *controller) updateInstanceDeletionSuspensionCondition(ctx context.Context, machine *v1alpha1.Machine, conditionStatus v1.ConditionStatus, conditionMessage string) (machineutils.RetryPeriod, error) {
```
takoverflow:
```suggestion
func (c *controller) updateInstanceDeletionSuspensionCondition(ctx context.Context, machine *v1alpha1.Machine, status v1.ConditionStatus, message string) (machineutils.RetryPeriod, error) {
```

Within the context of the function it's pretty clear what the status and message are referring to, so the prefix can be dropped from the argument names

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1922,6 +1944,67 @@ func (c *controller) deleteVM(ctx context.Context, deleteMachineRequest *driver.
...
+	clone := machine.DeepCopy()
+
+	conditionFound := false
+	conditionUpdated := false
+	for index, condition := range clone.Status.Conditions {
```
takoverflow: Can this not re-use `getMachineCondition`? That would also allow for getting rid of the `conditionFound`, `conditionUpdated` bools.

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1922,6 +1944,67 @@ func (c *controller) deleteVM(ctx context.Context, deleteMachineRequest *driver.
...
+		if condition.Type != v1alpha1.InstanceDeletionSuspended {
+			continue
+		}
+		conditionFound = true
+		if condition.Status == conditionStatus && condition.Message == conditionMessage {
+			break
```
takoverflow: Why would you not perform an early `return` here?

takoverflow (changes requested):
Thanks for the PR, just took a cursory glance and had some initial comments PTAL

pkg/util/provider/machinecontroller/machine_util.go
CaptainIRS: I had thought about this, but ideally I feel if there was an intermediate between `ShortRetry` and `MediumRetry` it would be great. You can imagine if we use `MediumRetry` and do a rolling update (considering the use-case of removing etcd members by observing this condition), it would take 9 minutes minimum to complete a rolling update, just waiting for requeue.

takoverflow: Wouldn't the external actor be removing the annotation once its job is done? That will lead to a machine update event, which should result in this machine being en-queued any which ways.
```go
	// to reconcile on change in annotations related to preservation
	if oldMachine.Annotations[machineutils.PreserveMachineAnnotationKey] != newMachine.Annotations[machineutils.PreserveMachineAnnotationKey] {
		c.enqueueMachine(newObj, "handling machine object preservation related UPDATE event")
		return
```

You can add a similar check in `updateMachine()` for the instance deletion annotations that checks if the annotation is removed/changed/added. WDYT?

This explicit retry shouldn't be this frequent.

pkg/apis/machine/v1alpha1/constants.go
```diff
@@ -11,6 +11,9 @@ const (
 	// the effective creation timeout for all Machine's belonging to this MachineDeployment. If specified, the value for this
 	// annotation takes precedence over the MachineDeployment.Spec.Template.Spec.MachineCreationTimeout.
 	AnnotationKeyMachineEffectiveCreationTimeout = "node.machine.sapcloud.io/effective-creation-timeout"
+	// AnnotationSuspendInstanceDeletionPrefix is the annotation key set on the Machine object that indicates that
```
takoverflow: Please add a comment in the `FAQ.md` answering
`Why is my machine not getting removed?` with details about the annotation and condition on the machine.

(Although this cannot be used by the end user, its useful information for operators that can be referred to)

pkg/apis/machine/v1alpha1/constants.go
```diff
@@ -11,6 +11,9 @@ const (
...
+	// the deletion of the VM instance shall be suspended after drain.
+	AnnotationSuspendInstanceDeletionPrefix = "suspend-instance-deletion.node.machine.sapcloud.io"
```
takoverflow: Please update the doc-string to denote that the annotation key can be used stand-alone as well and not just as a prefix since you seem to be targeting that use-case as well with your implementation.

pkg/util/annotations/annotations_test.go
```diff
@@ -312,4 +312,26 @@ var _ = Describe("annotations", func() {
 		)
 	})
 
+	Describe("#GetInstanceDeletionSuspensionMessage", func() {
+		It("returns a deterministic message for all suspension annotations", func() {
+			machine := &v1alpha1.Machine{ObjectMeta: metav1.ObjectMeta{
+				Annotations: map[string]string{
+					v1alpha1.AnnotationSuspendInstanceDeletionPrefix + "/etcd-member-removal":      "my-controller",
```
takoverflow: Please add another annotation/test for
> // Support the base annotation as a boolean-style hook.

pkg/util/provider/machinecontroller/machine_test.go
```diff
@@ -3117,6 +3134,310 @@ var _ = Describe("machine", func() {
 					),
 				},
 			}),
+			Entry("Deletion suspension annotation creates InstanceDeletionSuspended=True after drain and blocks VM deletion", &data{
```
takoverflow:
```suggestion
			Entry("Deletion suspension annotation adds condition InstanceDeletionSuspended=True after drain and blocks VM deletion", &data{
```

pkg/util/annotations/annotations.go
```diff
@@ -109,3 +109,63 @@ func GetEffectiveMachineCreationTimeout(object runtime.Object) (*metav1.Duration
...
+	owner   string
+}
+
+func getInstanceDeletionSuspensions(machine *v1alpha1.Machine) []instanceDeletionSuspension {
+	if machine == nil {
+		return nil
+	}
+
+	var suspensions []instanceDeletionSuspension
+	if owner, exists := machine.Annotations[v1alpha1.AnnotationSuspendInstanceDeletionPrefix]; exists {
```
thiyyakat: You probably need to do a `nil` check for Annotations before L149.

thiyyakat: You probably need to do a `nil` check for `machine.Annotations` before L149.

CaptainIRS: I'm not sure if this is an issue though, since `machine.Annotations` is not a pointer. `machine == nil` is being checked earlier, so I think this should be safe.

thiyyakat: You're right. Here since we are _reading_ from a possibly nil map and not writing to it, it wouldn't be a problem 👍

pkg/util/provider/machinecontroller/machine_util.go
CaptainIRS: Done in eace893a34c109cf86a97979d5d280a375b8c73a. Note that I've modified the implementation slightly to return the pointer to the original condition rather than a pointer to a copy (since in the previous iteration, it was returned after iterating in a range loop, which copies the value to a new variable).

pkg/util/provider/machinecontroller/machine_util.go
CaptainIRS: I've consolidated this in eace893a34c109cf86a97979d5d280a375b8c73a.

pkg/util/provider/machinecontroller/machine_util.go
CaptainIRS: Addressed in d32f4522f548b5c9ab31757be8a6183b8f902f00.

pkg/util/provider/machinecontroller/machine_util.go
CaptainIRS: Done in https://github.com/gardener/machine-controller-manager/commit/eace893a34c109cf86a97979d5d280a375b8c73a.

pkg/util/provider/machinecontroller/machine_util.go
CaptainIRS: This is no longer needed after using the exported `GetMachineCondition` in https://github.com/gardener/machine-controller-manager/commit/eace893a34c109cf86a97979d5d280a375b8c73a.

pkg/apis/machine/v1alpha1/constants.go
CaptainIRS: Done in https://github.com/gardener/machine-controller-manager/commit/eace893a34c109cf86a97979d5d280a375b8c73a.

pkg/apis/machine/v1alpha1/constants.go
CaptainIRS: Done in https://github.com/gardener/machine-controller-manager/commit/eace893a34c109cf86a97979d5d280a375b8c73a.

pkg/util/annotations/annotations_test.go
CaptainIRS: Done in https://github.com/gardener/machine-controller-manager/commit/eace893a34c109cf86a97979d5d280a375b8c73a.

pkg/util/provider/machinecontroller/machine_test.go
CaptainIRS: Changed in https://github.com/gardener/machine-controller-manager/commit/eace893a34c109cf86a97979d5d280a375b8c73a.

pkg/util/provider/machinecontroller/machine_util.go
CaptainIRS: Got it, thanks for the suggestion. I've fixed this in 2785e0dd077f48f5ac2736bdaf6607fe185baaca. I'm now enqueueing in the machineTerminationQueue if annotation has changed and if machine is under termination. Verified that changes to annotation without deletionTimestamp being set also don't unnecessarily put it in the queue:
<details>
<summary>logs</summary>

```
machine-controller-manager-76b6d44fb5-fdxwz machine-controller-manager-provider-local I0922 10:12:07.477515       1 machine.go:87] Skipping other non-spec updates for machine "shoot--local--local-local-z1-64cb7-cr9l8"
machine-controller-manager-76b6d44fb5-fdxwz machine-controller-manager I0922 10:12:07.482785       1 machineset.go:526] Processing the machineset "shoot--local--local-local-z1-64cb7" with replicas 2 associated with machine class: "shoot--local--local-local-z1-6bd28"
machine-controller-manager-76b6d44fb5-fdxwz machine-controller-manager-provider-local I0922 10:12:28.224739       1 machine.go:87] Skipping other non-spec updates for machine "shoot--local--local-local-z1-64cb7-cr9l8"
machine-controller-manager-76b6d44fb5-fdxwz machine-controller-manager I0922 10:12:28.227101       1 machineset.go:526] Processing the machineset "shoot--local--local-local-z1-64cb7" with replicas 2 associated with machine class: "shoot--local--local-local-z1-6bd28"
```

</details>

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1922,6 +1943,46 @@ func (c *controller) deleteVM(ctx context.Context, deleteMachineRequest *driver.
...
+		if conditionStatusChanged {
+			condition.LastTransitionTime = condition.LastHeartbeatTime
+		}
+		condition.Message = message
+		conditionUpdated = true
+	}
+
+	if !conditionUpdated {
+		return 0, nil
+	}
```
takoverflow:
```suggestion
		
	} else {
	    // condition not updated
		return 0, nil
	}
```

The bool `conditionUpdated` isn't needed then, since you only set it to `true` if any of the above cases are matched.

CaptainIRS: Ah missed this one, thanks for the catch. Fixed in da38a588fc8131f3e032d73037312439243908c0.

takoverflow:
Can you add an intergration test for this annotation as well
At https://github.com/gardener/machine-controller-manager/blob/1a7592cd9311f588e8b770709061d661c3cfce59/pkg/test/integration/common/framework.go#L701
Which tests machine deletion for a machine with this annotation, waits for condition to be added, then the annotation is removed by the test to allow VM deletion to proceed.

pkg/util/provider/machinecontroller/machine_util.go
takoverflow: Thanks

CaptainIRS:
I've added integration tests, I've also verified with simulated provider. PTAL, thanks.

<details>
<summary>logs</summary>

```
Machine controllers test machine resource deletion when machines available should suspend deletion until the suspension annotation is removed
/Users/whatever/go/src/github.com/gardener/machine-controller-manager/pkg/test/integration/common/framework.go:741
  > Enter [BeforeEach] Machine controllers test @ 09/23/26 09:20:50.389
  STEP: Checking machineController process is running @ 09/23/26 09:20:50.389
  STEP: Checking machineControllerManager process is running @ 09/23/26 09:20:50.389
  STEP: Checking nodes in target cluster are healthy @ 09/23/26 09:20:50.389
  < Exit [BeforeEach] Machine controllers test @ 09/23/26 09:20:50.394 (5ms)
  > Enter [It] should suspend deletion until the suspension annotation is removed @ 09/23/26 09:20:50.394
  STEP: Creating a machine for the suspension test @ 09/23/26 09:20:50.394
  STEP: Adding the instance deletion suspension annotation @ 09/23/26 09:20:56.421
  STEP: Deleting the machine @ 09/23/26 09:20:56.434
  STEP: Waiting for InstanceDeletionSuspended=True @ 09/23/26 09:20:56.442
  STEP: Removing the instance deletion suspension annotation @ 09/23/26 09:21:08.473
  STEP: Waiting until the machine object is deleted @ 09/23/26 09:21:08.487
  < Exit [It] should suspend deletion until the suspension annotation is removed @ 09/23/26 09:21:24.529 (34.135s)
• [34.140 seconds]
```

</details>

pkg/apis/machine/v1alpha1/constants.go
```diff
@@ -11,6 +11,12 @@ const (
 	// the effective creation timeout for all Machine's belonging to this MachineDeployment. If specified, the value for this
 	// annotation takes precedence over the MachineDeployment.Spec.Template.Spec.MachineCreationTimeout.
 	AnnotationKeyMachineEffectiveCreationTimeout = "node.machine.sapcloud.io/effective-creation-timeout"
+	// AnnotationSuspendInstanceDeletionPrefix is the annotation key or prefix set on the Machine object to indicate that
+	// deletion of the VM instance shall be suspended after drain. It can be used as a standalone key or with a slash
+	// and purpose suffix. For example:
+	//   suspend-instance-deletion.node.machine.sapcloud.io: my-controller
+	//   suspend-instance-deletion.node.machine.sapcloud.io/my-reason: my-controller
+	AnnotationSuspendInstanceDeletionPrefix = "suspend-instance-deletion.node.machine.sapcloud.io"
```
aaronfern: Since the convention so far has been `AnnotationKey...` and I would like to keep that
```suggestion
	AnnotationKeySuspendInstanceDeletionPrefix = "suspend-instance-deletion.node.machine.sapcloud.io"
```

pkg/util/annotations/annotations.go
```diff
@@ -109,3 +109,58 @@ func GetEffectiveMachineCreationTimeout(object runtime.Object) (*metav1.Duration
...
+	for _, suspension := range suspensions {
+		detail := suspension.owner
+		if suspension.purpose != "" {
+			detail += " for " + suspension.purpose
+		}
+		details = append(details, detail)
+	}
+
+	return "Instance Deletion suspended by " + strings.Join(details, ", ") + ".", true
+}
```
aaronfern, line 145: As the number of annotation grow, the size of this returned message grows too.
In theory, there could be too many annotations and could result in the condition message getting too large.
Do you think it would make sense to limit the message to contain some `n` owner details and if there are more add "and len(details)-n other owners" to the end of the string?
wdyt?

pkg/test/integration/common/framework.go
```diff
@@ -736,6 +737,94 @@ func (c *IntegrationTestFramework) ControllerTests() {
...
+							MachineV1alpha1().
+							Machines(controlClusterNamespace).
+							Get(ctx, helpers.McName, metav1.GetOptions{})
+						if err != nil {
+							return err
+						}
+						if machine.Status.CurrentStatus.Phase != v1alpha1.MachineRunning {
+							return fmt.Errorf("machine phase is %s", machine.Status.CurrentStatus.Phase)
+						}
+						return nil
```
aaronfern: There's a utility `AreMachinesRunning` in machine_resources.go that you can reuse

takoverflow (commented):
/lgtm

Thanks for the PR!

pkg/apis/machine/v1alpha1/constants.go
CaptainIRS: Done in f77fe6d8f437a92cba4200b222191be1c76c7252.

pkg/util/annotations/annotations.go
CaptainIRS: Addressed in f77fe6d8f437a92cba4200b222191be1c76c7252. I've added truncation logic based on the size of the message string (256 characters for the details). Rest of the details are just clubbed into "x other owner(s)".

pkg/test/integration/common/framework.go
CaptainIRS: Thanks for the suggestion, addressed in f77fe6d8f437a92cba4200b222191be1c76c7252.

pkg/util/annotations/annotations.go
```diff
@@ -109,3 +112,66 @@ func GetEffectiveMachineCreationTimeout(object runtime.Object) (*metav1.Duration
...
+	details := make([]string, 0, len(suspensions))
+	for i, suspension := range suspensions {
+		detail := suspension.owner
+		if suspension.purpose != "" {
+			detail += " for " + suspension.purpose
+		}
+		details = append(details, detail)
+		if len(strings.Join(details, ", ")) > maxInstanceDeletionSuspensionMessageLength {
+			if i == 0 {
+				details = append(details[:i], fmt.Sprintf("%d owner(s)", len(suspensions)-i))
```
aaronfern: nit, but the append is a bit redundant here, no?
Maybe something like `details = []string{fmt.Sprintf("%d owner(s)", len(suspensions)-i)}`?

pkg/util/provider/machinecontroller/machine_test.go
```diff
@@ -1575,6 +1576,22 @@ var _ = Describe("machine", func() {
...
+					Expect(actualCondition.Status).To(Equal(expectedCondition.Status))
+					Expect(actualCondition.Message).To(Equal(expectedCondition.Message))
+				}
+
+				if _, suspended := annotationsutils.IsInstanceDeletionSuspended(machine); suspended {
+					actualCondition := machineutils.GetMachineCondition(machine, v1alpha1.InstanceDeletionSuspended)
+					Expect(actualCondition).ToNot(BeNil())
+					Expect(actualCondition.Type).To(Equal(v1alpha1.InstanceDeletionSuspended))
+					Expect(actualCondition.Status).To(Equal(corev1.ConditionTrue))
+					Expect(actualCondition.Message).To(Equal("Instance Deletion suspended by my-controller for etcd-member-removal."))
```
gagan16k: Could we abstract `my-controller` and `etcd-member-removal` as consts, so that any future tests can use it instead of a string?

pkg/util/annotations/annotations.go
```diff
@@ -109,3 +112,66 @@ func GetEffectiveMachineCreationTimeout(object runtime.Object) (*metav1.Duration
...
+	owner   string
+}
+
+func getInstanceDeletionSuspensions(machine *v1alpha1.Machine) []instanceDeletionSuspension {
+	if machine == nil {
+		return nil
+	}
+
+	var suspensions []instanceDeletionSuspension
+	if owner, exists := machine.Annotations[v1alpha1.AnnotationKeySuspendInstanceDeletionPrefix]; exists {
```
gagan16k, line 158: Could we have validation for `owner` to be a non empty string? Or else we can get malformed messages like `"Instance Deletion suspended by  for xxx-reason."`

pkg/util/annotations/annotations.go
CaptainIRS: Ah, you're right, my bad. Fixed in 222befaadf99f41d5f5848647080b350c42911bd.

pkg/util/annotations/annotations.go
```diff
@@ -109,3 +112,66 @@ func GetEffectiveMachineCreationTimeout(object runtime.Object) (*metav1.Duration
 	}
 	return &metav1.Duration{Duration: effectiveMachineCreationTimeout}, nil
 }
+
+// IsInstanceDeletionSuspended returns a deterministic, bounded human-readable message for
+// instance-deletion suspension annotations on the machine and whether any are set.
+func IsInstanceDeletionSuspended(machine *v1alpha1.Machine) (string, bool) {
+	suspensions := getInstanceDeletionSuspensions(machine)
+	if len(suspensions) == 0 {
+		return "", false
```
gagan16k, line 121: Since we return `false` only when the returned string is `""`, do we need the boolean return value?

docs/faq.md
```diff
@@ -341,6 +341,7 @@ Though following could be the reasons but not limited to:
   - Long term: Please set more appropriate PDBs which allow disruption of at least one pod.
 - Expired cloud credentials can block the deletion of the machine from infrastructure.
 - Cloud provider can't delete the machine due to internal errors. Such situations are best debugged by using cloud provider specific CLI or cloud console.
+- A controller may intentionally suspend VM deletion by adding an annotation such as `suspend-instance-deletion.node.machine.sapcloud.io/my-reason: my-controller`. Check the `Machine`'s annotations and `.status.conditions` for `type: InstanceDeletionSuspended` with `status: "True"`; remove the suspension annotation once the controller's work is complete so MCM can continue deletion.
```
aaronfern: nit: just to make it clear that `my-reason` isn't actually part of the annotation
```suggestion
- A controller may intentionally suspend VM deletion by adding an annotation such as `suspend-instance-deletion.node.machine.sapcloud.io/<my-reason>: <my-controller>`. Check the `Machine`'s annotations and `.status.conditions` for `type: InstanceDeletionSuspended` with `status: "True"`; remove the suspension annotation once the controller's work is complete so MCM can continue deletion.
```

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -61,6 +62,12 @@ func (c *controller) updateMachine(oldObj, newObj any) {
 		klog.Errorf("couldn't convert to machine resource from object")
 		return
 	}
+	oldSuspensionMessage, oldSuspended := annotationsutils.IsInstanceDeletionSuspended(oldMachine)
+	newSuspensionMessage, newSuspended := annotationsutils.IsInstanceDeletionSuspended(newMachine)
+	if (oldSuspended != newSuspended || oldSuspensionMessage != newSuspensionMessage) && c.shouldMachineBeMovedToTerminatingQueue(newMachine) {
```
gagan16k: Can we gate annotation checks on `c.shouldMachineBeMovedToTerminatingQueue(newMachine)` so that we do not have to do all these things for every machine update?
Future pre delete hooks(if any) can also reuse this block to run their checks only in the case the machine is marked for termination.

aaronfern (commented):
Last nit, otherwise lgtm

docs/faq.md
CaptainIRS: Sure, fixed in 4e952597288f6a6506db7656388f8c8931c3e6a3.

aaronfern (commented):
Thanks for the PR and for making the changes!
/lgtm

gagan16k (commented):
Thanks for the PR!

pkg/apis/machine/v1alpha1/machine_types.go
```diff
@@ -264,6 +264,9 @@ const (
 	PreservedNodeDrainUnsuccessful string = "Preserved node could not be drained."
 )
 
+// InstanceDeletionSuspended is a node condition type to allow end-users to know that the deletion of the VM instance is suspended
+const InstanceDeletionSuspended corev1.NodeConditionType = "InstanceDeletionSuspended"
```
elankath: Please rename this constant to `ConditionInstanceDeletionSuspended`

pkg/util/provider/machinecontroller/machine_test.go
CaptainIRS: Addressed in 42c30d3cd. I've named the constants as `terminationHook*` so that it could be irrespective of the name of the new hook introduced.

pkg/util/annotations/annotations.go
CaptainIRS: I'm not sure if validation is the right approach. In 42c30d3cd, I've made it such that an empty owner will be printed as `unknown owner`, so that messages don't read like they're malformed. Please let me know if you have a different approach in mind.

pkg/util/annotations/annotations.go
CaptainIRS: I think it is good to be explicit, at least from the caller's perspective (since this is similar to the familiar patterns of returning "ok", "exists", etc. as bool in the second parameter in Go). But I'm open to changing it too if you feel this is unnecessary. WDYT?

pkg/util/provider/machinecontroller/machine.go
CaptainIRS: Good point, it does reduce the number of updates. Addressed in 42c30d3cd.

pkg/apis/machine/v1alpha1/machine_types.go
CaptainIRS: Addressed in [98013d4](https://github.com/gardener/machine-controller-manager/pull/1158/commits/98013d4c5a0a5db6fbfd9cdfbedce007da543e4e).

pkg/apis/machine/v1alpha1/constants.go
```diff
@@ -11,6 +11,12 @@ const (
 	// the effective creation timeout for all Machine's belonging to this MachineDeployment. If specified, the value for this
 	// annotation takes precedence over the MachineDeployment.Spec.Template.Spec.MachineCreationTimeout.
 	AnnotationKeyMachineEffectiveCreationTimeout = "node.machine.sapcloud.io/effective-creation-timeout"
+	// AnnotationKeySuspendInstanceDeletionPrefix is the annotation key or prefix set on the Machine object to indicate that
+	// deletion of the VM instance shall be suspended after drain. It can be used as a standalone key or with a slash
+	// and purpose suffix. For example:
+	//   suspend-instance-deletion.node.machine.sapcloud.io: my-controller
+	//   suspend-instance-deletion.node.machine.sapcloud.io/my-reason: my-controller
+	AnnotationKeySuspendInstanceDeletionPrefix = "suspend-instance-deletion.node.machine.sapcloud.io"
```
elankath, line 19: This should be be renamed to `node.machine.sapcloud.io.suspend-instance-deletion` instead so all annotations start with same namespace.

pkg/util/annotations/annotations.go
elankath: I think empty owner/ purpose should be disallowed, ie it should be ignored/skipped with a warning log. Please remove this base annotation check with empty purpose/owner.

pkg/test/integration/common/framework.go
```diff
@@ -736,6 +737,87 @@ func (c *IntegrationTestFramework) ControllerTests() {
 					}
 
 				})
+
+				ginkgo.It("should suspend deletion until the suspension annotation is removed", func() {
+					suspensionAnnotationKey := v1alpha1.AnnotationKeySuspendInstanceDeletionPrefix + "/integration-test"
+
+					ginkgo.By("Creating a machine for the suspension test")
```
gagan16k, line 744: Since #1126 is now merged, could you rebase this PR and club machine creation with the [first spec ](https://github.com/gardener/machine-controller-manager/blob/master/pkg/test/integration/common/framework.go#L684)of the test?
This is because on slow providers like **Azure**, we could add a significant amount of time(~ 9mins) to the overall test, mostly waiting for the machine creation

<details>
<summary>Test Log</summary>

```
Machine controllers test machine resource deletion when machines available should suspend deletion until the suspension annotation is removed
/Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:741
  > Enter [BeforeEach] Machine controllers test @ 09/23/26 14:24:52.257
  STEP: Checking machineController process is running @ 09/23/26 14:24:52.257
  STEP: Checking machineControllerManager process is running @ 09/23/26 14:24:52.257
  STEP: Checking nodes in target cluster are healthy @ 09/23/26 14:24:52.257
  < Exit [BeforeEach] Machine controllers test @ 09/23/26 14:24:53.238 (981ms)
  > Enter [It] should suspend deletion until the suspension annotation is removed @ 09/23/26 14:24:53.238
  STEP: Creating a machine for the suspension test @ 09/23/26 14:24:53.239
  ------------------------------
  Automatically polling progress:
    Machine controllers test machine resource deletion when machines available should suspend deletion until the suspension annotation is removed (Spec Runtime: 5m0.982s)
      /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:741
      In [It] (Node Runtime: 5m0.001s)
        /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:741
        At [By Step] Creating a machine for the suspension test (Step Runtime: 5m0.001s)
          /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:744

        Spec Goroutine
        goroutine 405 [select]
          github.com/onsi/gomega/internal.(*AsyncAssertion).match(0x603aa7a6c9a0, {0x103229a70, 0x603aa7fdf4f0}, 0x1, {0x0, 0x0, 0x0})
            /Users/*******/go/pkg/mod/github.com/onsi/gomega@v1.38.2/internal/async_assertion.go:558
          github.com/onsi/gomega/internal.(*AsyncAssertion).Should(0x603aa7a6c9a0, {0x103229a70, 0x603aa7fdf4f0}, {0x0, 0x0, 0x0})
            /Users/*******/go/pkg/mod/github.com/onsi/gomega@v1.38.2/internal/async_assertion.go:145
        > github.com/gardener/machine-controller-manager/pkg/test/integration/common.(*IntegrationTestFramework).ControllerTests.func1.2.1.2()
            /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:751
              | 	c.pollingInterval).
              | 	WithArguments(ctx, []string{helpers.McName}, controlClusterNamespace).
              > 	Should(gomega.BeTrue())
              |
              | ginkgo.By("Adding the instance deletion suspension annotation")
          github.com/onsi/ginkgo/v2/internal.extractBodyFunction.func3({0x102fe0740?, 0x0?})
            /Users/*******/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.27.2/internal/node.go:524
          github.com/onsi/ginkgo/v2/internal.(*Suite).runNode.func3()
            /Users/*******/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.27.2/internal/suite.go:942
          github.com/onsi/ginkgo/v2/internal.(*Suite).runNode in goroutine 25
            /Users/*******/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.27.2/internal/suite.go:907
  ------------------------------
  Automatically polling progress:
    Machine controllers test machine resource deletion when machines available should suspend deletion until the suspension annotation is removed (Spec Runtime: 6m0.988s)
      /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:741
      In [It] (Node Runtime: 6m0.006s)
        /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:741
        At [By Step] Creating a machine for the suspension test (Step Runtime: 6m0.006s)
          /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:744

        Spec Goroutine
        goroutine 405 [select]
          github.com/onsi/gomega/internal.(*AsyncAssertion).match(0x603aa7a6c9a0, {0x103229a70, 0x603aa7fdf4f0}, 0x1, {0x0, 0x0, 0x0})
            /Users/*******/go/pkg/mod/github.com/onsi/gomega@v1.38.2/internal/async_assertion.go:558
          github.com/onsi/gomega/internal.(*AsyncAssertion).Should(0x603aa7a6c9a0, {0x103229a70, 0x603aa7fdf4f0}, {0x0, 0x0, 0x0})
            /Users/*******/go/pkg/mod/github.com/onsi/gomega@v1.38.2/internal/async_assertion.go:145
        > github.com/gardener/machine-controller-manager/pkg/test/integration/common.(*IntegrationTestFramework).ControllerTests.func1.2.1.2()
            /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:751
              | 	c.pollingInterval).
              | 	WithArguments(ctx, []string{helpers.McName}, controlClusterNamespace).
              > 	Should(gomega.BeTrue())
              |
              | ginkgo.By("Adding the instance deletion suspension annotation")
          github.com/onsi/ginkgo/v2/internal.extractBodyFunction.func3({0x102fe0740?, 0x0?})
            /Users/*******/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.27.2/internal/node.go:524
          github.com/onsi/ginkgo/v2/internal.(*Suite).runNode.func3()
            /Users/*******/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.27.2/internal/suite.go:942
          github.com/onsi/ginkgo/v2/internal.(*Suite).runNode in goroutine 25
            /Users/*******/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.27.2/internal/suite.go:907
  ------------------------------
  Automatically polling progress:
    Machine controllers test machine resource deletion when machines available should suspend deletion until the suspension annotation is removed (Spec Runtime: 7m0.99s)
      /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:741
      In [It] (Node Runtime: 7m0.009s)
        /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:741
        At [By Step] Creating a machine for the suspension test (Step Runtime: 7m0.009s)
          /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:744

        Spec Goroutine
        goroutine 405 [select]
          github.com/onsi/gomega/internal.(*AsyncAssertion).match(0x603aa7a6c9a0, {0x103229a70, 0x603aa7fdf4f0}, 0x1, {0x0, 0x0, 0x0})
            /Users/*******/go/pkg/mod/github.com/onsi/gomega@v1.38.2/internal/async_assertion.go:558
          github.com/onsi/gomega/internal.(*AsyncAssertion).Should(0x603aa7a6c9a0, {0x103229a70, 0x603aa7fdf4f0}, {0x0, 0x0, 0x0})
            /Users/*******/go/pkg/mod/github.com/onsi/gomega@v1.38.2/internal/async_assertion.go:145
        > github.com/gardener/machine-controller-manager/pkg/test/integration/common.(*IntegrationTestFramework).ControllerTests.func1.2.1.2()
            /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:751
              | 	c.pollingInterval).
              | 	WithArguments(ctx, []string{helpers.McName}, controlClusterNamespace).
              > 	Should(gomega.BeTrue())
              |
              | ginkgo.By("Adding the instance deletion suspension annotation")
          github.com/onsi/ginkgo/v2/internal.extractBodyFunction.func3({0x102fe0740?, 0x0?})
            /Users/*******/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.27.2/internal/node.go:524
          github.com/onsi/ginkgo/v2/internal.(*Suite).runNode.func3()
            /Users/*******/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.27.2/internal/suite.go:942
          github.com/onsi/ginkgo/v2/internal.(*Suite).runNode in goroutine 25
            /Users/*******/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.27.2/internal/suite.go:907
  ------------------------------
  Automatically polling progress:
    Machine controllers test machine resource deletion when machines available should suspend deletion until the suspension annotation is removed (Spec Runtime: 8m0.994s)
      /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:741
      In [It] (Node Runtime: 8m0.012s)
        /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:741
        At [By Step] Creating a machine for the suspension test (Step Runtime: 8m0.012s)
          /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:744

        Spec Goroutine
        goroutine 405 [select]
          github.com/onsi/gomega/internal.(*AsyncAssertion).match(0x603aa7a6c9a0, {0x103229a70, 0x603aa7fdf4f0}, 0x1, {0x0, 0x0, 0x0})
            /Users/*******/go/pkg/mod/github.com/onsi/gomega@v1.38.2/internal/async_assertion.go:558
          github.com/onsi/gomega/internal.(*AsyncAssertion).Should(0x603aa7a6c9a0, {0x103229a70, 0x603aa7fdf4f0}, {0x0, 0x0, 0x0})
            /Users/*******/go/pkg/mod/github.com/onsi/gomega@v1.38.2/internal/async_assertion.go:145
        > github.com/gardener/machine-controller-manager/pkg/test/integration/common.(*IntegrationTestFramework).ControllerTests.func1.2.1.2()
            /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:751
              | 	c.pollingInterval).
              | 	WithArguments(ctx, []string{helpers.McName}, controlClusterNamespace).
              > 	Should(gomega.BeTrue())
              |
              | ginkgo.By("Adding the instance deletion suspension annotation")
          github.com/onsi/ginkgo/v2/internal.extractBodyFunction.func3({0x102fe0740?, 0x0?})
            /Users/*******/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.27.2/internal/node.go:524
          github.com/onsi/ginkgo/v2/internal.(*Suite).runNode.func3()
            /Users/*******/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.27.2/internal/suite.go:942
          github.com/onsi/ginkgo/v2/internal.(*Suite).runNode in goroutine 25
            /Users/*******/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.27.2/internal/suite.go:907
  ------------------------------
  STEP: Adding the instance deletion suspension annotation @ 09/23/26 14:32:54.091
  STEP: Deleting the machine @ 09/23/26 14:32:54.631
  STEP: Waiting for InstanceDeletionSuspended=True @ 09/23/26 14:32:54.898
  STEP: Removing the instance deletion suspension annotation @ 09/23/26 14:33:08.794
  STEP: Waiting until the machine object is deleted @ 09/23/26 14:33:09.472
  ------------------------------
  Automatically polling progress:
    Machine controllers test machine resource deletion when machines available should suspend deletion until the suspension annotation is removed (Spec Runtime: 9m0.997s)
      /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:741
      In [It] (Node Runtime: 9m0.015s)
        /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:741
        At [By Step] Waiting until the machine object is deleted (Step Runtime: 43.807s)
          /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:813

        Spec Goroutine
        goroutine 405 [select]
          golang.org/x/net/http2.(*ClientConn).internalRoundTrip(0x603aa7e0a8c0, 0x603aa8115180, 0x0)
            /Users/*******/go/pkg/mod/golang.org/x/net@v0.55.0/http2/transport.go:1030
          golang.org/x/net/http2.(*ClientConn).roundTrip(...)
            /Users/*******/go/pkg/mod/golang.org/x/net@v0.55.0/http2/transport.go:945
          golang.org/x/net/http2.(*ClientConn).RoundTrip(0x603aa7df8510?, 0x603aa7e0a8c0?)
            /Users/*******/go/pkg/mod/golang.org/x/net@v0.55.0/http2/clientconn.go:13
          golang.org/x/net/http2.(*Transport).roundTripViaPool(0x603aa792c5a0, 0x603aa8115180, {0x88?, 0x5?}, {0x1032279f0, 0x603aa7927560})
            /Users/*******/go/pkg/mod/golang.org/x/net@v0.55.0/http2/transport_common.go:279
          golang.org/x/net/http2.(*Transport).roundTripOpt(0x603aa792c5a0, 0x603aa8115180, {0x0?, 0x0?})
            /Users/*******/go/pkg/mod/golang.org/x/net@v0.55.0/http2/transport.go:413
          golang.org/x/net/http2.(*Transport).RoundTripOpt(...)
            /Users/*******/go/pkg/mod/golang.org/x/net@v0.55.0/http2/transport_common.go:231
          golang.org/x/net/http2.(*Transport).RoundTrip(0x603aa7d305f8?, 0x100ab7430?)
            /Users/*******/go/pkg/mod/golang.org/x/net@v0.55.0/http2/transport_common.go:226
          golang.org/x/net/http2.noDialH2RoundTripper.RoundTrip({0x603aa7b364e0?}, 0x603aa8115180?)
            /Users/*******/go/pkg/mod/golang.org/x/net@v0.55.0/http2/transport.go:2875
          net/http.(*Transport).roundTrip(0x603aa7b364e0, 0x603aa8115180)
            /opt/homebrew/Cellar/go@1.26/1.26.7/libexec/src/net/http/transport.go:623
          net/http.(*Transport).RoundTrip(0x10308a380?, 0x603aa81cc270?)
            /opt/homebrew/Cellar/go@1.26/1.26.7/libexec/src/net/http/roundtrip.go:33
          k8s.io/client-go/transport.(*userAgentRoundTripper).RoundTrip(0x603aa78c7aa0, 0x603aa8115040)
            /Users/*******/go/pkg/mod/k8s.io/client-go@v0.34.0/transport/round_trippers.go:174
          net/http.send(0x603aa8115040, {0x10321bb00, 0x603aa78c7aa0}, {0x1031ac460?, 0x603aa7d30b68?, 0x0?})
            /opt/homebrew/Cellar/go@1.26/1.26.7/libexec/src/net/http/client.go:264
          net/http.(*Client).send(0x603aa795f920, 0x603aa8115040, {0x603aa7d30ba8?, 0x100715b0c?, 0x0?})
            /opt/homebrew/Cellar/go@1.26/1.26.7/libexec/src/net/http/client.go:185
          net/http.(*Client).do(0x603aa795f920, 0x603aa8115040)
            /opt/homebrew/Cellar/go@1.26/1.26.7/libexec/src/net/http/client.go:733
          net/http.(*Client).Do(...)
            /opt/homebrew/Cellar/go@1.26/1.26.7/libexec/src/net/http/client.go:592
          k8s.io/client-go/rest.(*Request).request(0x603aa81b4f00, {0x10323d988, 0x103410da0}, 0x603aa7d30ff8)
            /Users/*******/go/pkg/mod/k8s.io/client-go@v0.34.0/rest/request.go:1088
          k8s.io/client-go/rest.(*Request).Do(0x603aa81b4f00, {0x10323d988, 0x103410da0})
            /Users/*******/go/pkg/mod/k8s.io/client-go@v0.34.0/rest/request.go:1136
          k8s.io/client-go/gentype.(*Client[...]).Get(0x103252ac0, {0x10323d988, 0x103410da0}, {0x101a81193, 0xc}, {{{0x0, 0x0}, {0x0, 0x0}}, {0x0, ...}})
            /Users/*******/go/pkg/mod/k8s.io/client-go@v0.34.0/gentype/type.go:165
        > github.com/gardener/machine-controller-manager/pkg/test/integration/common/helpers.(*Cluster).IsMachineDeleted(0x102f97de0?, {0x10323d988, 0x103410da0}, {0x101a81193, 0xc}, {0x603aa783819a?, 0x10307b120?})
            /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/helpers/machine_resources.go:107
              | 	MachineV1alpha1().
              | 	Machines(namespace).
              > 	Get(ctx, machineName, metav1.GetOptions{})
              |
              | return errors.IsNotFound(err)
          reflect.Value.call({0x102efd480?, 0x603aa7ff2020?, 0x603aa7d31c60?}, {0x101a75dde, 0x4}, {0x603aa7d18cc0, 0x3, 0x10071c70c?})
            /opt/homebrew/Cellar/go@1.26/1.26.7/libexec/src/reflect/value.go:586
          reflect.Value.Call({0x102efd480?, 0x603aa7ff2020?, 0x0?}, {0x603aa7d18cc0?, 0x0?, 0x603aa7a2e770?})
            /opt/homebrew/Cellar/go@1.26/1.26.7/libexec/src/reflect/value.go:369
          github.com/onsi/gomega/internal.(*AsyncAssertion).buildActualPoller.func3()
            /Users/*******/go/pkg/mod/github.com/onsi/gomega@v1.38.2/internal/async_assertion.go:337
          github.com/onsi/gomega/internal.(*AsyncAssertion).match(0x603aa7a62230, {0x103229a70, 0x603aa7ff2040}, 0x1, {0x0, 0x0, 0x0})
            /Users/*******/go/pkg/mod/github.com/onsi/gomega@v1.38.2/internal/async_assertion.go:560
          github.com/onsi/gomega/internal.(*AsyncAssertion).Should(0x603aa7a62230, {0x103229a70, 0x603aa7ff2040}, {0x0, 0x0, 0x0})
            /Users/*******/go/pkg/mod/github.com/onsi/gomega@v1.38.2/internal/async_assertion.go:145
        > github.com/gardener/machine-controller-manager/pkg/test/integration/common.(*IntegrationTestFramework).ControllerTests.func1.2.1.2()
            /Users/*******/go/src/github.com/gagan16k/machine-controller-manager/pkg/test/integration/common/framework.go:819
              | 			c.pollingInterval).
              | 			WithArguments(ctx, helpers.McName, controlClusterNamespace).
              > 			Should(gomega.BeTrue())
              | 	})
              | })
          github.com/onsi/ginkgo/v2/internal.extractBodyFunction.func3({0x102fe0740?, 0x0?})
            /Users/*******/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.27.2/internal/node.go:524
          github.com/onsi/ginkgo/v2/internal.(*Suite).runNode.func3()
            /Users/*******/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.27.2/internal/suite.go:942
          github.com/onsi/ginkgo/v2/internal.(*Suite).runNode in goroutine 25
            /Users/*******/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.27.2/internal/suite.go:907
  ------------------------------
  < Exit [It] should suspend deletion until the suspension annotation is removed @ 09/23/26 14:34:11.536 (9m18.272s)
• [559.253 seconds]
```
</details>

pkg/util/annotations/annotations.go
```diff
@@ -109,3 +112,66 @@ func GetEffectiveMachineCreationTimeout(object runtime.Object) (*metav1.Duration
...
+	}
+
+	details := make([]string, 0, len(suspensions))
+	for i, suspension := range suspensions {
+		detail := suspension.owner
+		if suspension.purpose != "" {
+			detail += " for " + suspension.purpose
+		}
+		details = append(details, detail)
+		if len(strings.Join(details, ", ")) > maxInstanceDeletionSuspensionMessageLength {
```
elankath, line 131: the actual message can be up to 256 + 33 = 289 characters, not 256. So either rename the constant name`maxInstanceDeletionSuspensionMessageLength` to `maxInstanceDeletionSuspensionDetailsLength` or change its value appropriately.

pkg/util/provider/machineutils/utils.go
```diff
@@ -124,6 +124,16 @@ const (
 	NodePreservedTaintKey = "node.machine.sapcloud.io/preserved"
 )
 
+// GetMachineCondition returns a condition matching the type from the machine's status.
+func GetMachineCondition(machine *v1alpha1.Machine, conditionType v1.NodeConditionType) *v1.NodeCondition {
```
elankath, line 128: This is a problematic compared to earlier `getMachineCondition`. We returns a pointer into the slice. The The old `getMachineCondition` returned `&cond` (pointer to loop variable copy), which
  was actually safer for mutation

`updateInstanceDeletionSuspensionCondition` is  silently modifying the original machine's
  conditions.

pkg/util/annotations/annotations.go
gagan16k: Makes sense, considering that the function reads like it returns a boolean yes/no

pkg/util/annotations/annotations.go
```diff
@@ -109,3 +112,69 @@ func GetEffectiveMachineCreationTimeout(object runtime.Object) (*metav1.Duration
...
+
+	return "Instance Deletion suspended by " + strings.Join(details, ", ") + ".", true
+}
+
+type instanceDeletionSuspension struct {
+	purpose string
+	owner   string
+}
+
+func getInstanceDeletionSuspensions(machine *v1alpha1.Machine) []instanceDeletionSuspension {
```
elankath, line 152: Plase do a check that you have `<prefix>/<purpose>: <owner>` always with non-empty purpose and non-empty owner. If purpose or owner is empty, kindly log with a warning and skip.

pkg/util/annotations/annotations.go
```diff
@@ -109,3 +112,69 @@ func GetEffectiveMachineCreationTimeout(object runtime.Object) (*metav1.Duration
...
+func IsInstanceDeletionSuspended(machine *v1alpha1.Machine) (string, bool) {
+	suspensions := getInstanceDeletionSuspensions(machine)
+	if len(suspensions) == 0 {
+		return "", false
+	}
+
+	details := make([]string, 0, len(suspensions))
+	for i, suspension := range suspensions {
+		detail := suspension.owner
+		if suspension.owner == "" {
```
elankath, line 127: Remove this - owner and purpose are required to be present.

pkg/util/provider/machinecontroller/machine.go
```diff
@@ -61,6 +62,14 @@ func (c *controller) updateMachine(oldObj, newObj any) {
 		klog.Errorf("couldn't convert to machine resource from object")
 		return
 	}
+	if c.shouldMachineBeMovedToTerminatingQueue(newMachine) {
+		oldSuspensionMessage, oldSuspended := annotationsutils.IsInstanceDeletionSuspended(oldMachine)
```
elankath, line 66: Why are we checking the message ? Shouldn't we only check the `bool` return value which can be simplified to `annotationsutils.IsInstanceDeletionSuspended(oldMachine) != annotationsutils.IsInstanceDeletionSuspended(newMachine)` ?

pkg/util/provider/machinecontroller/machine_util.go
```diff
@@ -1922,6 +1943,41 @@ func (c *controller) deleteVM(ctx context.Context, deleteMachineRequest *driver.
 	return retryRequired, err
 }
 
+func (c *controller) updateInstanceDeletionSuspensionCondition(ctx context.Context, machine *v1alpha1.Machine, status v1.ConditionStatus, message string) (machineutils.RetryPeriod, error) {
+	clone := machine.DeepCopy()
+
+	condition := machineutils.GetMachineCondition(clone, v1alpha1.ConditionInstanceDeletionSuspended)
+	if condition == nil && status == v1.ConditionTrue {
```
elankath, line 1950: Why can't you use `nodeops.CloneAndAddCondition` here ? Also according to k8s best practice, a CamelCase machine-readable identifier for reason should be set on the condition. You can choose something appropriate.

```go
      clone.Status.Conditions = nodeops.CloneAndAddCondition(clone.Status.Conditions, v1.NodeCondition{
          Type:               v1alpha1.ConditionInstanceDeletionSuspended,
          Status:             status,
          LastHeartbeatTime:  now,
          LastTransitionTime: now,
          Message:            message,
          Reason: "CamelCaseReason"
      })
```
You can modify `CloneAndAddCondition` to add a `updated bool` returned as second parameter if you wish it to also handle the 'has changed' check.

elankath (changes requested):

pkg/util/annotations/annotations.go
```diff
@@ -109,3 +112,69 @@ func GetEffectiveMachineCreationTimeout(object runtime.Object) (*metav1.Duration
 	}
 	return &metav1.Duration{Duration: effectiveMachineCreationTimeout}, nil
 }
+
+// IsInstanceDeletionSuspended returns a deterministic, bounded human-readable message for
+// instance-deletion suspension annotations on the machine and whether any are set.
+func IsInstanceDeletionSuspended(machine *v1alpha1.Machine) (string, bool) {
```
elankath, line 118: With non-empty purpose and owner, this can be just simplified to: (haven't tested), without all that slice indexing stuff

```
 func IsInstanceDeletionSuspended(machine *v1alpha1.Machine) (string, bool) {
        suspensions := getInstanceDeletionSuspensions(machine)
        if len(suspensions) == 0 {
                return "", false
        }

        const prefix = "Instance Deletion suspended by "
        var details []string
        for _, s := range suspensions {
                detail := s.owner + " for " + s.purpose
                if len(prefix+strings.Join(append(details, detail), ", ")+".") > maxInstanceDeletionSuspensionMessageLength {
                        omitted := len(suspensions) - len(details)
                        if len(details) == 0 {
                                details = []string{fmt.Sprintf("%d owner(s)", len(suspensions))}
                        } else {
                                details = append(details, fmt.Sprintf("and %d other owner(s)", omitted))
                        }
                        break
                }
                details = append(details, detail)
        }

        return prefix + strings.Join(details, ", ") + ".", true
  }
```

elankath (changes requested):

